PR #31, opened 2022-06-06
Add SendEvent API + more API doc improvements

cgillum:
~~This PR makes the `TaskOrchestrationShim` type from internal to public. This is to support alternate hosts, such as a PowerShell-based host.~~ __(we scrapped this idea for now because of maintenance problems we anticipate in the future)__.

This PR contains a variety of changes:
- A new `SendEvent` API for orchestrations
- Improved API documentation
- Improved error handling
- Other misc. updates

Because there is new public surface area in this change, I'm also incrementing the version from 0.4.1-beta to **0.5.0-beta**.


davidmrdavid (approved):
LGTM! It would be good to test this in MyGet with the PS SDK prior to making a release. It may be the case that this is the only dependency we needed to make public (I hope so!) but I can't be certain until we try :)

cgillum:
@davidmrdavid here's a copy you can try out: https://www.myget.org/feed/azure-appservice/package/nuget/Microsoft.DurableTask.Client/0.5.0-beta.

One change you'll need to make is to use the `WorkerContext` class instead of `FunctionsWorkerContext`.

cgillum:
FYI @jviau I unfortunately lost track of this PR + some other pending changes on my local machine. I've updated it with the latest from main, but I had to make a few additional changes to some of your changes. I'll try to add comments in the code changes directly.

jviau:
I am a bit worried about making these public, it is going to hamper our ability to refactor them without breaking changes. I'd like to understand the Powershell scenarios to see if there is an alternative way we can accomplish this.

To unblock Powershell for now, I am fine moving forward with this PR. But I'd like to re-evaluate and possible change this before GA.

cgillum:
@jviau per our discussion, I went ahead and reverted the changes that made the shim and related types public. I've also renamed this PR accordingly. I'd still like to keep the other changes that ended up getting piled onto this PR, though. Let me know if you think this PR is good to merge.

jviau (approved):

PR #37, opened 2022-08-15
Microsoft.Azure.Functions.Worker 1.8.0 version

jorgelevy:
Microsoft.Azure.Functions.Worker has been released for 1.8.0 version


jviau (approved):

cgillum:
I'm going to go ahead and merge this, but it looks like there's an even more recent update available (v1.1.0.0), so we should try to upgrade to that as a separate PR when we get a chance.

Thanks @jorgelevy for this contribution!

PR #38, opened 2022-08-16
Expose ParentInstance in TaskOrchestrationContext

### LINKED ISSUE #23: Add ParentInstanceId to TaskOrchestrationContext
It's useful for a sub-orchestrator to know the ID of its parent orchestration instance. This should be exposed to the context parameter.

Durable Functions (in-process) support this:

* https://github.com/Azure/azure-functions-durable-extension/blob/bcdc74bc8a0319221d46b55980ee18ca8261179e/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs#L44

jviau:
This adds `ParentInstance` property `TaskOrchestrationContext`.

**WARNING**: this is a breaking change - adding a new `abstract` property will break any projects which also try to implement this class. @cgillum - what is the process for breaking changes like this?

resolves #23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,3 +3,9 @@
 ## v0.4.1-beta
 
 Initial public release
+
+## Unreleased changes
+
+### Breaking changes
+
+- Added new abstract property `TaskOrchestrationContext.ParentInstance`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 ﻿# Changelog
 
-## v0.4.1-beta
-
-Initial public release
-
-## Unreleased changes
+## v0.5.0-beta
 
 ### Breaking changes
 
 - Added new abstract property `TaskOrchestrationContext.ParentInstance`.
+
+## v0.4.1-beta
+
+Initial public release
--- a/src/DurableTask/TaskOrchestrationContext.cs
+++ b/src/DurableTask/TaskOrchestrationContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.Core;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.DurableTask;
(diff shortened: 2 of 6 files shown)


src/DurableTask/TaskOrchestrationContext.cs
```diff
@@ -24,6 +25,11 @@ public abstract class TaskOrchestrationContext
     /// </summary>
     public abstract string InstanceId { get; }
 
+    /// <summary>
+    /// Gets the parent instance. Null if no parent orchestration.
```
cgillum, line 29: nit, this style is a bit more consistent with other properties: `"Gets the parent instance or <c>null</c> if there is no parent orchestration."`

src/DurableTask/TaskOrchestrationContext.cs
```diff
@@ -24,6 +25,11 @@ public abstract class TaskOrchestrationContext
...
+    /// </summary>
+    public abstract ParentInstance? Parent { get; }
```
cgillum, line 31: For breaking changes like this, there are two things we can consider:
1. We can just make the change and document it in the CHANGELOG.md file. It's expected that we'll make some breaking changes prior to releasing the "GA" version of the SDK. No need for a major version bump yet.
2. Instead of making this `abstract`, we could make it `virtual` with a default implementation that does something such as `throw new NotImplementedException()`.

jviau: Went with option 1. Would rather capitalize on the preview state and do the right thing now.

CHANGELOG.md
```diff
@@ -3,3 +3,9 @@
 ## v0.4.1-beta
 
 Initial public release
+
+## Unreleased changes
+
+### Breaking changes
+
+- Added new abstract property `TaskOrchestrationContext.ParentInstance`.
```
cgillum, line 11: The `v0.4.1` release already went out, so let's add to the top of this file a new `v0.5.0-beta` section and put your new content there. Here's how I've been doing it for the MSSQL backend for DTFx: https://github.com/microsoft/durabletask-mssql/blob/main/CHANGELOG.md (I originally copied the approach of the [KEDA project's CHANGELOG.md file](https://github.com/kedacore/keda/blob/main/CHANGELOG.md)).

cgillum (approved):
LGTM! Just one comment about the CHANGELOG.md file:

PR #40, opened 2022-08-17
Suppress build warnings

jviau:
- Fix test project nuget package reference `0.4.0-beta` -> `0.4.1-beta`
- Suppress logger message warnings - they are intentional. We want those values as dimensions, but not in the message.
- Suppress packaging of samples
- Suppress erroneous NU5128 with generator project via `SuppressDependenciesWhenPacking`


cgillum (approved):
LGTM!

PR #41, opened 2022-08-23
Update to DTFx gRPC 0.3.1, add support for purge via filter

### LINKED ISSUE #3: Purge instance support
### Requirement

Orchestration clients should be able to purge orchestration data, as described [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#purge-instance-history). Purging single instances and multiple instances using filters should be supported.

Note that this functionality also needs to be added to the sidecar.

One or more integration tests covering this scenario should also be added.

### Implementation notes

The following integration test for the MSSQL backend can be used as a reference for writing tests in this project: https://github.com/microsoft/durabletask-mssql/blob/24fdfe6979f5af3d57b4f52aaa3eb30a2ba5038f/test/DurableTask.SqlServer.Tests/Integration/PurgeTests.cs#L35

jviau:
This PR adds support for the gRPC `PurgeInstances` API. Both via filter for purging multiple instances, or by instance ID for purging a single one.

Resolves #3


src/DurableTask/PurgeInstancesFilter.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Query for purging orchestration instances.
+/// </summary>
+/// <param name="CreatedFrom">Date created from.</param>
+/// <param name="CreatedTo">Date created to.</param>
+/// <param name="Statuses">The statuses.</param>
+public record PurgeInstancesFilter(
```
jviau, line 12: I have redefined a type here because I saw that we already redefined some `DurableTask.Core` types in this package, such as `OrchestrationRuntimeStatus`. Do we want to continue down this line, or should I just use [DurableTask.Core..PurgeInstancesFilter](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/PurgeInstanceFilter.cs) instead?

cgillum: I'm trying to avoid surfacing up DurableTask.Core types in this SDK. I want to keep it as an implementation detail just in case we decide to replace it with something different/more optimal.

src/DurableTask/DurableTaskClient.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.Core;
```
cgillum, line 7: If possible, I'd like to avoid exposing DurableTask.Core types in the public API.

src/DurableTask/Grpc/ProtoUtils.cs
```diff
@@ -176,6 +175,29 @@ static Timestamp ConvertTimestamp(DateTime dateTime)
...
 
+    internal static Timestamp? ToTimestamp(this DateTime? dateTime)
+        => dateTime.HasValue ? dateTime.Value.ToTimestamp() : null;
+
+    internal static Timestamp ToTimestamp(this DateTimeOffset dateTime) => Timestamp.FromDateTimeOffset(dateTime);
+
+    internal static Timestamp? ToTimestamp(this DateTimeOffset? dateTime)
+        => dateTime.HasValue ? dateTime.Value.ToTimestamp() : null;
+
+#pragma warning disable 0618
```
cgillum, line 186: Which warning is this that we're disabling? Can you add a comment to help make that more clear?

src/DurableTask/LogHelpers.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Non-generated helpers for <see cref="ILogger" />.
+/// </summary>
+static class LogHelpers
+{
+    public static void PurgingInstances(this ILogger logger, PurgeInstancesFilter filter)
```
cgillum, line 13: Do we need a separate static class for this or can we put it in Logs.cs?

jviau: I can try again, but putting it in `Logs.cs` broke the code gen. Not sure if that was just my editor going haywire or not though.

cgillum (approved):
LGTM!

PR #42, opened 2022-08-29
Add support for multi instance query

### LINKED ISSUE #2: Multi-instance query support
### Requirement

Orchestration clients should be able to issue filtered queries to the state store for orchestration instances, as described [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=csharp#query-instances-with-filters).

This will require work both in the SDK and sidecar projects.

One or more integration tests should also be written which covers this scenario.

jviau:
This PR introduces support for querying multiple orchestration instances over gRPC. The implementation also takes a _page_ out of Azure SDKs book by copying their `AsyncPageable` design (pun intended). This abstraction layer is important as we want to accomplish 2 primary goals with our async enumerable implementation:

1. I want to enumerate all pages right away: `AsyncPageable<T>` implements `IAsyncEnumerable<T>`.
2. I want to enumerate a page at a time and/or pause/resume paging: `AsyncPageable<T>.AsPages()` allows for this via `.AsPages(continuationToken).FirstAsync()` from `System.Linq.Async` package. 
    - This means queries can be paused/resumed/rehydrated in an abstract fashion. This is important for DTFx usage itself, with this design someone can easily break a query down into a page per activity (so we don't have to serialize it all at once!).
    - We can even provide an example of an orchestration yield results in pages via `AsyncPageable<T>` itself.

resolves #2


src/DurableTask/AsyncPageable{T}.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// A collection of values that may take multiple service requests to iterate over.
+/// </summary>
+/// <typeparam name="T">The type of the value.</typeparam>
+public abstract class AsyncPageable<T> : IAsyncEnumerable<T>
+    where T : notnull
+{
+    // This code was adapted from Azure SDK AsyncPageable.
```
cgillum, line 13: Is there a link to source code you can provide as part of this comment?

src/DurableTask/AsyncPageable{T}.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </param>
+    /// <returns>An async enumerable of pages.</returns>
+    public abstract IAsyncEnumerable<Page<T>> AsPages(
+        string? continuationToken = default, int? pageSizeHint = default);
+
+    /// <inheritdoc/>
+    public async IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+    {
+        await foreach (Page<T> page in this.AsPages()
+            .ConfigureAwait(false).WithCancellation(cancellationToken))
```
cgillum, line 35: Is `.ConfigureAwait(false)` really necessary? It makes me really nervous because of how it can cause code to break in certain contexts.

src/DurableTask/DurableTaskClient.cs
```diff
@@ -191,7 +191,12 @@ public abstract class DurableTaskClient : IAsyncDisposable
     /// <inheritdoc cref="WaitForInstanceStartAsync(string, CancellationToken, bool)"/>
     public abstract Task<OrchestrationMetadata?> GetInstanceMetadataAsync(string instanceId, bool getInputsAndOutputs);
 
-    // TODO: Multi-instance query
+    /// <summary>
+    /// Queries orchestration instances 
+    /// </summary>
+    /// <param name="filter">Filters down the instances included in the query.</param>
+    /// <returns></returns>
```
cgillum, line 198: Let's add a description for the return value.

src/DurableTask/Pageable.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Runtime.CompilerServices;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Pageable helpers.
+/// </summary>
+static class Pageable
+{
+    // This code was adapted from Azure SDK AsyncPageable.
```
cgillum, line 13: Is there a link to source code you can provide as part of this comment?

src/DurableTask/Pageable.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        async IAsyncEnumerable<Page<T>> AsPagesCore(
+            string? continuationToken = default,
+            int? pageSizeHint = default,
+            [EnumeratorCancellation] CancellationToken cancellation = default)
+        {
+            do
+            {
+                Page<T> page = await this.pageFunc(continuationToken, pageSizeHint, cancellation)
+                    .ConfigureAwait(false);
```
cgillum, line 74: Same question about the necessity of `.ConfigureAwait(false)`.

cgillum (commented):
Looking good so far! I'll admit that this pageable stuff is new to me. Will you be able to add some tests to help verify the functionality (and provide examples of how to use the API)?

src/DurableTask/AsyncPageable{T}.cs
jviau: It is safe (and recommended) for libraries to call `ConfigureAwait(false)`. You will continue on the thread pool. but only your continuation is affected. This will not affect the `ConfigureAwait` behavior of the caller to this function - so if they are on a specialized context, their continuation will still be posted back to their special context when our work completes.

jviau: Can you explain how it would cause code to break in certain contexts? I can only see it breaking if this function itself depended on a specific context. Or it was calling into something that also depended on a specific context. However, this code does not have any of those dependencies.

src/DurableTask/Pageable.cs
jviau: Same as above - this code has no reliance on any specific context, so `ConfigureAwait(false)` is safe to use. Calling code can still control their own awaiter behavior, allowing them to still be posted back to their own context.

src/DurableTask/AsyncPageable{T}.cs
cgillum: > You will continue on the thread pool, but only your continuation is affected.

DTFx orchestration code is an example where forcing a callback onto the thread pool can break the application. Obviously, we don't expect anyone writing orchestration code to call into these APIs, but there may be other cases where thread pool threads are not desirable.

That said, this might be a complete knowledge gap on my part. When you say, "your continuation", do you mean that the caller of the async enumerator will have its callbacks posted to any context it requires? So as an example, if I (for some strange reason) needed to call this from an Orchestration thread, would the callbacks still be posted back to the orchestration thread instead of a worker pool thread?

jviau: I actually do see this type as very useful in orchestrations. Someone could write an activity which yields a page at a time, and then use an `AsyncPageable<T>` extension method to create a streamlined experience for calling the activity from an orchestration. How will this `ConfigureAwait(false)` affect that though is a great question. I forgot the detail of DTFx sync context is it based entirely around executing inline. Will need to think about this a bit more.

jviau: I did some testing (the integration test infra you have works great btw!) and yeah - `ConfigureAwait(false)` breaks using this within an orchestration.

jviau: I have removed `ConfigureAwait(false)` for now.

jviau:
The latest iteration includes a bunch of changes to other files as part of adding unit tests.

- Added FluentAssertions and System.Linq.Async package to test projects
- Added / rearranged build/eng files so that unit test projects are strong name signed and can have `InternalsVisibleTo` added.
    - Added some msbuild settings to test projects, like implicit usings. There are using changes to other files as a result.
 - Added unit tests and an integration test.

jviau:
@cgillum the integration test I added shows an example of the benefit of this abstraction layer. Instead of the static switch case implementation there though, an activity would in turn call an API or something that itself is paged. For a CosmosDB query, they could return a single page at a time there. For a query to the API I just introduced, the implementation would be something like `return client.GetInstances({ continuation, pageSize }).AsPages().FirstAsync();`.

This won't really be exposed right away to customers as the `Pageable` helper is `internal` still. I want to think about the `ConfigureAwait` issue some more before exposing that. Ideally we would also introduce a strongly typed request-style API for the `OrchestrationContext`, and then we can add overloads for activities which return `Page<T>` and automatically wrap them into an `AsyncPageable<T>` for the caller.

Directory.Build.props
```diff
@@ -0,0 +1,17 @@
...
+    <LangVersion>10.0</LangVersion>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <!-- Signing -->
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)eng/key.snk</AssemblyOriginatorKeyFile>
+    <MicrosoftPublicKey>00240000048000009400000006020000002400005253413100040000010001000505410141442095e0b0466df68ac32d158abdf0bd9cf26407d7a9d20b93656530556f23979f20b067628b9baee75fc6bba5c349519585a2852c1843bc61d74c4cbf9d80429cbdcbf609ea70fd62061eb65f40a6c9d505dd71cb119ef51f589a014d3b56159abbc38825fafbd119b6e97ebd9a5f3862a5c06220c680a6ac9eec</MicrosoftPublicKey>
```
cgillum, line 14: I'd avoid calling this a "Microsoft" public key since that sounds a little more official than it really is (this is simply a key that I generated on my desktop). Maybe just `StrongNamePublicKey`?

test/DurableTask.Sdk.Tests/PageableIntegrationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License.
+
+using FluentAssertions;
+using Microsoft.DurableTask.Grpc;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DurableTask.Tests;
+
+public class PageableIntegrationTests : IntegrationTestBase
```
cgillum, line 11: I was expecting that we'd have an integration test that uses your new `DurableTaskClient.GetInstances(OrchestrationQuery?)` API to query against created orchestration instances. Is that coming for this PR?

jviau: Oh okay. Yeah let me add that.

jviau: They are added. Also added some `GetInstanceMetadata` tests. Hopefully these tests show the purpose of `AsyncPageable<T>`. One of the main things hit upon in that test is how pageable can be used to capture a query and then pause/resume it at different points via continuation tokens.

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            using (new AssertionScope())
+            {
+                metadata.Name.Should().Be(OrchestrationName);
+                metadata.InstanceId.Should().StartWith("GetInstances_EndToEnd-");
+                metadata.RuntimeStatus.Should().Be(status);
+
+                // InMemoryOrchestrationService always returns these in a query.
+                // metadata.FailureDetails.Should().BeNull();
+                // metadata.SerializedInput.Should().BeNull();
+                // metadata.SerializedOutput.Should().BeNull();
```
cgillum, line 70: Do these need to be removed (or uncommented)?

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+
+        OrchestrationQuery query = new() { InstanceIdPrefix = "GetInstances_EndToEnd" };
+
+        static async Task ForEachAsync(Func<string, Task> func)
```
cgillum, line 76: nit: based on the name, I mistakenly thought this was a generic For-Each implementation, but I now see it's hardcoded to two orchestration instances. Maybe name this `ForEachOrchestrationAsync`, or something less generic?

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            await func("GetInstances_EndToEnd-1");
+            await func("GetInstances_EndToEnd-2");
+        }
+
+        await using DurableTaskGrpcWorker server = await this.StartAsync();
+        DurableTaskClient client = this.CreateDurableTaskClient();
+
+        string notIncluded = await client.ScheduleNewOrchestrationInstanceAsync(OrchestrationName, input: false);
```
cgillum, line 85: Unused local variable? If it's just to indicate that this instance is distinct from the others, maybe a comment would be better? It's hard to know without context if an unused local variable is a bug or not.

cgillum (approved):
Thanks for adding the new tests! Left a few comments, but nothing blocking.

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
jviau: I have them there commented because I _would_ like to verify that behavior, but `InMemoryOrchestrationService` from the sidecar library always returns those, even though I have `FetchInputsAndOutputs = false`. The comment is a reminder incase anyone else tries to verify the same thing I did.

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
jviau: Updated the test to use this variable.

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
jviau: Updated the comment and inverted the statement, so we will know when to correct the assertion.

test/DurableTask.Sdk.Tests/Grpc/DurableTaskGrpcClientIntegrationTests.cs
```diff
@@ -67,9 +67,17 @@ static void AssertMetadata(OrchestrationMetadata metadata, OrchestrationRuntimeS
                 // InMemoryOrchestrationService always returns these in a query.
                 // The NotBeNull() here is to force this test failure when correct behavior
                 // is added, so we remember to change bellow to "NotNull()".
-                metadata.FailureDetails.Should().NotBeNull();
+                metadata.FailureDetails.Should().BeNull();
```
cgillum, line 70: The code no longer matches the comment?

PR #47, opened 2022-09-22
Fix code generation for Durable Functions function-based syntax.

### LINKED ISSUE #35: The type of namespace could not be found for POCO input and outputs
Appears that the generated code does not include a `using` statement to reference the namespace of the POCO types. A workaround can be to place the POCO types in the `Microsoft.DurableTask` namespace.

Example:

```csharp
using Microsoft.Azure.Functions.Worker;
using Microsoft.Azure.Functions.Worker.Http;
using Microsoft.DurableTask;
using System;
using System.Threading;
using System.Threading.Tasks;

namespace FunctionApp1_DotnetIsolated
{
    public class Person
    {
        public string Name { get; set; }
    }

    public class Message
    {
        public string Value { get; set; }
    }

    public class Function1
    {
        [Function(nameof(Function1))]
        public async Task<HttpResponseData> RunAsync([HttpTrigger] HttpRequestData req, [DurableClient] DurableClientContext durableContext)
        {
            var person = BinaryData.FromStream(req.Body).ToObjectFromJson<Person>();
            var instanceId = await durableContext.Client.ScheduleNewOrchestrationInstanceAsync(nameof(SayHelloOrchestrator), input: person);
            var result = await durableContext.Client.WaitForInstanceCompletionAsync(instanceId, CancellationToken.None, getInputsAndOutputs: true).ConfigureAwait(false);

            var response = req.CreateResponse();
            await response.WriteAsJsonAsync(result.ReadOutputAs<Message>()).ConfigureAwait(false);

            return response;
        }
    }

    public class SayHelloOrchestrator
    {
        [Function(nameof(SayHelloOrchestrator))]
        public async Task<Message> RunAsync([OrchestrationTrigger] TaskOrchestrationContext context)
        {
            var person = context.GetInput<Person>();
            var message = await context.CallActivityAsync<Message>(nameof(GetMessageActivity), person);
            return message;
        }
    }

    public class GetMessageActivity
    {
        [Function(nameof(GetMessageActivity))]
        public Task<Message> RunAsync([ActivityTrigger] Person person)
        {
            var message = new Message { Value = $"Hello {person.Name}!" };
            return Task.FromResult(message);
        }
    }
}
```

Generated code:

_It is also interesting to note the double `Task<Task<T>>` return type, not sure if this is expected, seems strange though._

```csharp
// <auto-generated/>
#nullable enable

using System;
using System.Threading.Tasks;
using Microsoft.Azure.Functions.Worker;
using Microsoft.Extensions.DependencyInjection;

namespace Microsoft.DurableTask
{
    public static class GeneratedDurableTaskExtensions
    {
        public static Task<Task<Message>> CallGetMessageActivityAsync(this TaskOrchestrationContext ctx, Person person, TaskOptions? options = null)
        {
            return ctx.CallActivityAsync<Task<Message>>("GetMessageActivity", person, options);
        }
    }
}
```

Package references:

```xml
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="0.4.1-beta" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ServiceBus" Version="5.4.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0-preview2" OutputItemType="Analyzer" />
<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.8.0" />
```

jviau:
This PR addresses 2 issues with function based syntax code gen for Durable Functions:

1. Functions returning `Task<T>` would get double nested as `Task<Task<T>>`
2. Functions using custom types would not resolve their fully qualified namespace, leading to compilation issues.

resolves #35

### DIFF
--- a/src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
+++ b/src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
@@ -60,7 +60,7 @@ namespace Microsoft.DurableTask.Generators.AzureFunctions
                 returnType = ((GenericNameSyntax)returnType).TypeArgumentList.Arguments[0];
             }
 
-            if (!SyntaxNodeUtility.TryGetParameter(method, kind, out TypedParameter? parameter) || parameter == null)
+            if (!SyntaxNodeUtility.TryGetParameter(model, method, kind, out TypedParameter? parameter) || parameter == null)
             {
                 return false;
             }
--- a/src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
+++ b/src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DurableTask.Generators.AzureFunctions
 {
     public enum DurableFunctionKind
     {
-        Unknonwn,
+        Unknown,
         Orchestration,
         Activity
     }
@@ -50,6 +50,14 @@ namespace Microsoft.DurableTask.Generators.AzureFunctions
             if (!SyntaxNodeUtility.TryGetReturnType(method, out TypeSyntax returnType))
             {
                 return false;
+            }
+
+            INamedTypeSymbol taskSymbol = model.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1")!;
+            INamedTypeSymbol returnSymbol = (INamedTypeSymbol)model.GetTypeInfo(returnType).Type!;
+            if (SymbolEqualityComparer.Default.Equals(returnSymbol.OriginalDefinition, taskSymbol))
+            {
+                // this is a Task<T> return value, lets pull out the generic.
+                returnType = ((GenericNameSyntax)returnType).TypeArgumentList.Arguments[0];
             }
 
             if (!SyntaxNodeUtility.TryGetParameter(method, kind, out TypedParameter? parameter) || parameter == null)
--- a/src/DurableTask.Generators/AzureFunctions/SyntaxNodeUtility.cs
+++ b/src/DurableTask.Generators/AzureFunctions/SyntaxNodeUtility.cs
@@ -7,7 +7,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace Microsoft.DurableTask.Generators.AzureFunctions
 {
-    public static class SyntaxNodeUtility
+    static class SyntaxNodeUtility
     {
         public static bool TryGetFunctionName(SemanticModel model, MethodDeclarationSyntax method, out string? functionName)
         {
@@ -39,11 +39,11 @@ namespace Microsoft.DurableTask.Generators.AzureFunctions
 
         public static bool TryGetFunctionKind(MethodDeclarationSyntax method, out DurableFunctionKind kind)
         {
-            var parameters = method.ParameterList.Parameters;
+            SeparatedSyntaxList<ParameterSyntax> parameters = method.ParameterList.Parameters;
 
             foreach (var parameterSyntax in parameters)
             {
-                var parameterAttributes = parameterSyntax.AttributeLists.SelectMany(a => a.Attributes);
+                IEnumerable<AttributeSyntax> parameterAttributes = parameterSyntax.AttributeLists.SelectMany(a => a.Attributes);
 
                 foreach (var attribute in parameterAttributes)
                 {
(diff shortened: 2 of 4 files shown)


src/DurableTask.Generators/AzureFunctions/SyntaxNodeUtility.cs
```diff
@@ -1,29 +1,33 @@
...
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace Microsoft.DurableTask.Generators.AzureFunctions
 {
-    public static class SyntaxNodeUtility
+    static class SyntaxNodeUtility
     {
         public static bool TryGetFunctionName(SemanticModel model, MethodDeclarationSyntax method, out string? functionName)
         {
+            functionName = null;
```
cgillum, line 14: nit: I prefer to leave this near the return statement so that the compiler can help me know when I've forgotten to correctly assign it to a non-null value.

cgillum (approved):
LGTM!

jviau:
@cgillum I found another case to fully qualify the type name.

src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
```diff
@@ -21,59 +21,79 @@ public class DurableFunction
...
+            ITypeSymbol returnType,
+            HashSet<string> requiredNamespaces)
         {
             this.FullTypeName = fullTypeName;
             this.RequiredNamespaces = requiredNamespaces;
             this.Name = name;
             this.Kind = kind;
             this.Parameter = parameter;
-            this.ReturnType = returnTypeSyntax;
+            this.ReturnType = SyntaxNodeUtility.GetRenderedTypeExpression(returnType, false);
```
cgillum, line 39: Does this `false` mean we don't support nullable annotations in return types?

src/DurableTask.Generators/AzureFunctions/DurableFunction.cs
```diff
@@ -21,59 +21,79 @@ public class DurableFunction
...
             {
+                function = null;
                 return false;
             }
 
             if (!SyntaxNodeUtility.TryGetQualifiedTypeName(model, method, out string? fullTypeName))
             {
+                function = null;
                 return false;
             }
```
cgillum, line 80: Would it make sense to move the `TryGetParameter` and `TryGetQualifiedName` checks further up, above the new return type code you added? Could that short-circuiting help save some CPU cycles or do we very rarely expect those methods to return false?

src/DurableTask.Generators/AzureFunctions/SyntaxNodeUtility.cs
```diff
@@ -57,55 +62,54 @@ public static bool TryGetFunctionKind(MethodDeclarationSyntax method, out Durabl
...
+                if (typeInfo is null)
                 {
                     return false;
                 }
 
-                if (toProcess is not PredefinedTypeSyntax && typeInfo.Type.ContainingNamespace.IsGlobalNamespace)
+                if (typeInfo.ContainingNamespace.IsGlobalNamespace)
                 {
                     requiredNamespaces = null;
                     return false;
```
cgillum, line 86: I know you didn't write this code, but is it correct for us to bail like this if we encounter a global namespace? Why wouldn't we continue to process remaining types in the `remaining` list?

src/DurableTask.Generators/AzureFunctions/SyntaxNodeUtility.cs
```diff
@@ -144,6 +152,27 @@ public static bool TryGetQualifiedTypeName(SemanticModel model, MethodDeclaratio
...
+
+            if (supportsNullable && symbol.IsReferenceType && symbol.NullableAnnotation != NullableAnnotation.Annotated)
+            {
+                symbol = symbol.WithNullableAnnotation(NullableAnnotation.Annotated);
+            }
+
+            string expression = symbol.ToString();
+            if (expression.StartsWith("System.") && symbol.ContainingNamespace.Name == "System")
+            {
+                expression = expression.Substring("System.".Length);
```
cgillum, line 170: We do this because the `System` namespace is implicitly imported by the compiler?

cgillum (commented):
I took a slightly deeper look this time through and have a few more questions that didn't come up from earlier iterations.

jviau:
I have opened #48 to track the comments here and other improvements to the generator.

PR #49, opened 2022-09-27
Switch netfx sample to sdk style csproj

jviau:


jviau:
@cgillum is it okay to switch this project to `net48` as well?

cgillum:
> @cgillum is it okay to switch this project to `net48` as well?

That’s fine by me, but any particular reason?

cgillum (approved):

jviau:
> > @cgillum
> >             
> >             Chris Gillum
> >             FTE is it okay to switch this project to `net48` as well?
> 
> That’s fine by me, but any particular reason?

Mostly for convenience. I think having `net48` installed will be more common that other versions. For example, installing Visual Studio and selecting net framework will install 48. To get other versions you need to explicitly select them from the individual components section.

cgillum:
> For example, installing Visual Studio and selecting net framework will install 48. To get other versions you need to explicitly select them from the individual components section.

That sounds like a good reason to me. Let's do it!

PR #50, opened 2022-09-27
Move all output to root out folder

jviau:
With this PR all output is moved from individual projects folder to a common `out` folder in the repo root. The specific structure is:

- `out/bin/{configuration}/{project-name}/` -> all final binary output (target framework is typically appended as well)
- `out/obj/{project-name}/` -> all intermediate output (configuration may also be appended)
- `out/pkg` -> all package / publish output. ie: nuget packages.
- `out/samples/` -> same structure as previous 3, but for all sample projects. This is so they are distinctly separate.


Directory.Build.props
```diff
@@ -7,10 +7,29 @@
...
+
+  <PropertyGroup>
+    <IsTestProject Condition="$(MSBuildProjectName.Contains('Test'))">true</IsTestProject>
+  </PropertyGroup>
+
   <!-- Signing -->
   <PropertyGroup>
     <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)eng/key.snk</AssemblyOriginatorKeyFile>
+    <AssemblyOriginatorKeyFile>$(EngRoot)build/key.snk</AssemblyOriginatorKeyFile>
```
cgillum, line 32: I wonder if we should consider a directory name other than `build`. Java tooling, for example, uses `build` as the name of an output directory. Not that Java conventions are of any concern to .NET developers generally, but it seems a little strange given my recent experience working with Java.

jviau: How does `targets` sound? And I reverted `key.snk` move to remain in `eng` directly.

cgillum (approved):

PR #52, opened 2022-09-29
Refactor and split assemblies / nuget packages

jviau:
This PR is a significant overhaul of our project layout, and thus assembly and nuget packages. It is focused on simply adding the new projects, removing the old ones, and moving files where they now belong. Some minimal code changes too place as part of this move just to solve some previously internal patterns now needing to be public (with an appropriate change to ensure it is an appropriate public exposure). Further PRs will work on flushing out this rearrangement. Some things to call out:

- `Worker.csproj` currently has minimal files in it. Eventually some common functionality will be moved from `Grpc.Worker.csproj` to here.
- Unit test projects do not have any tests (yet), so a placeholder is there to prevent warnings during `dotnet test`.
- Samples and Generators.Tests are referencing only the old nuget package, as the durable functions extensions nuget ref causes issues. This will be addressed in time.
- Some unused files / methods were removed (`AsyncLock` for example).


samples/Directory.Build.props
```diff
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <SubOutRoot>samples/</SubOutRoot>
+    <IsSampleProject>true</IsSampleProject>
```
cgillum, line 5: Is this property used anywhere?

src/Directory.Build.targets
```diff
@@ -0,0 +1,19 @@
...
+    Condition=" '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)../, $(_DirectoryBuildTargetsFile)))' != '' " />
+
+  <PropertyGroup>
+    <!-- FileVersionRevision is expected to be set by the CI. -->
+    <FileVersion Condition="'$(FileVersionRevision)' != ''">$(VersionPrefix).$(FileVersionRevision)</FileVersion>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(IncludeShared)' != 'false'">
+    <Compile Include="$(SrcRoot)Shared/**/*.cs" LinkBase="Shared" Visible="false" />
+    <EmbeddedResource Include="$(SrcRoot)Shared/**/*.resx" LinkBase="Shared" Visible="false">
```
cgillum, line 13: Do we have resource files or is this for just in case we add them later?

src/Abstractions/TaskOptions.cs
```diff
@@ -28,7 +25,10 @@ internal TaskOptions(Builder builder)
     /// </summary>
     public CancellationToken CancellationToken { get; }
 
-    internal AsyncRetryHandler? RetryHandler { get; }
+    /// <summary>
+    /// Gets the retry handler.
+    /// </summary>
+    public AsyncRetryHandler? RetryHandler { get; }
```
cgillum, line 31: Why did this need to be made public?

src/Client/Core/OrchestrationMetadata.cs
```diff
@@ -48,41 +36,43 @@ public sealed class OrchestrationMetadata
     /// <value>The unique ID of the orchestration instance.</value>
     public string InstanceId { get; }
 
+    public DataConverter? DataConverter { get; init; }
```
cgillum, line 39: We'll need XML doc comments on this new public property.

cgillum: But curious why you decided to make this public?

cgillum (commented):
Added some initial feedback. Haven't gone through the whole PR yet.

One thing: can we remove the {T} suffixes from the file names (I noticed them in the Abstractions project)? I worry about potential problems these special characters could cause on different shells/file systems. My recollection of the .NET Framework code is that they don't include the type parameter names in the file names.

src/Directory.Build.targets
jviau: In case we add them later. I included this because the extra logic here is essential for making shared resx files work - didn't want to lose that. But we should be moving to resx files eventually for localization.

src/Abstractions/Abstractions.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageDescription>Abstractions for the DurableTask Framework (DTFx).</PackageDescription>
```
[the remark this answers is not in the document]
jviau: Got it

src/Abstractions/TaskOptions.cs
jviau: It is used in the worker project. Given `AsyncRetryHandler` is a public delegate, I am not concerned with this being public. But if you have concerns, I will make a note to re-evaluate this later and see if there is a better way to expose this publicly.

src/Client/Core/OrchestrationMetadata.cs
jviau: I had to do some refactoring of this type to remove gRPC dependency and allow for public construction as it is used in the client project. So the client needs to provide the data converter here.

jviau:
> One thing: can we remove the {T} suffixes from the file names (I noticed them in the Abstractions project)? I worry about potential problems these special characters could cause on different shells/file systems. My recollection of the .NET Framework code is that they don't include the type parameter names in the file names.

Sure we can go with something else, but we do need _something_ to identify generic types. What style of generic file naming would you prefer? The conventions I have noticed are:

1. `fileName{T}` -> "stylecop" style https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#file-naming-conventions
2. ``fileName\`1`` -> "metadata" style https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md#file-naming-conventions
3. `fileNameT` -> as seen in dotnet runtime repo
4. `fileNameOfT` -> as seen in dotnet runtime and aspnetcore repo

cgillum:
Why do we need to identify generic types? `List<T>` is contained in [List.cs](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs). This was the recommended practice back in the days of the Framework Design Guidelines - not sure if that has changed.

jviau:
Primary reason is for file name collisions. If I have both `Pageable` and `Pageable<T>` then I need two file names. `Pageable.cs` and `Pageable??.cs`. Then after that is consistency, if we are now naming some generic files differently to avoid collisions, lets name them all differently.

cgillum:
The .NET framework, WCF, and WF always dealt with this by putting all type variations in the same file. So in your example, both `Pageable` and `Pageable<T>` would go into Pageable.cs. Example: [Activity.cs](https://referencesource.microsoft.com/#System.Activities/System/Activities/Activity.cs,587e9a69776429da), which defines both `Activity` and `Activity<TResult>`. But if the latest generation of the .NET development team is using `{T}` in the file name, then I don't have any strong grounds to oppose this. It definitely surprises me, though, since it's a break from a longstanding convention.

jviau:
Hmm okay - we are already doing the style you mention here, I will continue with that. We can change it later if need be.

cgillum (commented):
I finished going through all the changes.

One question I had that isn't specific to a single file: I noticed that the Worker.Grpc project contains a lot of non-gRPC types, like `TaskActivityShim`, `TaskOrchestrationShim`, `WorkerContext`, etc. Why wouldn't those types go in the core Worker project?

src/Worker/Grpc/SdkUtils.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask;
+
+static class SdkUtils
```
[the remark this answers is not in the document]
jviau: This will be one of my follow up items to find a way to nicely share internals between gRPC projects.

src/Client/Grpc/SdkUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+    static class EnvironmentVariables
+    {
+        // All environment variables should be prefixed with DURABLETASK_
+        public const string SidecarHost = "DURABLETASK_SIDECAR_HOST";
+        public const string SidecarPort = "DURABLETASK_SIDECAR_PORT";
+    }
+
+    internal class EmptyServiceProvider : IServiceProvider
+    {
+        public static EmptyServiceProvider Instance => new();
```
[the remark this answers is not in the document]
jviau: Good call! Let me fix that.

src/Client/Grpc/ProtoUtils.cs
```diff
@@ -0,0 +1,111 @@
...
+    internal static P.OrchestrationStatus ToGrpcStatus(this OrchestrationRuntimeStatus status)
+        => status switch
+        {
+            OrchestrationRuntimeStatus.Canceled => P.OrchestrationStatus.Canceled,
+            OrchestrationRuntimeStatus.Completed => P.OrchestrationStatus.Completed,
+            OrchestrationRuntimeStatus.ContinuedAsNew => P.OrchestrationStatus.ContinuedAsNew,
+            OrchestrationRuntimeStatus.Failed => P.OrchestrationStatus.Failed,
+            OrchestrationRuntimeStatus.Pending => P.OrchestrationStatus.Pending,
+            OrchestrationRuntimeStatus.Running => P.OrchestrationStatus.Running,
+            OrchestrationRuntimeStatus.Terminated => P.OrchestrationStatus.Terminated,
```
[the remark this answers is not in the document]
jviau: Okay to leave this for a separate PR? I want to keep functional changes minimal in this PR - so we can verify it still behaves as previously.

src/Abstractions/Abstractions.csproj
```diff
@@ -0,0 +1,23 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.11.1" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
+    <PackageReference Include="System.Collections.Concurrent" Version="4.3.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
+    <PackageReference Include="System.Text.Json" Version="6.0.0" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
jviau: Ah let me check this.

samples/WebAPI/WebAPI.csproj
```diff
@@ -9,15 +9,8 @@
...
-    <ProjectReference Include="$(SrcRoot)DurableTask\Microsoft.DurableTask.csproj" />
-  </ItemGroup>
-
-  <!-- Source generation -->
-  <ItemGroup>
-    <!-- Note that this is not a "normal" ProjectReference.
-         It needs the additional 'OutputItemType' and 'ReferenceOutputAssembly' attributes. -->
-    <ProjectReference Include="$(SrcRoot)DurableTask.Generators\Microsoft.DurableTask.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+    <PackageReference Include="Microsoft.DurableTask.Client" Version="0.4.1-beta" />
+    <PackageReference Include="Microsoft.DurableTask.Generators" Version="0.4.1-beta" />
```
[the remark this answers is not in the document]
jviau: I will be revisiting the sample projects soon. I was running into issues with duplicate types when I had the p2p reference.

samples/ConsoleApp/ConsoleApp.csproj
```diff
@@ -8,16 +8,9 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
+    <PackageReference Include="Microsoft.DurableTask.Client" Version="0.4.1-beta" />
+    <PackageReference Include="Microsoft.DurableTask.Generators" Version="0.4.1-beta" />
```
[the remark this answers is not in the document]
jviau: I will be switching it over to a p2p reference eventually, so no need to manually update it.

jviau:
> One question I had that isn't specific to a single file: I noticed that the Worker.Grpc project contains a lot of non-gRPC types, like TaskActivityShim, TaskOrchestrationShim, WorkerContext, etc. Why wouldn't those types go in the core Worker project?

This is something I will address in another PR. But yes, I definitely want to move it to the Worker project. Just it will require work to detangle the internals and think of a good public surface to expose.

cgillum (approved):
LGTM!

PR #53, opened 2022-09-30
Add SolutionItems.csproj to hold random files (For visiblity in VS)

jviau:


cgillum:
Why is this better than using Visual Studio's built-in feature for referencing random files, which is what we've been using already?

image

jviau:
Main benefit is globs. No need to manually add them to the solution file each time. They can just be dropped in the directory and they will be included.

jviau:
I can rename this "misc" if you want so it looks basically the same in VS

cgillum (approved):
I see. It's a little odd, but I can see how this would be easier to work with compared to the built-in VS feature.  I assume this is a pattern used elsewhere, given that there's a `Microsoft.Build.NoTargets` SDK type?

jviau:
Yes it is odd. The solution file format is kind of a dinosaur and has not had any improvements in a long time. But yeah, this is one of the scenarios the NoTargets SDK was built for. It comes from an MSBuild team member as well.

PR #54, opened 2022-09-30
Add global json, dirs.proj, tweak CI definitions

jviau:
- Introduce `global.json` to pin the dotnet SDK versions we use (this is so there are no unforeseen issues with differences between SDK versions, and "works on my machine" as a result)
- Add `sln/dirs.proj` so we can `dotnet build` for that directory to build just source
- Tweak CI files (github and ADO) to use global.json and dirs.proj
- Addition tweaks to CI:
    - github: add pack and upload steps, so you can grab artifacts from a PR build inspect/verify them before trying the release build. Now also builds release config (to ensure it is as similar to actual release as possible).
    - ADO: replace VSBuild with dotnet build step for consistency

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -18,7 +18,7 @@ steps:
   displayName: 'Use the .NET 6 SDK'
   inputs:
     packageType: 'sdk'
-    version: '6.0.x'
+    useGlobalJson: true
 
 # Start by restoring all the dependencies. This needs to be its own task.
 - task: DotNetCoreCLI@2
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -22,22 +22,19 @@ steps:
 
 # Start by restoring all the dependencies. This needs to be its own task.
 - task: DotNetCoreCLI@2
-  displayName: 'Restore nuget dependencies'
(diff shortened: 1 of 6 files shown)


azure-pipelines-release.yml
```diff
@@ -18,26 +18,23 @@ steps:
...
     command: restore
     verbosityRestore: Minimal
-    projects: 'src/**/*.csproj'
+    projects: 'src/dirs.proj'
 
-# Build the entire solution. This will also build all the tests, which
-# isn't strictly necessary...
-- task: VSBuild@1
+# Build source directory
+- task: DotNetCoreCLI@2
```
cgillum, line 32: Just for my own curiosity, why'd you decide to go with `DotNetCoreCLI` over `VSBuild`?

cgillum (approved):

azure-pipelines-release.yml
jviau: 1) keep it consistent with our other tasks. 2) `VSBuild` can actually have differences with `DotNetCoreCLI`, as they potentially include a different set of targets. By ensuring we use dotnet CLI we won't accidentally take on a dependency to VS-only targets.

PR #55, opened 2022-10-03
Extract DurableTask.Core shims to common Worker package

jviau:
This PR extracts the shims to `Microsoft.DurableTask.Worker` and allows for public usage of them via `DurableTaskShimFactory` abstraction. This change also involved some other refactors:

1. Replaced DurableTask.Core "ParentInstance" usage with our own type.
2. Refactoring of DurableTaskGrpcWorker and its builder to leverage the shim factory.
3. Moving `FuncTaskOrchestrator` to worker package (from abstractions).
4. Renaming of `OrchestrationRunner` to `GrpcOrchestrationRunner` (as it is gRPC specific).


src/Worker/Core/Shims/DurableTaskShimFactory.cs
```diff
@@ -0,0 +1,70 @@
...
+/// <summary>
+/// A shim factory for bridging between 
+/// </summary>
+/// <remarks>
+/// This class is intended for use with alternate .NET-based durable task runtimes. It's not intended for use
+/// in application code.
+/// </remarks>
+public class DurableTaskShimFactory
+{
+    public static readonly DurableTaskShimFactory Default = new();
```
cgillum, line 22: Consider making this a property instead of a field since it's public.

src/Worker/Core/Shims/DurableTaskShimFactory.cs
```diff
@@ -0,0 +1,70 @@
...
+        DataConverter? dataConverter = null,
+        ILoggerFactory? loggerFactory = null,
+        TimerOptions? timerOptions = null)
+    {
+        this.dataConverter = dataConverter ?? JsonDataConverter.Default;
+        this.loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+        this.timerOptions = timerOptions ?? new();
+    }
+
+    public TaskActivity CreateActivity(TaskName name, ITaskActivity activity)
```
cgillum, line 44: XML Doc comments for public members?

src/Worker/Core/Shims/TaskActivityShim.cs
```diff
@@ -3,7 +3,7 @@
 
 using DurableTask.Core;
 
-namespace Microsoft.DurableTask;
+namespace Microsoft.DurableTask.Worker.Shims;
```
cgillum, line 6: Looking at the classes in this file, are they redundant with the `FuncTaskActivity` code you're adding?

src/Worker/Core/Shims/TaskOrchestrationShim.cs
```diff
@@ -14,44 +15,38 @@ namespace Microsoft.DurableTask;
...
-        JsonDataConverterShim converterShim = new(this.workerContext.DataConverter);
+        JsonDataConverterShim converterShim = new(this.invocationContext.DataConverter);
         innerContext.MessageDataConverter = converterShim;
         innerContext.ErrorDataConverter = converterShim;
 
-        object? input = this.workerContext.DataConverter.Deserialize(rawInput, this.implementation.InputType);
+        object? input = this.invocationContext.DataConverter.Deserialize(rawInput, this.implementation.InputType);
 
-        this.wrapperContext = new(innerContext, this.name, this.workerContext, this.runtimeState, input);
+        ILogger contextLogger = this.invocationContext.LoggerFactory.CreateLogger("Microsoft.DurableTask");
```
cgillum, line 44: Previously we encapsulated the creation of loggers inside `SdkUtils.GetLogger(loggerFactory)`. Is there a reason to not do that here?

jviau: Don't have `SdkUtils` here. I also plan on refactoring that later. We should not be using a single category, but rather multiple categories. We just have to make sure they all start with `Microsoft.DurableTask`. So this one I actually want to change to `.CreateLogger<TaskOrchestrationContext>()` (in another PR).

src/Worker/Core/Shims/DurableTaskShimFactory.cs
```diff
@@ -0,0 +1,103 @@
...
+using Microsoft.DurableTask.Converters;
+using Microsoft.DurableTask.Options;
+using Microsoft.DurableTask.Shims;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Microsoft.DurableTask.Worker.Shims;
+
+/// <summary>
+/// A shim factory for bridging between
```
cgillum, line 14: Unfinished comment

cgillum (approved):
Just one minor comment note, but otherwise LGTM! It's really nice to see the structure of the project take a more mature shape like this.

PR #56, opened 2022-10-04
Enable NET analyzers

jviau:
This PR enables .NET6 recommended analyzers. Additionally, it marks `TaskName` as a `readonly` struct.


cgillum (approved):
LGTM! It seems I need to pay more attention to the tools the .NET team has been making available. I didn't know about these analyzers.

PR #57, opened 2022-10-04
Add stylecop analyzers - enable for abstractions project only (for now)

jviau:
This PR adds [StyleCop Analyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers), allowing us to enforce code style at build. To start, only the `Abstractions` project has been onboarded (and all warnings addressed). Future PRs will enable other projects one by one.


src/Abstractions/TaskFailureDetails.cs
```diff
@@ -34,7 +35,8 @@ public override string ToString()
     /// </remarks>
     /// <typeparam name="T">The type of exception to test against.</typeparam>
     /// <returns>Returns <c>true</c> if the <see cref="ErrorType"/> value matches <typeparamref name="T"/>; <c>false</c> otherwise.</returns>
-    public bool IsCausedBy<T>() where T : Exception
+    public bool IsCausedBy<T>()
+        where T : Exception
```
cgillum, line 39: I'm not sure I like this rule. I feel like it's nicer to have generic constraints on the same line as the method signature. Thoughts?

.editorconfig
```diff
@@ -210,4 +210,9 @@ csharp_style_prefer_method_group_conversion = true:silent
...
 tab_width = 4
 indent_size = 4
-end_of_line = crlf
\ No newline at end of file
+end_of_line = crlf
+
+# StyleCop rules
+dotnet_diagnostic.SA1400.severity = none  # Access modifier must be declared
+dotnet_diagnostic.SA1402.severity = none  # File may only contain a single type
+dotnet_diagnostic.SA1649.severity = none  # File name must match type name
```
cgillum, line 218: Should we put the file header rule in .editorconfig too? I believe VS knows to automatically add the header for new files when you do this, which is really nice when accepting PRs from external contributors.

cgillum (approved):

.editorconfig
jviau: We have duplicate file header enforcement rules. IDE0073 and SA1633. I have set IDE0073 as a suggestion for now (and .editorconfig already had `file_header-template` configured). I am leaving in SA1633 as a warning while projects get onboarded one by one. Eventually though I will switch IDE0073 to warning, enable enforce style on build, and then turn off SA1633.

src/Abstractions/TaskFailureDetails.cs
jviau: Suppressed this rule. I like having it on a new line, but don't feel strongly about it - so we can leave it to a case by case preference.

PR #61, opened 2022-10-13
Add client builder pattern and gRPC implementation

jviau:
This PR mirrors almost exactly what the worker builder PR is doing, but for the client. These builders are kept separate as we believe it will be a common use case to want one, but not the other. For example, you may want just the client in your Web API for interacting with orchestrations, but no need to reference the worker there. It _may_ be more common for workers to also want a client, but for the time being we will require registering the client separately. In the future we can consider making a Worker -> Client dependency and letting the worker register a pairing client. There is nothing stopping an individual implementation from doing this themselves. For example, a `Microsoft.DurableTask.Worker.InMemory` implementation would want to reference both client and worker packages, and `IDurableTaskBuilder.UseInMemory()` would register both worker and client.

Some distinctions with worker PR:

1. Clients are fetched from the service container via `IDurableTaskClientProvider`. This is because there may be _multiple_ named clients available and so direct DI would not be able to know which client you want - thus the extra hop of `IDurableTaskClientProvider.GetClient(string name)`.
2. In scenarios where you know you will have only a single client, you can call `IDurableTaskClientBuilder.RegisterDirectly()` to inject that `DurableTaskClient` as a singleton directly into the provider.

# Example

``` CSharp
IHost host = Host.CreateDefaultBuilder(args)
    .ConfigureServices(services =>
    {
        services.AddDurableTaskClient(builder =>
        {
            builder.Configure(x => x.DataConverter = new MyDataConverter());
            builder.UseGrpc(/* address of gRPC sidecar */);
            builder.RegisterDirectly(); // if you want to DI this client directly
        });
    });
```


src/Client/Grpc/DependencyInjection/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+/// </summary>
+static class ServiceProviderExtensions
+{
+    /// <summary>
+    /// Gets the current options from the service provider.
+    /// </summary>
+    /// <typeparam name="TOptions">The option type to get.</typeparam>
+    /// <param name="serviceProvider">The service provider.</param>
+    /// <param name="name">The name of the options. <c>null</c> for default name.</param>
+    /// <returns>The options, if available. Null otherwise.</returns>
```
cgillum, line 20: nit: For consistency with the previous line, Null --> `<c>null</c>`

test/Grpc.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -107,4 +94,31 @@ protected IReadOnlyCollection<LogEntry> GetLogs()
         Assert.True(foundCategory);
         return logs;
     }
+
+    protected struct HostLifetime : IAsyncDisposable
```
cgillum, line 98: I'm curious how you came up with the name `HostLifetime`? It confused me because of how similar it is to [IHostLifetime](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.ihostlifetime?view=dotnet-plat-ext-6.0), which I think is unrelated.

src/Client/Core/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+    {
+        readonly Dictionary<string, IDurableTaskClientBuilder> builders = new();
+        readonly IServiceCollection services;
+
+        public BuilderContainer(IServiceCollection services)
+        {
+            this.services = services;
+        }
+
+        public IDurableTaskClientBuilder Get(string name, out bool added)
```
cgillum, line 96: Maybe rename to `GetOrAdd`?

cgillum (approved):
These changes look good to me. I can't remember if we discussed this to a conclusion yet, but I'm curious how this impacts the samples. I was surprised that this PR doesn't include any changes to sample apps?

test/Grpc.IntegrationTests/IntegrationTestBase.cs
jviau: Being test code, I wasn't too concerned with the most ideal name. Just wanted something to represent the lifetime of the host within this test.

PR #65, opened 2022-10-14
Add and consume validation helpers, remove unused interface

jviau:
Adds some validation helpers to simplify argument verification, leveraging the `CallerArgumentExpressionAttribute`. Also removes `IDurableTaskRegistry` as that is no longer used - the concrete options type in Worker.Core.csproj is used now.


cgillum:
Looks like there's a test failure:
```
[xUnit.net 00:00:01.15]     Microsoft.DurableTask.Worker.Tests.DefaultDurableTaskBuilderTests.BuildTarget_ValidType_Sets [FAIL]
  Failed Microsoft.DurableTask.Worker.Tests.DefaultDurableTaskBuilderTests.BuildTarget_ValidType_Sets [18 ms]
  Error Message:
   System.ArgumentNullException : Value cannot be null. (Parameter 'value')
  Stack Trace:
     at Microsoft.DurableTask.Check.NotNull[T](T argument, String name) in /home/runner/work/durabletask-dotnet/durabletask-dotnet/src/Shared/Validation/Check.cs:line 67
   at Microsoft.DurableTask.Check.ConcreteType[TImplements](Type argument, String name) in /home/runner/work/durabletask-dotnet/durabletask-dotnet/src/Shared/Validation/Check.cs:line 107
   at Microsoft.DurableTask.Worker.DefaultDurableTaskBuilder.set_BuildTarget(Type value) in /home/runner/work/durabletask-dotnet/durabletask-dotnet/src/Worker/Core/DependencyInjection/DefaultDurableTaskBuilder.cs:line [40](https://github.com/microsoft/durabletask-dotnet/actions/runs/3253189718/jobs/5340212656#step:7:41)
   at Microsoft.DurableTask.Worker.Tests.DefaultDurableTaskBuilderTests.BuildTarget_ValidType_Sets() in /home/runner/work/durabletask-dotnet/durabletask-dotnet/test/Worker/Core.Tests/DependencyInjection/DefaultDurableTaskBuilderTests.cs:line 28
```

src/Abstractions/TaskActivityBase.cs
```diff
@@ -94,6 +94,7 @@ public abstract class TaskActivityBase<TInput, TOutput> : ITaskActivity
     Exception DefaultNotImplementedException()
     {
         return new NotImplementedException(
-            $"{this.GetType().Name} needs to override {nameof(this.OnRun)} or {nameof(this.OnRunAsync)} with an implementation.");
+            $"{this.GetType().Name} needs to override {nameof(this.OnRun)} or {nameof(this.OnRunAsync)} with an"
+            + " implementation.");
```
cgillum, line 98: Part of me feels like we're following this line length guideline a bit too religiously. I think we should consider it more of a guide to help keep the code readable. Changes like these, however, where we introduce a newline for a single word seem to make the code less readable, IMO. Thoughts?

cgillum (approved):

src/Abstractions/TaskActivityBase.cs
jviau: Yeah that is fair - I will undo those changes.

jviau: I undid some of them. If the last word/statement crosses 120 I will leave it be. If an entirely new word starts past 120, then I move everything 120 and up to the new line.

PR #67, opened 2022-10-20
Extract IDurableTaskFactory, enclose task invocations with a service scope.

jviau:
This PR does the following:

1. Extracts a contract `IDurableTaskFactory` and makes the existing `DurableTaskFactory` implementation `internal`.
    - The goal here is two-fold: 1) hide our internal implementation and improve unit testability. and 2) to set us up for user supplied and aggregate factories (post GA, will need more design/thinking).
2. Adds `IServiceProvider` argument to `ITaskOrchestrator` construction. However, **it is not used** at the moment.
    - This is to make the structure consistent with `ITaskActivity` construction.
    - We can use this in the future to provide a _limited and customized_ set of services to `ITaskOrchestrator` construction (replay safe `ILogger`, `IOptions<T>` which have been snapshotted in place as a history event).
3. Encloses `ITaskOrchestrator` and `ITaskActivity` invocations in a new service scope. This enables use of scoped services for invocations.

### DIFF
--- a/src/Worker/Core/DurableTaskFactory.cs
+++ b/src/Worker/Core/DurableTaskFactory.cs
@@ -8,10 +8,10 @@ namespace Microsoft.DurableTask.Worker;
 /// <summary>
 /// A factory for creating orchestrators and activities.
 /// </summary>
-public sealed class DurableTaskFactory
+sealed class DurableTaskFactory : IDurableTaskFactory
 {
     readonly IReadOnlyDictionary<TaskName, Func<IServiceProvider, ITaskActivity>> activities;
-    readonly IReadOnlyDictionary<TaskName, Func<ITaskOrchestrator>> orchestrators;
+    readonly IReadOnlyDictionary<TaskName, Func<IServiceProvider, ITaskOrchestrator>> orchestrators;
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


cgillum:
From my own education, can you provide an example of this: _This enables use of scoped services for invocations._?  I'm not familiar with how scoped services work and I'm curious what the key scenarios are.

cgillum (approved):
These code changes look good to me. I just had the question about how the scoped service providers will be used.

jviau:
This gives a good overview: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-6.0

The gist of it is you have 3 service lifetimes:

1. Singleton - only 1 instance of this service, all imports get the same copy
2. Transient - every import is a different copy
3. Scoped - 1 instance _per_ scope. This lets you share copies of a service within a given scope, but not globally.

The best example of scoped services in AspNetCore requests. Every request is its own service scope and thus you can share a scoped service import between different parts of the request processing pipeline via DI practices.

With this PR, orchestration and activities are invoked in their own service scope, so users can register/inject scoped services for consumption there. Without the scope, the DI container may throw on import of a scoped service, as the root container typically does not support importing scoped services (because then they leak into a singleton lifetime).

PR #70, opened 2022-10-25
Refactor DurableTaskRegistry, move bulk of methods to extensions

jviau:
~This PR refactors the `DurableTaskRegistry` to have the core implementation be the minimal required functionality. Then all of the other functionalities are moved to extension methods. The reason I like this approach is it forces a code "funnel", where we can force consistency in behavior by having a minimal API surface on the core type, and then all extension methods must go through this API surface.~

This PR refactors `DurableTaskRegistry` to ensure all registration cases are covered and organizes them into partial classes for activities/orchestrators. The "core" functions are left in the root class file.


cgillum:
I'm not opposed to this change, but I'll admit I don't understand what the benefit is. In my experience, extension methods are confusing because 1) they can't be unit tested directly (doesn't apply here since this is a sealed class) and 2) it makes searching through the code harder. You mentioned the ability to ensure all methods funnel into a few core ones, but why is it beneficial to create self-imposed restrictions like this?

jviau:
That is fair. Would you rather these remain as instance methods?

cgillum:
That would be my preference, mainly as a means of prioritizing code simplicity.

jviau:
Done - they are now instance methods.

src/Worker/Core/DurableTaskRegistry.cs
```diff
@@ -2,176 +2,73 @@
...
+    public DurableTaskRegistry AddOrchestrator(TaskName name, Func<ITaskOrchestrator> factory)
     {
         Check.NotDefault(name);
-        Check.NotNull(implementation);
-        if (this.orchestratorsBuilder.ContainsKey(name))
+        Check.NotNull(factory);
+        if (this.activitiesBuilder.ContainsKey(name))
         {
-            throw new ArgumentException($"A task orchestrator named '{name}' is already added.", nameof(name));
+            throw new ArgumentException($"A TaskOrchestrator named '{name}' is already added.", nameof(name));
```
cgillum, line 68: If we're going to use actual class names, then we should probably use `nameof` instead of hardcoding them.

src/Worker/Core/DurableTaskRegistry.cs
```diff
@@ -2,176 +2,73 @@
...
 
 namespace Microsoft.DurableTask.Worker;
 
 /// <summary>
 /// Options for the Durable Task worker.
 /// </summary>
-public sealed class DurableTaskRegistry
+public sealed partial class DurableTaskRegistry
 {
+    static readonly Task<object?> CompletedNullTask = Task.FromResult<object?>(null);
```
cgillum, line 13: Not used?

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
```diff
@@ -136,7 +136,7 @@ public async Task RetryActivityFailures(int expectedNumberOfAttempts)
                 {
                     await ctx.CallActivityAsync("Foo", options: retryOptions);
                 })
-                .AddActivity("Foo", context =>
+                .AddActivity("Foo", (TaskActivityContext context) =>
```
cgillum, line 139: Curious: was this change caused by an overload ambiguity introduced in this PR?

jviau: Yes, exactly that. There is ambiguity between `Func<IServiceProvider, ITaskActivity>` and `Action<TaskActivityContext>` - this is because it throws an exception and return value cannot be inferred. Only a minor inconvenience because I imagine customers will typically have a return value (which will help resolve the correct overload).

src/Worker/Core/DurableTaskRegistry.cs
jviau: Done

test/Worker/Core.Tests/DurableTaskRegistryTests.Activities.cs
```diff
@@ -0,0 +1,190 @@
...
+
+namespace Microsoft.DurableTask.Worker.Tests;
+
+public partial class DurableTaskRegistryTests
+{
+    [Fact]
+    public void AddActivity_DefaultName_Throws()
+    {
+        DurableTaskRegistry registry = new();
+        Action act = () => registry.AddActivity(default, (IServiceProvider _) => new TestActivity());
```
jviau, line 12: After adding all these overloads (which I think having all of them is great), automatic overload resolution is not great anymore. It is actually resolving _incorrect_ overloads. I would like to improve this to prevent accidentally using the wrong overload. I am thinking the best route is to differentiate some of these by name. At least different names for the lower level _factory_ method and the lambda func-implementation helpers. @cgillum any preference or thoughts on naming?

cgillum: oi...naming is hard. I think the principle to follow is to ensure that users can easily discover the method that they need, so having clarity and consistency is key. For example, if you want to rename the func-implementation helpers, then `AddActivityFunc` and `AddOrchestratorFunc` seem like good choices. They have the same prefixes and are sufficiently descriptive, IMO.

test/Abstractions.Tests/TaskNameTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        act.Should().ThrowExactly<ArgumentException>().WithParameterName("name");
+    }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("")]
+    public void Conversion_EqualToDefault(string? str)
+    {
+        TaskName name = str!;
+        name.Should().Be(default(TaskName));
```
cgillum, line 28: It seems a little inconsistent that you can create a null/empty `TaskName` using the implicit operator but not explicitly using the constructor.

jviau: Fair - would you like to allow non-null/empty values in the constructor? Or to throw in the implicit operator?

test/Worker/Core.Tests/DurableTaskRegistryTests.Activities.cs
jviau: Those names sound great to me! Will update with that. Thank you.

test/Abstractions.Tests/TaskNameTests.cs
cgillum: I'm thinking it makes sense to throw in the implicit operator. However, I can also see an argument against that which is that implicit operators shouldn't throw (I don't know if that guideline exists, but it does seem strange that this is possible). So maybe we should just allow empty values and put the validation burden on the consumers of `TaskName`?

jviau: Yes, there is a guideline to never throw in an implicit operator. I went with the 2nd option but had to do some slightly odd stuff that goes against explicit nullability. I had to get a null name match up to the default struct - which means version must also be null. I think this is acceptable for now, but may get ugly when version can be explicitly set.

cgillum (approved):

PR #71, opened 2022-10-25
Remove setting of TaskName.Version for now

jviau:
Due to the fact we do not consider `TaskName.Version` as part of equality checks, it is effectively unusable to customers. You cannot register two tasks with the same name, but different versions. Additionally incoming work requests of the same name will always resolve to the same instance regardless of version matching. I am opting to remove the ability to set this for now as it is misleading to customers. We can add it back in when task versioning is actually supported.

### DIFF
--- a/src/Abstractions/TaskName.cs
+++ b/src/Abstractions/TaskName.cs
@@ -14,11 +14,10 @@ public readonly struct TaskName : IEquatable<TaskName>
     /// Initializes a new instance of the <see cref="TaskName"/> struct.
     /// </summary>
     /// <param name="name">The name of the task.</param>
-    /// <param name="version">The version of the task, if applicable.</param>
-    public TaskName(string name, string? version = null)
+    public TaskName(string name)
     {
-        this.Name = name ?? throw new ArgumentNullException(nameof(name));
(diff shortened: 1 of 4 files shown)


src/Abstractions/TaskName.cs
```diff
@@ -33,7 +32,8 @@ public TaskName(string name, string? version = null)
     /// Gets the version of the task.
     /// </summary>
     /// <remarks>
-    /// Task versions are currently experimental and their role may change over time.
+    /// Task versions is currently locked to <see cref="string.Empty" /> as it is not yet integrated into task
+    /// identification. This is being left here as we intend to support it soon.
     /// </remarks>
     public string Version { get; }
```
cgillum, line 38: Would it make sense to put an `[Obsolete]` attribute on this field or do we still reference it in some places?

cgillum (approved):

src/Abstractions/TaskName.cs
jviau: We do reference it. I don't want to add `[Obsolete]` because I don't want to discourage logic from taking an actual dependency on it one day not being `""`.

PR #77, opened 2022-11-10
Extract interface slice from client for code-gen

jviau:
This PR extracts a very small interface from `DurableTaskClient` into the Abstractions layer to enable code-generation. Instead of moving all the client types over to abstractions, this smaller contract was chosen to be extracted as this is the only method code-generation will need for now. This interface is public, but lives in the `Microsoft.DurableTask.Internal` namespace and has a disclaimer telling customers to not use it directly.


cgillum (approved):
:shipit:

PR #78, opened 2022-11-10
Update Samples (Console, Netfx, Functions)

jviau:
Updates the ConsoleApp and NetFxConsoleApp samples to the latest way of using the packages.


samples/ConsoleApp/Program.cs
```diff
@@ -1,46 +1,47 @@
...
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Worker;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+IHost host = Host.CreateDefaultBuilder(args)
+    .ConfigureServices(services =>
     {
-        tasks.AddOrchestrator("HelloSequence", async context =>
+        services.AddDurableTaskClient(builder =>
```
jviau, line 12: Right now, client and worker must be registered independently. I am not against worker implementations registering their corresponding client automatically. But I'd like to gather feedback on that before making that choice for the gRPC worker.

samples/ConsoleApp/Program.cs
```diff
@@ -1,46 +1,47 @@
...
         {
-            var greetings = new List<string>
-            {
-                await context.CallActivityAsync<string>("SayHello", "Tokyo"),
-                await context.CallActivityAsync<string>("SayHello", "London"),
-                await context.CallActivityAsync<string>("SayHello", "Seattle"),
-            };
+            builder.UseGrpc();
+            builder.RegisterDirectly();
```
jviau, line 15: This line registers the `DurableTaskClient` directly to the service collection. I am wondering if we should register the "Default" (no name provided) client to the service container?

cgillum: I'm afraid I don't understand the type architecture well enough to understand the question. I also worry customers might struggle to understand like I am. The call to `host.Services.GetRequiredService<DurableTaskClient>()` makes me a tiny bit uneasy since it's not particularly discoverable, but I think it's okay. I do worry about this `builder.RegisterDirectly()` requirement though. I would never think to call that. Could be make this the default configuration?

samples/ConsoleApp/Program.cs
cgillum: I feel like the common case would be registering both, and that registering one or the other would be the advanced case. Could we simply have an `AddDurableTask` extension method that registers both? Or is that complicated because we've put them in two separate packages with no top-level package to give developers everything?

samples/ConsoleApp/ConsoleApp.csproj
```diff
@@ -7,9 +7,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
-    <PackageReference Include="Microsoft.DurableTask.Client" Version="0.4.1-beta" />
```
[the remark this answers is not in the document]
jviau: Sure - I will introduce the commented-out package references when we release the next NuGet package

samples/ConsoleApp/Program.cs
```diff
@@ -1,46 +1,68 @@
...
-{
-    builder.SetMinimumLevel(LogLevel.Debug);
-    builder.AddSimpleConsole(options =>
+IHost host = Host.CreateDefaultBuilder(args)
+    .ConfigureServices(services =>
     {
-        options.SingleLine = true;
-        options.UseUtcTimestamp = true;
-        options.TimestampFormat = "yyyy-mm-ddThh:mm:ss.ffffffZ ";
-    });
```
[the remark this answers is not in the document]
jviau: `Host.CreateDefaultBuilder()` includes a bunch of default logging setup.

samples/ConsoleApp/Program.cs
jviau: #81 partially addresses this - removing the need for `RegisterDirectly` in this case. I will update this PR to reflect that.

samples/ConsoleApp/Program.cs
jviau: I agree the common case will probably be registering both _for worker scenarios_ only. I think it will also be common to want to register only the client (e.g., Web API with client, but worker is in a separate console app). I think in the future we should explore `AddDurableTaskWorker` also registering the corresponding client, but yes, the package separation complicates it. I am open to adding a Worker --> Client package reference, but it will require some design around integrating the calls together. I think this can be left to post-GA.

samples/ConsoleApp/Program.cs
cgillum: Are we able to remove this now?

cgillum (approved):

PR #79, opened 2022-11-10
Update code generators for recent changes

jviau:
This PR updates all non-functions code generation to accommodate the recent changes made. The output code is mostly the same, the only meaningful difference is how nullability is handled. If a type is nullable, then `input` is optional and default value is used. Otherwise, input is required. This requires parameter re-ordering, which means these extension methods deviate from the structure of the stringly-typed ones, namely `instanceId` and `input` order is swapped. I will have a follow-up PR (#80) to make this consistent.

Functions unit tests have been skipped in this PR as they will not be addressed until the durable functions SDK is updated with these latest changes.

** UPDATE **
New changes:
- `Microsoft.DurableTask.Generators` is now an optional package.
  - version changed to `1.0.0-preview.1`
  - no longer automatically brought in when referencing other packages.
- `GeneratedDurableTaskExtensions.AddAllGeneratedTasks` made `internal` (from `public`)
  - This is also to avoid conflicts when multiple files have this method generated. When wanting to expose to external consumes, a new extension method can be manually authored in the desired namespace and with an appropriate name which calls this method.

### DIFF
--- a/samples/WebAPI/Controllers/OrderProcessingController.cs
+++ b/samples/WebAPI/Controllers/OrderProcessingController.cs
@@ -32,8 +32,7 @@ public class OrderProcessingController : ControllerBase
         // Generate an order ID and start the order processing workflow orchestration
         string orderId = $"{orderInfo.Item}-{Guid.NewGuid().ToString()[..4]}";
         await this.durableTaskClient.ScheduleNewProcessOrderOrchestratorInstanceAsync(
-            instanceId: orderId,
-            input: orderInfo);
+            orderInfo, new() { InstanceId = orderId });
 
         // Return 202 with a link to the GetOrderStatus API
         return this.AcceptedAtAction(
--- a/samples/WebAPI/Controllers/OrderProcessingController.cs
+++ b/samples/WebAPI/Controllers/OrderProcessingController.cs
@@ -32,7 +32,7 @@ public class OrderProcessingController : ControllerBase
         // Generate an order ID and start the order processing workflow orchestration
         string orderId = $"{orderInfo.Item}-{Guid.NewGuid().ToString()[..4]}";
         await this.durableTaskClient.ScheduleNewProcessOrderOrchestratorInstanceAsync(
-            orderInfo, new() { InstanceId = orderId });
+            orderInfo, new StartOrchestrationOptions() { InstanceId = orderId });
 
         // Return 202 with a link to the GetOrderStatus API
         return this.AcceptedAtAction(
--- a/samples/WebAPI/Controllers/OrderProcessingController.cs
+++ b/samples/WebAPI/Controllers/OrderProcessingController.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
 using WebAPI.Models;
 
 namespace WebAPI.Controllers;
--- a/samples/WebAPI/Orchestrations/ProcessOrderOrchestrator.cs
+++ b/samples/WebAPI/Orchestrations/ProcessOrderOrchestrator.cs
@@ -7,9 +7,9 @@ using WebAPI.Models;
 namespace WebAPI.Orchestrations;
 
 [DurableTask]
-public class ProcessOrderOrchestrator : TaskOrchestratorBase<OrderInfo, OrderStatus>
+public class ProcessOrderOrchestrator : TaskOrchestrator<OrderInfo, OrderStatus>
 {
-    protected override async Task<OrderStatus?> OnRunAsync(TaskOrchestrationContext context, OrderInfo? orderInfo)
+    public override async Task<OrderStatus> RunAsync(TaskOrchestrationContext context, OrderInfo?orderInfo)
     {
         if (orderInfo == null)
         {
--- a/samples/WebAPI/Orchestrations/ProcessOrderOrchestrator.cs
+++ b/samples/WebAPI/Orchestrations/ProcessOrderOrchestrator.cs
@@ -9,7 +9,7 @@ namespace WebAPI.Orchestrations;
 [DurableTask]
 public class ProcessOrderOrchestrator : TaskOrchestrator<OrderInfo, OrderStatus>
 {
-    public override async Task<OrderStatus> RunAsync(TaskOrchestrationContext context, OrderInfo?orderInfo)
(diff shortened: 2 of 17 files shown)


samples/WebAPI/Controllers/OrderProcessingController.cs
```diff
@@ -31,8 +32,7 @@ public async Task<ActionResult> CreateOrder([FromBody] OrderInfo orderInfo)
         // Generate an order ID and start the order processing workflow orchestration
         string orderId = $"{orderInfo.Item}-{Guid.NewGuid().ToString()[..4]}";
         await this.durableTaskClient.ScheduleNewProcessOrderOrchestratorInstanceAsync(
-            instanceId: orderId,
-            input: orderInfo);
+            orderInfo, new() { InstanceId = orderId });
```
cgillum, line 35: Do I recall correctly that you had a `SubOrchestratorOptions.WithInstanceId(...)` or similar method? If so, that might be a nicer snippet to use here to make the code a bit more readable, and to make it more obvious that there are other options that can be configured. I like the `new()` pattern, but only when the type it refers to is obvious, which isn't the case here.

samples/WebAPI/Orchestrations/ProcessOrderOrchestrator.cs
```diff
@@ -7,9 +7,9 @@
 namespace WebAPI.Orchestrations;
 
 [DurableTask]
-public class ProcessOrderOrchestrator : TaskOrchestratorBase<OrderInfo, OrderStatus>
+public class ProcessOrderOrchestrator : TaskOrchestrator<OrderInfo, OrderStatus>
 {
-    protected override async Task<OrderStatus?> OnRunAsync(TaskOrchestrationContext context, OrderInfo? orderInfo)
+    public override async Task<OrderStatus> RunAsync(TaskOrchestrationContext context, OrderInfo?orderInfo)
```
cgillum, line 12: Are you missing a space between `OrderInfo?` and `orderInfo` here?

samples/WebAPI/Program.cs
```diff
@@ -3,11 +3,19 @@
...
 
 // Add all the generated tasks
-builder.Services.AddDurableTask(taskRegistry => taskRegistry.AddAllGeneratedTasks());
+builder.Services.AddDurableTaskWorker(builder =>
+{
+    builder.AddTasks(r => r.AddAllGeneratedTasks());
+    builder.UseGrpc();
+});
+
+builder.Services.AddDurableTaskClient(b => b.UseGrpc().RegisterDirectly());
```
cgillum, line 18: Is the `.RegisterDirectly()` call still required after some of your more recent changes?

samples/WebAPI/WebAPI.csproj
```diff
@@ -9,8 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.DurableTask.Client" Version="0.4.1-beta" />
-    <PackageReference Include="Microsoft.DurableTask.Generators" Version="0.4.1-beta" />
+    <ProjectReference Include="$(SrcRoot)Worker/Grpc/Worker.Grpc.csproj" />
+    <ProjectReference Include="$(SrcRoot)Client/Grpc/Client.Grpc.csproj" />
+    <ProjectReference Include="$(SrcRoot)Generators/Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
cgillum, line 14: Previously the generated was bundled with the SDK nuget package. Any particular reason to separate it out in your refactor?

test/Generators.Tests/ClassBasedSyntaxTests.cs
```diff
@@ -10,55 +10,46 @@ public class ClassBasedSyntaxTests
     const string GeneratedClassName = "GeneratedDurableTaskExtensions";
     const string GeneratedFileName = $"{GeneratedClassName}.cs";
 
-    [Fact]
-    public Task Orchestrators_PrimitiveTypes()
+    [Theory]
+    [InlineData("int", "int input")]
+    [InlineData("int?", "int? input = default")]
+    [InlineData("string", "string input")]
+    [InlineData("string?", "string? input = default")]
```
cgillum, line 17: I like these permutations you added

test/Generators.Tests/AzureFunctionsTests.cs
```diff
@@ -11,7 +11,7 @@ public class AzureFunctionsTests
     const string GeneratedClassName = "GeneratedDurableTaskExtensions";
     const string GeneratedFileName = $"{GeneratedClassName}.cs";
 
-    [Fact]
+    [Fact(Skip = "Durable Functions Extension out of date")]
```
cgillum, line 14: Let's open a GitHub issue to track re-enabling these.

samples/WebAPI/Controllers/OrderProcessingController.cs
jviau: That is for sub orchestrations, this one would be `StartOchestrationOptions`, which does not inherit from the same `TaskOptions`.

But yes, I can make the type explicit.

samples/WebAPI/Program.cs
jviau: It is not, removed.

samples/WebAPI/WebAPI.csproj
jviau: It is bundled with it, but that bundling only works property via `PackageReference`. P2p reference needs to be explicit.

test/Generators.Tests/AzureFunctionsTests.cs
jviau: Done: #85

samples/WebAPI/Program.cs
```diff
@@ -2,9 +2,9 @@
 // Licensed under the MIT License.
 
 using System.Text.Json.Serialization;
-using Microsoft.DurableTask;
 using Microsoft.DurableTask.Client;
 using Microsoft.DurableTask.Worker;
+using WebAPI;
```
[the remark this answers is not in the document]
jviau: It is because I changed the namespace of `GeneratedDurableTaskExtensions` to be the project root namespace, and this file does not actually have a namespace (minimal APIs).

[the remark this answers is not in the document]
cgillum: Ah, I see that you updated the PR description with more context.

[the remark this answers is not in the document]
jviau: Undid this change.

cgillum (approved):
LGTM!

PR #80, opened 2022-11-11
Refactor TaskOptions, wrap InstanceId into them

jviau:
A problem that arose while working on code-generation was parameter ordering inconsistencies between the stringly-typed methods and generated methods. This PR aims to improve this, while also setting up our APIs to be able to expand input options without breaking changes.

### Summary of Changes:

- Refactor `TaskOptions` to be a record type
    - records `with { }` operator is a perfect fit for these options.
- Removed `CancellationToken` from `TaskOptions`. We can add it in without breaking anything when we do intend to support it.
    - I also removed it because I am unsure if having cancellation as a property on the options is the correct choice, or if it should be its own parameter to mesh with cancellation token practices better.
- Pulled the retry options into their own type that uses implicit operators to give a union-esque behavior.
- Added `OrchestrationOptions`, a derived `TaskOptions` which allows for specifying instance ID.
    - `TaskOrchestrationContext` sub-orchestration overloads which take in instance ID have been removed. Instead, you pass in the derived options type and the implementation will use `GetInstanceId()` to see if a specified instance ID was supplied.
- A similar approach was taken for `DurableTaskClient.ScheduleNewOrchestrationInstanceAsync`. Instance ID and start-time are now wrapped into a single `StartOrchestrationOptions` object.

### DIFF
--- a/src/Abstractions/Internal/IOrchestrationSubmitter.cs
+++ b/src/Abstractions/Internal/IOrchestrationSubmitter.cs
@@ -23,21 +23,15 @@ public interface IOrchestrationSubmitter
     /// doing so can result in application failures when updating to a new DurableTask release.
     /// </summary>
     /// <param name="orchestratorName">The name of the orchestrator to schedule.</param>
-    /// <param name="instanceId">
-    /// The unique ID of the orchestration instance to schedule. If not specified, a randomGUID value is used.
-    /// </param>
     /// <param name="input">
     /// The optional input to pass to the scheduled orchestration instance. This must be a serializable value.
     /// </param>
-    /// <param name="startTime">
-    /// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
-    /// is specified, the orchestration instance will be scheduled immediately.
-    /// </param>
+    /// <param name="options">The options to start the new orchestration with.</param>
     /// <returns>
     /// A task that completes when the orchestration instance is successfully scheduled. The value of this task is
-    /// the instance ID of the scheduled orchestration instance. If a non-null <paramref name="instanceId"/> parameter
-    /// value was provided, the same value will be returned by the completed task.
+    /// the instance ID of the scheduled orchestration instance. If a non-null instance ID was provided via
+    /// <paramref name="options" />, the same value will be returned by the completed task.
     /// </returns>
     Task<string> ScheduleNewOrchestrationInstanceAsync(
-        TaskName orchestratorName, string? instanceId = null, object? input = null, DateTimeOffset? startTime = null);
+        TaskName orchestratorName, object? input = null, StartOrchestrationOptions? options = null);
 }
--- a/src/Abstractions/TaskOptions.cs
+++ b/src/Abstractions/TaskOptions.cs
@@ -41,54 +41,44 @@ public record TaskOptions
     /// </summary>
     /// <param name="handler">The handler to convert from.</param>
     /// <returns>A <see cref="TaskOptions" /> built from the handler.</returns>
-    public static TaskOptions FromRetryHandler(RetryHandler handler)
-    {
-        Check.NotNull(handler);
-        return FromRetryHandler(context => Task.FromResult(handler.Invoke(context)));
-    }
+    public static TaskOptions FromRetryHandler(RetryHandler handler) => new(handler);
 
     /// <summary>
-    /// Returns a new <see cref="OrchestrationOptions" /> with the provided instance ID. This can be used when starting
-    /// a new sub-orchestration to specify the instance ID.
+    /// Returns a new <see cref="SubOrchestrationOptions" /> with the provided instance ID. This can be used when
+    /// starting a new sub-orchestration to specify the instance ID.
     /// </summary>
     /// <param name="instanceId">The instance ID to use.</param>
-    /// <returns>A new <see cref="OrchestrationOptions" />.</returns>
-    public OrchestrationOptions WithInstanceId(string instanceId) => new(this, instanceId);
-
-    /// <summary>
-    /// Gets the instance ID, if available, for this options instance.
-    /// </summary>
-    /// <returns>The orchestration instance ID if available, <c>null</c> otherwise.</returns>
-    internal string? GetInstanceId() => this is OrchestrationOptions options ? options.InstanceId : null;
+    /// <returns>A new <see cref="SubOrchestrationOptions" />.</returns>
+    public SubOrchestrationOptions WithInstanceId(string instanceId) => new(this, instanceId);
 }
 
 /// <summary>
 /// Options that can be used to control the behavior of orchestrator task execution. This derived type can be used to
 /// supply extra options for orchestrations.
 /// </summary>
-public record OrchestrationOptions : TaskOptions
+public record SubOrchestrationOptions : TaskOptions
 {
     /// <summary>
-    /// Initializes a new instance of the <see cref="OrchestrationOptions"/> class.
+    /// Initializes a new instance of the <see cref="SubOrchestrationOptions"/> class.
     /// </summary>
     /// <param name="retry">The task retry options.</param>
     /// <param name="instanceId">The orchestration instance ID.</param>
-    public OrchestrationOptions(TaskRetryOptions? retry = null, string? instanceId = null)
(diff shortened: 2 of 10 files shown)


src/Abstractions/TaskOptions.cs
```diff
@@ -6,153 +6,61 @@ namespace Microsoft.DurableTask;
...
     /// <summary>
-    /// Convenience method from creating a <see cref="TaskOptions"/> object from a <see cref="RetryHandler"/>.
+    /// Gets the instance ID, if available, for this options instance.
     /// </summary>
-    /// <param name="retryHandler">The task retry handler to configure.</param>
-    /// <param name="cancellationToken">Optional cancellation token for canceling the task.</param>
-    /// <returns>Returns a newly created <see cref="TaskOptions"/> object.</returns>
-    public static TaskOptions FromRetryHandler(RetryHandler retryHandler, CancellationToken cancellationToken = default)
+    /// <returns>The orchestration instance ID if available, <c>null</c> otherwise.</returns>
+    internal string? GetInstanceId()
```
jviau, line 41: Might need to make this `public` for external implementors to consume.

src/Abstractions/TaskOptions.cs
```diff
@@ -6,153 +6,61 @@ namespace Microsoft.DurableTask;
...
-            this.CancellationToken = cancellationToken;
-            return this;
-        }
+/// <summary>
+/// Options that can be used to control the behavior of orchestrator task execution. This derived type can be used to
+/// supply extra options for orchestrations.
+/// </summary>
+/// <param name="InstanceId">The orchestration instance ID to use. <c>null</c> to have one generated.</param>
+/// <param name="Retry">The retry options. <c>null</c> for no retries.</param>
+public record OrchestrationOptions(string? InstanceId = null, TaskRetryOptions? Retry = null)
```
jviau, line 53: Open to suggestions for a better name (if one is needed that is). Maybe `SubOrchestrationOptions` because it technically only applies to sub-orchestrations?

src/Abstractions/TaskOptions.cs
```diff
@@ -6,153 +6,61 @@ namespace Microsoft.DurableTask;
...
+/// Options for submitting new orchestrations via the client.
+/// </summary>
+/// <param name="InstanceId">
+/// The unique ID of the orchestration instance to schedule. If not specified, a new GUID value is used.
+/// </param>
+/// <param name="StartAt">
+/// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
+/// is specified, the orchestration instance will be scheduled immediately.
+/// </param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null);
```
jviau, line 66: Similar here re: naming. Alternative could be `NewOrchestrationOptions`.

src/Abstractions/Internal/IOrchestrationSubmitter.cs
```diff
@@ -23,21 +23,15 @@ public interface IOrchestrationSubmitter
...
     /// <returns>
     /// A task that completes when the orchestration instance is successfully scheduled. The value of this task is
-    /// the instance ID of the scheduled orchestration instance. If a non-null <paramref name="instanceId"/> parameter
-    /// value was provided, the same value will be returned by the completed task.
+    /// the instance ID of the scheduled orchestration instance. If a non-null instance ID was provided via
+    /// <paramref name="options" />, the same value will be returned by the completed task.
     /// </returns>
     Task<string> ScheduleNewOrchestrationInstanceAsync(
-        TaskName orchestratorName, string? instanceId = null, object? input = null, DateTimeOffset? startTime = null);
+        TaskName orchestratorName, object? input = null, StartOrchestrationOptions? options = null);
```
cgillum, line 36: This is a great change.

src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -260,10 +260,9 @@ public async Task<T> WaitForExternalEvent<T>(string eventName, TimeSpan timeout)
     /// <typeparam name="TResult">
     /// The type into which to deserialize the sub-orchestrator's output.
     /// </typeparam>
-    /// <inheritdoc cref="CallSubOrchestratorAsync(TaskName, string?, object?, TaskOptions?)"/>
+    /// <inheritdoc cref="CallSubOrchestratorAsync(TaskName, object?, TaskOptions?)"/>
     public abstract Task<TResult> CallSubOrchestratorAsync<TResult>(
         TaskName orchestratorName,
-        string? instanceId = null,
         object? input = null,
         TaskOptions? options = null);
```
cgillum, line 267: Have you considered simply changing `TaskOptions` to `OrchestrationOptions` here? I wonder if that would make setting the instance ID more discoverable.

src/Abstractions/TaskOptions.cs
cgillum: External implementors of what, specifically?

src/Abstractions/TaskOptions.cs
cgillum: I was thinking the same thing about `SubOrchestrationOptions` as being a better name, mainly because I can't think of a case where we'd want to reuse this type for anything other than sub-orchestrations.

src/Abstractions/TaskOptions.cs
cgillum: The risky thing about `NewOrchestrationOptions` is that users might wrongly assume that it also applies to sub-orchestrations.

cgillum (commented):
I love these changes. I know you did this mainly for simplifying code-gen, but I think these are independently great changes for making the public APIs more maintainable.

src/Abstractions/Internal/IOrchestrationSubmitter.cs
jviau: Yeah, I like it because as we add more start options (e.g. Tags), we don't need to change the contract. Just add it to the options object.

src/Abstractions/TaskOptions.cs
jviau: Functions implementation of `TaskOrchestrationContext` is one example - it would want access to this. But I think I can address this with a refactor.

src/Abstractions/TaskOrchestrationContext.cs
jviau: The intent here was to allow re-use of the same `TaskOptions` object between activity invocation and sub-orchestration invocation, with appending instance ID as the less-often use case.

There are 3 ways I see to make it more discoverable:

1. [**Current**] - have a `.WithInstanceId(string? instanceId)` method on `TaskOptions`. - this shows you _can_ use instanceId if you want to.
2. Add an overload which takes in `OrchestrationOptions` and passes to the `TaskOptions` overload.
    - A little counter-intuitive from implementers accessing `InstanceId`, but it saves object creation/copying in what I assume is the more common scenario (supplying retry options, no instance Id - just use `TaskOptions`)
3. Like option 2, but reverse: `TaskOptions` overload passes to `OrchestrationOptions` overload.
    - Implementations have access to `InstanceId` without any special logic. However, using `TaskOptions` will always lead to an object copy.

src/Abstractions/TaskOptions.cs
jviau: I opted to remove this. Implementors should determine their own way to test for instanceId.

src/Abstractions/TaskOrchestrationContext.cs
jviau: After further though, I kept with the current style. The reason is the options parameter style is meant to be flexible, different implementors can define their own options. For example, maybe durable functions extension wants to define a `FunctionsTaskOptions` with special functions-only behavior. Going with option 3 would prevent this behavior as custom types would always be lost as it is converted to `SubOrchestrationOptions`. This leaves option 1 or 2, and option 2 would not be a breaking change, so we can add that later if we get customer feedback about discoverability concerns.

cgillum: I see your point about reusing the same `TaskOptions` object between activity and sub-orchestrations - I agree there is value there. I think I also see your point about defining custom options types. I do worry about discoverability, but custom instance IDs probably enough of a special case that it's not going to hurt anyone to have to find this simple extension method. I'm good to keep the design as-is.

cgillum (approved):
LGTM!

PR #81, opened 2022-11-11
Auto-register default DurableTaskClient. Fix multiple registrations

jviau:
This PR performs two key changes:

1. Fixes an issue where we would register multiple `DurableTaskClient` to the DI container, hampering the ability to directly import a client in the multi-client scenario.
2. Auto-registers the default client directly to the DI container. This will help the common scenario with default (nameless) `DurableTaskClient`s.

### DIFF
--- a/src/Client/Core/DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Client/Core/DependencyInjection/ServiceCollectionExtensions.cs
@@ -43,7 +43,16 @@ public static class ServiceCollectionExtensions
             // The added toggle logic is because we cannot use TryAddEnumerable logic as
             // we would have to dynamically compile a lambda to have it work correctly.
             ConfigureDurableOptions(services, name);
-            services.AddSingleton(sp => builder.Build(sp));
+
+            // We do not want to register DurableTaskClient type directly so we can keep a max of 1 DurableTaskClients
+            // registered, allowing for direct-DI of the default client.
+            services.AddSingleton(sp => new DefaultDurableTaskClientProvider.ClientContainer(builder.Build(sp)));
+
+            if (name == Options.DefaultName)
+            {
+                // If we have the default options name here, we will inject this client directly.
+                builder.RegisterDirectly();
+            }
         }
 
         return services;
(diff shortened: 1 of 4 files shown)


src/Client/Core/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -43,7 +43,16 @@ public static class ServiceCollectionExtensions
             // The added toggle logic is because we cannot use TryAddEnumerable logic as
             // we would have to dynamically compile a lambda to have it work correctly.
             ConfigureDurableOptions(services, name);
-            services.AddSingleton(sp => builder.Build(sp));
+
+            // We do not want to register DurableTaskClient type directly so we can keep a max of 1 DurableTaskClients
+            // registered, allowing for direct-DI of the default client.
+            services.AddSingleton(sp => new DefaultDurableTaskClientProvider.ClientContainer(builder.Build(sp)));
```
cgillum, line 49: I'm not sure I understand what problem is being solved. Why do we need a maximum of 1 `DurableTaskClient` registered? And why isn't `AddSingleton` sufficient for this?  What are we accomplishing by wrapping it in a container?

jviau: The current design allows for multiple `DurableTaskClient` to be configured and registered by an application. But they cannot **all** be registered to the DI container, because then any import of `DurableTaskClient` is ambiguous - the container does not know _which one_ you want to import. To rectify this, we need to ensure only a _single_ `DurableTaskClient` is registered. The `services.AddSingleton(sp => new builder.Build(sp));` call was breaking that - it caused each named client to be registered separately. To solve that I had to introduce a layer of separation for that registration, which in this case is just a simple wrapper type. This is just so the registered type is different, and it is not turning the `DurableTaskClient` to `IEnumerable` int he container.

cgillum: So this wrapper type is simply a mechanism to distinguish the default client from the named clients? How does it change the code a user needs to write to resolve a default `DurableTaskClient` object?

jviau: No change to their code, this is a fix so that it _can_ be done via injecting `DurableTaskClient` into your types constructor.

cgillum (approved):

PR #82, opened 2022-11-14
Allow for inheriting DurableTask client/worker options

jviau:
This PR refactors the worker and client options pattern to reduce total options needed for the final classes. This is done via the following:

1. `DurableTaskWorkerOptions` and `DurableTaskClientOptions` are now inheritable.
2. You can _optionally_ inherit from those types and use the `.UseBuildTarget<TTarget, TOptions>()` extension for the respective builder
3. Configuration done on the base type will be mapped to your derived type.
4. This is all optional - an implementation can choose to only use the base options type, or no options at all.

Additional changes in this PR:

1. Made `JsonDataConverter` constructor public (so users can specify their own options)
2. Added "Worker" to the naming of all the builder types in `Microsoft.DurableTask.Worker`.
3. Exposed a public simpler ctor for `GrpcDurableTaskClient` and made the type public. This is to enable construction during `IInputConverter` in durable functions.

### DIFF
--- a/src/Client/Grpc/GrpcDurableTaskClient.cs
+++ b/src/Client/Grpc/GrpcDurableTaskClient.cs
@@ -4,6 +4,7 @@
 using System.Text;
 using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using static Microsoft.DurableTask.Protobuf.TaskHubSidecarService;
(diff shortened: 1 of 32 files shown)


src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -14,34 +15,42 @@ namespace Microsoft.DurableTask.Client.Grpc;
...
     /// <param name="name">The name of the client.</param>
-    /// <param name="options">The common client options.</param>
-    /// <param name="grpcOptions">The gRPC client options.</param>
+    /// <param name="options">The gRPC client options.</param>
     /// <param name="logger">The logger.</param>
+    [ActivatorUtilitiesConstructor]
     public GrpcDurableTaskClient(
-        string name,
-        DurableTaskClientOptions options,
-        IOptionsMonitor<GrpcDurableTaskClientOptions> grpcOptions,
```
cgillum, line 34: I guess it's nice that we have just one "options" here instead of two now. :)

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -320,13 +329,13 @@ OrchestrationMetadata CreateMetadata(P.OrchestrationState state, bool includeInp
...
     {
-        if (this.grpcOptions.Channel is Channel c)
+        if (this.options.Channel is Channel c)
         {
             channel = c;
             return default;
         }
 
-        string address = string.IsNullOrEmpty(this.grpcOptions.Address) ? "127.0.0.1:4001" : this.grpcOptions.Address!;
+        string address = string.IsNullOrEmpty(this.options.Address) ? "127.0.0.1:4001" : this.options.Address!;
```
cgillum, line 338: BTW, I've been meaning to open a PR for this, but would you be willing to change this default address to `"localhost:4001"` for me? I noticed that `localhost` works better in mixed environments, such as when a client is running on Windows and the server is running on a local WSL2 virtual machine (I discovered this when testing this SDK against Dapr). `127.0.0.1` from a Windows client unfortunately wasn't able to reach an endpoint running on WSL2.

src/Worker/Grpc/GrpcDurableTaskWorker.cs
```diff
@@ -52,13 +50,13 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
     {
-        if (this.grpcOptions.Channel is Channel c)
+        if (this.options.Channel is Channel c)
         {
             channel = c;
             return default;
         }
 
-        string address = string.IsNullOrEmpty(this.grpcOptions.Address) ? "127.0.0.1:4001" : this.grpcOptions.Address!;
+        string address = string.IsNullOrEmpty(this.options.Address) ? "127.0.0.1:4001" : this.options.Address!;
```
cgillum, line 59: Same as with my other comment, wondering if you'd be willing to swap `127.0.0.1` for `localhost` here too...

cgillum (approved):

src/Worker/Grpc/GrpcDurableTaskWorker.cs
jviau: Done

src/Client/Grpc/GrpcDurableTaskClient.cs
jviau: Done

src/Client/Grpc/GrpcDurableTaskClient.cs
jviau: Yes, that was my primary goal here.

PR #87, opened 2022-11-18
Update sign task, snupkg symbols, readme.md

jviau:
1. Fixes the path for the sign task `src` --> `out/bin`
2. Switches debug symbols from `embedded` to be delivered via `snupkg`.
    - This reduces the size of our nupkgs, and lets debuggers download the pdb's separately only when needed (this happens automatically for the user via NuGet and snupkg's)
3. Update `readme.md` with the recent API changes, also removed the feature parity callout as we are still addressing that aspect.
4. Moved `InternalsVisibleTo.cs` to be generated earlier in build to address source link issue.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -25,8 +25,6 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
     <PackageReleaseNotes>$(RepositoryUrl)/releases/</PackageReleaseNotes>
     <PackageTags>Microsoft Durable Task Orchestration Workflow Activity Reliable DTFx</PackageTags>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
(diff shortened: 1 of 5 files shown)


src/Directory.Build.props
```diff
@@ -26,6 +25,8 @@
     <PackageProjectUrl>$(RepositoryUrl)</PackageProjectUrl>
     <PackageReleaseNotes>$(RepositoryUrl)/releases/</PackageReleaseNotes>
     <PackageTags>Microsoft Durable Task Orchestration Workflow Activity Reliable DTFx</PackageTags>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
cgillum, line 28: I've never dealt with snupkg's. How do they get published?

jviau: The same as a nupkg - `nuget push *.snupkg`

cgillum (approved):

PR #98, opened 2022-12-17
Multi-target gRPC packages. Use protobuf directly

### LINKED ISSUE #96: Use Grpc.Net.Client when targeting non-netfx
`Grpc.Core` is out-of-support, replaced with `Grpc.Net.Client`. However, the new library drops netfx support. So, we will need to take on the follow work:

- [ ] Submodule https://github.com/microsoft/durabletask-protobuf directly, as `Microsoft.DurableTask.Sidecar.Protobuf` references `Grpc.Core`.
- [ ] Multi-target `Microsoft.DurableTask.Client.Grpc` and `Microsoft.DurableTask.Worker.Grpc` to `netstandard2.0` and `net6.0` (or net5.0?), and use `Grpc.Core` and `Grpc.Net.Client` under the two targets respectively.

jviau:
This PR does some significant overhauling of gRPC dependencies. The changes breakdown as follows:

1. `Microsoft.DurableTask.Sidecar.Protobuf` dependency removed. https://github.com/microsoft/durabletask-protobuf is not a direct submodule.
1. `Microsoft.DurableTask.Grpc` project & package added, which is purely the dotnet gRPC services for the above protobuf spec.
    - This means it is a duplicate of `Microsoft.DurableTask.Sidecar.Protobuf`, but I figure this belongs in this repo anyways. I am open to a different naming.
    - This does have a meaningful (and necessary) change from the  `Microsoft.DurableTask.Sidecar.Protobuf`: it is multi-target of `netstandard2.0` and `net6.0`. Using `Grpc.Core` and `Grpc.Net.Client` respectively.'
    - I intentionally did not add the `Shared/Grpc` files to this project because I am still unsure what the full picture of this new project is yet. Does it take a dependency on `Microsoft.DurableTask.Abstractions`? Should its name be changed as it is confusing with the two existing gRPC packages?
1. `Microsoft.DurableTask.Client.Grpc` and `Microsoft.DurableTask.Worker.Grpc` have also been multi-targeted, using the newer `Grpc.Net.Client` when targeting `net6.0` and above.
1. Several types were affected to accomplish the multi-targeting. They are minimal changes, mostly branching between `Grpc.Core.Channel` and `Grpc.Net.Client.GrpcChannel` via `#ifdef` statements. 
1. Some gRPC test projects were updated to also be multi-target
    - I couldn't update the integration tests, as they rely on `Microsoft.DurableTask.Sidecar` which is not compatible with netfx.
1. Unnecessary packages were removed from `Microsoft.DurableTask.Worker.Grpc`

Resolves #96

### DIFF
--- a/src/Grpc/Grpc.csproj
+++ b/src/Grpc/Grpc.csproj
@@ -1,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <PackageDescription>The gRPC Protobuf .NET services for Durable Task Framework.</PackageDescription>
+  </PropertyGroup>
+
+  <!-- Version info -->
+  <PropertyGroup>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <VersionSuffix>rc.1</VersionSuffix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Grpc.Tools" Version="2.51.0" PrivateAssets="All" />
+    <PackageReference Include="Google.Protobuf" Version="3.21.12" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Grpc.Core" Version="2.46.5" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <PackageReference Include="Grpc.Net.Client" Version="2.50.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Protobuf Include="$(EngRoot)proto/protos/orchestrator_service.proto" GrpcServices="both" Link="proto/orchestrator_service.proto" />
+  </ItemGroup>
+
+</Project>
--- a/src/Worker/Grpc/Worker.Grpc.csproj
+++ b/src/Worker/Grpc/Worker.Grpc.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-	  <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <PackageDescription>The gRPC worker for the Durable Task Framework.</PackageDescription>
     <EnableStyleCop>true</EnableStyleCop>
   </PropertyGroup>
--- a/src/Worker/Grpc/Worker.Grpc.csproj
+++ b/src/Worker/Grpc/Worker.Grpc.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+	  <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <PackageDescription>The gRPC worker for the Durable Task Framework.</PackageDescription>
     <EnableStyleCop>true</EnableStyleCop>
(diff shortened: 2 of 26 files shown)


src/Worker/Grpc/Worker.Grpc.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+	  <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```
cgillum, line 4: Looks like a `\t` snuck in here.

test/Worker/Grpc.Tests/DependencyInjection/DurableTaskBuilderExtensionsTests.cs
```diff
@@ -1,9 +1,17 @@
...
-using Grpc.Core;
 using Microsoft.Extensions.DependencyInjection;
 
+#if NET6_0_OR_GREATER
+using Grpc.Net.Client;
+#endif
+
+#if NETFRAMEWORK
+using Grpc.Core;
+using GrpcChannel = Grpc.Core.Channel;
```
cgillum, line 12: Is this something that global usings support? I'm wondering if there is a simple way we can reduce some of this duplication.

src/Grpc/Grpc.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <PackageDescription>The gRPC Protobuf .NET services for Durable Task Framework.</PackageDescription>
+  </PropertyGroup>
```
cgillum, line 6: What's the name of this nuget package? I assume it's hidden away via some dynamic MSBuild configuration?

cgillum (approved):
Yikes, the M1 macOS compatibility issue is unfortunate - I was really hoping we wouldn't need to multitarget like this. 😅

Overall changes look good. However, I'd avoid including `Sidecar` in the package name. `Microsoft.DurableTask.Protobuf` sounds good to me.

Unrelated to this PR, do you think we need a similar change for the Durable extension that runs in the Azure Functions host?

src/Grpc/Grpc.csproj
jviau: It is `Microsoft.DurableTask.Grpc`

test/Worker/Grpc.Tests/DependencyInjection/DurableTaskBuilderExtensionsTests.cs
jviau: Yeah! let me try that.

jviau:
> Unrelated to this PR, do you think we need a similar change for the Durable extension that runs in the Azure Functions host?

Yes, I will look into making a similar change there.

src/Grpc/Grpc.csproj
cgillum: I'm good with `Microsoft.DurableTask.Grpc` since we have gRPC-specific dependencies we're pulling in.

R00iBaard:
Hey, 

Thanks for all the hard work! Just a question, should the latest release 1.0.0 resolve this issue without manual intervention? Still get the following error on my mac m1

image

jviau:
@R00iBaard we still need to update the Durable Functions host [extension](https://github.com/Azure/azure-functions-durable-extension/tree/dev/src/WebJobs.Extensions.DurableTask). When that is multi-targeted, then it should work without any manual intervention.

R00iBaard:
> @R00iBaard we still need to update the Durable Functions host [extension](https://github.com/Azure/azure-functions-durable-extension/tree/dev/src/WebJobs.Extensions.DurableTask). When that is multi-targeted, then it should work without any manual intervention.

ah great news! Thanks so much

PR #101, opened 2022-12-20
Add Benchmarks project

jviau:
Adds benchmark project and some minor benchmarks.

## Background

I was making a change which I _thought_ would be a performance improvement, so I added the benchmark project to validate that. Turns out the perf change was negligible, so I undid that change. But might as well check in this project for future use.


cgillum (approved):
I'm fine with having this, especially if it helps us avoid unnecessary performance optimizations. :)

PR #102, opened 2022-12-22
Add Client.OrchestrationServiceClientShim

### LINKED ISSUE #92: Backwards compatible client for DurableTask standalone via IOrchestrationServiceClient
It would be useful for us to add a client implementation which allows you to provide an `IOrchestrationServiceClient` directly, so that users can query backends without needing the sidecar or to be in Durable Functions.

jviau:
This PR adds `Microsoft.DurableTask.Client.CompatShim`, which implements a `DurableTaskClient` backed by a `IOrchestrationServiceClient` directly for talking to the backend. This allows for interacting with the backend in scenarios where you do not have a gRPC endpoint. Primarily you use Durable Functions, but want to use a client to talk to the same backend from outside of Azure Functions.

resolves #92 

## TODO

- [x] Unit tests for `ShimDurableTaskClient`.


cgillum (commented):
It's not immediately clear to me what we want to achieve with this. Would you be able to show some sample code that uses this new feature?

jviau:
This originates from https://github.com/Azure/azure-functions-durable-extension/issues/2250 - it is to support a minimal client to access your backend without needing a sidecar. The primary use case would be for interacting with Durable Functions backend from outside Functions environment - say an AspNet Core app.

src/Client/CompatShim/ShimDurableTaskClient.cs
```diff
@@ -0,0 +1,261 @@
...
+                throw new InvalidOperationException($"Orchestration with instanceId '{instanceId}' does not exist");
+            }
+
+            if (metadata.RuntimeStatus != OrchestrationRuntimeStatus.Pending)
+            {
+                // TODO: Evaluate what to do with "Suspended" state. Do we wait on that?
+                return metadata;
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(5), cancellation);
```
cgillum, line 211: 5 seconds is pretty long. Was this chosen arbitrarily or did you copy it from somewhere else? I wonder if 1 second would be better...

cgillum (commented):
What about "OrchestrationServiceClientShim" as an alternative to "CompatShim"? Compat-shim is ambiguous because it's not clear what we're trying to be compatible with (for example, I assumed incorrectly that this was about backwards compatibility with Durable Functions).

My other thought is that in order for this to be useful in a Durable Functions context, we'll want to have a mechanism for registering an `IOrchestrationServiceClient` that can automatically detect and configure the right backend to use, based on the content of host.json. I'm not exactly sure how to do this given that the SDK is separate from the host, but otherwise users will be burdened with the task of initializing the correct IOrchestrationServiceClient implementation themselves, which I don't think they'll have any idea how to do.

src/Client/CompatShim/ShimDurableTaskClient.cs
jviau: Changed to 1s. I think I used 5s as there was an instance of a 5s delay in the gRPC processor, but that is for another use case anyways.

jviau:
`Microsoft.DurableTask.Client.OrchestrationServiceShim` or `Microsoft.DurableTask.Client.OrchestrationServiceClientShim` as the package name?

As for contents of `host.json`: this extension is **not** meant to be used in a Functions context. In Functions you can use the gRPC client. This is meant for a scenario where someone uses Durable Functions, but then also has some non-Functions apps that want to talk to the backend. We don't have an official/GA gRPC sidecar yet for them to use the gRPC client, so this is to fill that gap for now.

jviau:
@cgillum I have renamed the new package.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -0,0 +1,261 @@
...
+                // TODO: Evaluate what to do with "Suspended" state. Do we wait on that?
+                return metadata;
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(1), cancellation);
+        }
+    }
+
+    [return: NotNullIfNotNull("state")]
+    OrchestrationMetadata? ToMetadata(Core.OrchestrationState? state, bool getInputsAndOutputs)
```
cgillum, line 216: I assume we'll rename this once you merge in your API rename PR.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    T CastClient<T>()
+    {
+        if (this.Client is T t)
+        {
+            return t;
+        }
+
+        throw new NotSupportedException($"Provided IOrchestrationClient does not implement {typeof(T)}.");
```
cgillum, line 243: Did you mean IOrchestrationServiceClient here instead of IOrchestrationClient?

cgillum (approved):
Just a couple non-blockers.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
jviau: No plans to rename this. It is a private method and it is converting to a `*Metadata` type, which is why I called it `ToMetadata`

PR #103, opened 2023-01-05
Update DurableTaskClient API to have consistent cancellation tokens

jviau:
This PR introduces some (breaking) changes to `DurableTaskClient`.

1. Ensure all `async` methods take in `CancellationToken`.
2. Ensure that the `CancellationToken` is the _last_ parameter
3. Add some QoL `virtual` overloads to get to the `CancellationToken` faster, skipping optional parameters.
    - This is also important when one of those parameters is `object`, to avoid mistakenly providing `CancellationToken` as that `object` parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ﻿# Changelog
 
+## v1.0.0
+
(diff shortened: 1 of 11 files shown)


cgillum (approved):
LGTM! Should we update the CHANGELOG.md file as well to record this breaking change?

PR #104, opened 2023-01-06
Allow for injecting ILogger in ITaskOrchestrator

### LINKED ISSUE #46: Provide replay-safe ILogger in orchestrations
At the moment, the orchestrator is created as a singleton using parameterless constructor. 

```cs
static readonly ITaskOrchestrator singletonOrchestrator = new DurablePocoPayloadSample.Orchestrator();
```

Would it be possible to create it using ActivatorUtilities.CreateInstance instead so we could inject ILogger, IOptions and the like?

jviau:
Introduces restricted dependency injection for `ITaskOrchestrator`. The first supported interfaces are `ILogger<T>` and `ILoggerFactory`. The resulting services will be/provide replay-safe loggers. Attempting to resolve any other type from the `IServiceProvider` will throw.

### TODO

- [x] Unit Tests
- [ ] Integration Tests

resolves #46

### DIFF
--- a/src/Worker/Core/Shims/TaskOrchestrationShim.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationShim.cs
@@ -47,6 +47,7 @@ partial class TaskOrchestrationShim : TaskOrchestration
         object? input = this.DataConverter.Deserialize(rawInput, this.implementation.InputType);
         ILogger contextLogger = this.LoggerFactory.CreateLogger("Microsoft.DurableTask");
         this.wrapperContext = new(innerContext, this.invocationContext, contextLogger, input);
+        TaskOrchestrationContextAccessor.Current = this.wrapperContext;
         object? output = await this.implementation.RunAsync(this.wrapperContext, input);
 
         // Return the output (if any) as a serialized string.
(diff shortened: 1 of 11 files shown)


src/Worker/Core/Shims/TaskOrchestrationShim.cs
```diff
@@ -47,6 +47,7 @@ partial class TaskOrchestrationShim : TaskOrchestration
         object? input = this.DataConverter.Deserialize(rawInput, this.implementation.InputType);
         ILogger contextLogger = this.LoggerFactory.CreateLogger("Microsoft.DurableTask");
         this.wrapperContext = new(innerContext, this.invocationContext, contextLogger, input);
+        TaskOrchestrationContextAccessor.Current = this.wrapperContext;
```
jviau, line 50: This should have been obvious to me, but the inner orchestrator is constructed before this point, so this fails. Will need to rethink how to smuggle the `TaskOrchestratorContext` for use in `OrchestrationLoggerFactory`.

jviau:
The issue I encountered is not easily solvable, requiring considerable changes to how things are constructed. Don't feel like addressing that right now, will come back to this approach later.

PR #110, opened 2023-01-19
Add new TaskOption overload for skipping input param

jviau:
Adds some overloads to protect against accidentally supplying `TaskOptions` as `input`.

### DIFF
--- a/samples/AzureFunctionsApp/AzureFunctionsApp.csproj
+++ b/samples/AzureFunctionsApp/AzureFunctionsApp.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <AzureFunctionsVersion>v4</AzureFunctionsVersion>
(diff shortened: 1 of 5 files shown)


samples/AzureFunctionsApp/AzureFunctionsApp.csproj
```diff
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
```
davidmrdavid, line 2: nitpick: should we remove this empty change?

src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -108,15 +108,30 @@ public abstract class TaskOrchestrationContext
...
-        return this.CallActivityAsync<object>(name, input, options);
-    }
+        => this.CallActivityAsync<object>(name, input, options);
+
+    /// <returns>
+    /// A task that completes when the activity completes or fails. The result of the task is the activity's return value.
+    /// </returns>
+    /// <inheritdoc cref="CallActivityAsync(TaskName, object?, TaskOptions?)"/>
+    public virtual Task CallActivityAsync(TaskName name, TaskOptions options)
+        => this.CallActivityAsync(name, null, options);
```
davidmrdavid, line 118: Should we really inherit the docs from the `CallActivityAsync` with the input parameter?

davidmrdavid: Same comment for the changes in CallSubOrchestrator

davidmrdavid (commented):
Almost ready to approve, just a minor question and nitpick

src/Abstractions/TaskOrchestrationContext.cs
jviau: To be honest, I don't really know the answer to that. @cgillum suggested I do it on another PR so I did 🤷

samples/AzureFunctionsApp/AzureFunctionsApp.csproj
jviau: This is intentional - adding a new line for formatting of the document.

davidmrdavid: Oh, interesting. So this is required as per some linter? Anyways, if it's intentional, it's fine

davidmrdavid (approved):
LGTM, although it would be good to hear from @cgillum on the `inheritdoc` question

src/Abstractions/TaskOrchestrationContext.cs
davidmrdavid: Well, even if this were off, it's a non-blocking concern

samples/AzureFunctionsApp/AzureFunctionsApp.csproj
jviau: Not required, my own preference.

PR #112, opened 2023-01-25
Make DurableTaskClient method names consistent

jviau:
This PR does the following:

1. Makes `PurgeInstancesFilter` parameters optional since they were already nullable.
2. Add some overloads for `Suspend` and `Resume`
3. Makes method names in `DurableTaskClient` more consistent.
    - `TerminateAsync` -> `TerminateInstanceAsync`
    - `GetInstances` -> `GetAllInstancesAsync`
    - `GetInstanceMetadataAsync` -> `GetInstanceAsync`
    - `PurgeInstances` -> `PurgeAllInstancesAsync`
    - `PurgInstanceMetadataAsync` -> `PurgeInstanceAsync`

### DIFF
--- a/src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
+++ b/src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
@@ -30,7 +30,7 @@ public static class DurableTaskClientExtensions
     {
         Check.NotNull(client);
         PurgeInstancesFilter filter = new(createdFrom, createdTo, statuses);
-        return client.PurgeInstanceMetadataAsync(filter, cancellation);
+        return client.PurgeInstancesAsync(filter, cancellation);
     }
 
     /// <summary>
--- a/src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
+++ b/src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
@@ -30,7 +30,7 @@ public static class DurableTaskClientExtensions
     {
         Check.NotNull(client);
         PurgeInstancesFilter filter = new(createdFrom, createdTo, statuses);
-        return client.PurgeInstancesAsync(filter, cancellation);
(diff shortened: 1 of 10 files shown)


cgillum (commented):
Just a quick FYI, the reordering of methods in DurableTaskClient.cs makes the PR really hard to read because of how the diffs get confused. I assume the actual delta is smaller than how it appears?

cgillum (approved):
LGTM! Let's also update the CHANGELOG file to record these breaking changes.

jviau:
Yeah, I called out all the actual changes in the PR description. Re-ordering was to group more similar methods together. Apologies for the confusion it caused.

jviau:
@cgillum I made some method name changes based on your suggestion of referencing Azure SDKs

src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
```diff
@@ -30,7 +30,7 @@ public static class DurableTaskClientExtensions
     {
         Check.NotNull(client);
         PurgeInstancesFilter filter = new(createdFrom, createdTo, statuses);
-        return client.PurgeInstancesAsync(filter, cancellation);
+        return client.PurgeAllInstancesAsync(filter, cancellation);
```
cgillum, line 33: In the places you observed, did they use "All" even for methods that accepted filters?

cgillum (commented):
These changes look good to me - just one question below.

src/Client/Core/DependencyInjection/DurableTaskClientExtensions.cs
jviau: Yes they did: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/compute/Azure.ResourceManager.Compute/src/Generated/VirtualMachineCollection.cs#L231

cgillum (approved):
:shipit:

PR #118, opened 2023-02-03
Add GetInstanceAsync overload to address plural typo

jviau:
I am never going to mentally recover from this typo.


cgillum (approved):
LGTM!

PR #121, opened 2023-02-07
Add request types to allow for mediator pattern invocation of tasks

### LINKED ISSUE #73: Mediator pattern

jviau:
This PR introduces the [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) for invoking of orchestrations and activities. This allows for users to invoke activities and orchestrations via strongly-typed request objects and also to build up complex invocation patterns through additional `TaskOrchestrationContext` extension methods and more derived request types. Some examples would be:

1. A polling pattern. We can introduce a `ILongRuningActivityRequest<TResult>` and a `RunAsync` extension method for it which will handle enqueuing a polling orchestration and polling activity.
2. A paging activity pattern - we can leverage our `AsyncPageable<T>` classes to allow for paging over activities via `IAsyncEnumerable<T>` in your code.

This will also replace the code-generation approach for strongly typed orchestration/activity invocations. In my opinion, I prefer this route over code-generation. It is more flexible and avoids function-name conflicts. We can still explore how we can leverage code-generation to add value to this pattern though - but it will always be an optional addition.

resolves #73

# TODO:

- [x] unit tests
- [x] samples
- [x] `DurableTaskClient` extension method for enqueuing orchestrations.


src/Abstractions/TaskActivity.cs
```diff
@@ -89,31 +83,31 @@ public abstract class TaskActivity<TInput, TOutput> : ITaskActivity
...
-            typedInput = typed;
-            return true;
-        }
-        else if (input is not null && typeof(TInput) != input.GetType())
-        {
-            typedInput = default;
-            return false;
-        }
+/// <inheritdoc cref="TaskActivity{TInput, Unit}" />
+public abstract class TaskActivity<TInput> : ITaskActivity
```
jviau, line 89: I realized a gap with the existing `TaskActivity<TInput, TOutput>` and `TaskOrchestrator<TInput, TOutput>` in the mediator pattern: generic constraints missing. Ideally, we would have this contract:

``` CSharp
public abstract class TaskActivity<TInput, TOutput> : ITaskActivity
    where TInput : IActivityRequest<TOutput>
```

The purpose here is to force the `TOutput` of the `IActivityRequest` and the `TaskActivity` to line up, so there is no risk of mismatched type failing at runtime. This would be a breaking change to introduce though. It also _forces_ you to use request objects as inputs - I would like to retain the ability to use existing types.

I could introduce another derived type for this? But what would the incentive be to use it over the existing types?

cgillum: A new derived type makes sense. As far as incentives for using it, that's a little tricky. On one hand, the incentive of having strongly typed means of calling activities/orchestrations is really nice. On the other hand, we have source generators for that which work with the existing abstractions.

One thought I just had is that Dapr Workflow (which will be another source of customers for this SDK) has a .NET SDK that's a wrapper around this SDK. Because it's a wrapper with completely different types, we can't really take advantage of the existing source generators. It feels like it would be a waste to duplicate that effort and create Dapr Workflow source generators. But if we have the typed mediator pattern in this SDK, we could change the Dapr Workflow SDK to use this new abstraction to provide a more typed experience (that project is in alpha, so we can make breaking changes still). Happy to discuss this further if you're interested.

YunchuWang:
Hi @jviau , thanks for opening this! any update on this?

jviau:
@YunchuWang this PR is quite old, would need some updates and could potentially leverage some new CSharp language features for an improved experience.

Next steps would be to first have a design discussion to see if the durable team wants to take this. I **highly** recommend it. In my usage of durable, it is far more convenient to use a mediator-style API then the source generated strongly-typed APIs. There is opportunity to leverage source-gen to help the mediator APIs also, so can get best of both worlds.

But if you are interested in this, lets sync with interested parties to go over what mediator offers and how it can be the path forward for durable APIs.

PR #127, opened 2023-02-15
Fix nuget determinism

jviau:
Investigating why CI build has this:
image


jviau:
Was not an issue in the end. I needed to also download the `snupkg` and have it beside the `nupkg` for NGPE to properly evaluate those fields.

PR #137, opened 2023-03-20
Fix null parent instance details

jviau:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/2394


cgillum (approved):
:shipit:

PR #172, opened 2023-07-27
Add DurableEntityClient abstractions

### LINKED ISSUE #166: Durable Entities: Client abstractions
## Overview

To support interacting with entities from a client, we need to design & add appropriate APIs to do as such from a `DurableTaskClient`.

### Requirements

The client additions should cover the same functionality from [IDurableEntityClient](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs).

- Signal an entity
- Read an entity state (this gets the entire state, not the response of an entity operation)
- Query entities
- Cull entity storage (removes empty entities, releases orphaned locks)

### Non-Requirements

- Not all exact overloads from the in-proc `IDurableEntityClient` will be supported.
- Invoking entity via interface proxy.

## Design

To avoid breaking changes but also avoid adding a complex plugin plattern (ala `HttpContext.Features`), we will simply design a new standalone abstract client `DurableEntityClient`. We will then add a single property onto `DurableTaskClient`:

``` CSharp
public abstract class DurableTaskClient
{
    public virtual DurableEntityClient Entities => throw new NotSupportedException($"The client {this.GetType()} does not support durable entities.");
}
```

jviau:
resolves #166 

This PR adds the `DurableEntityClient` abstract class and supporting types. The exact API from what is present in DF today is not exact, but the overall features are all there.

# DurableEntityClient

- Get single entity with or without state
- Get/query entities
- Clean entity storage

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.ComponentModel;
+using Microsoft.DurableTask.Client.Entites;
 using Microsoft.DurableTask.Internal;
 
 namespace Microsoft.DurableTask.Client;
@@ -40,6 +41,15 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// Gets the name of the client.
     /// </summary>
     public string Name { get; }
+
+    /// <summary>
+    /// Gets the <see cref="DurableEntityClient"/> for interacting with durable entities.
+    /// </summary>
+    /// <remarks>
+    /// Not all clients support durable entities. Refer to a specific client implementation for verifying support.
+    /// </remarks>
+    public virtual DurableEntityClient Entities =>
+        throw new NotSupportedException($"{this.GetType()} does not support durable entities.");
 
     /// <inheritdoc cref="ScheduleNewOrchestrationInstanceAsync(TaskName, object, StartOrchestrationOptions, CancellationToken)"/>
     public virtual Task<string> ScheduleNewOrchestrationInstanceAsync(
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 using System.ComponentModel;
-using Microsoft.DurableTask.Client.Entites;
+using Microsoft.DurableTask.Client.Entities;
 using Microsoft.DurableTask.Internal;
 
 namespace Microsoft.DurableTask.Client;
--- a/src/Client/Core/Entites/CleanEntityStorage.cs
+++ b/src/Client/Core/Entites/CleanEntityStorage.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Client.Entites;
+
+/// <summary>
+/// Request struct for <see cref="DurableEntityClient.CleanEntityStorageAsync"/>.
+/// </summary>
+public readonly record struct CleanEntityStorageRequest
+{
+    /// <summary>
+    /// Gets a value indicating whether to remove empty entities.
+    /// </summary>
+    /// <remarks>
+    /// An entity is considered empty, and is removed, if it has no state, is not locked.
+    /// </remarks>
+    public bool RemoveEmptyEntities { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether to release orphaned locks or not.
+    /// </summary>
+    /// <remarks>
+    /// Locks are considered orphaned, and are released, and if the orchestration that holds them is not in state
+    /// <see cref="OrchestrationRuntimeStatus.Running"/>. This hould not happen under normal circumstances, but can
+    /// occur if the orchestration instance holding the lock exhibits replay nondeterminism failures, or if it is
+    /// explicitly purged.
+    /// </remarks>
+    public bool ReleaseOrphanedLocks { get; init; }
+}
+
+/// <summary>
+/// Result struct for <see cref="DurableEntityClient.CleanEntityStorageAsync"/>.
+/// </summary>
+public readonly struct CleanEntityStorageResult
+{
+    /// <summary>
+    /// Gets the number of empty entities removed.
+    /// </summary>
+    public int EmptyEntitiesRemoved { get; init; }
+
+    /// <summary>
+    /// Gets the number of orphaned locks that were removed.
+    /// </summary>
+    public int OrphanedLocksRemoved { get; init; }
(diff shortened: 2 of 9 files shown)


src/Client/Core/Entites/EntityQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// The default page size.
+    /// </summary>
+    public const int DefaultPageSize = 100;
+
+    /// <summary>
+    /// Gets the optional name of the entity to query for.
+    /// </summary>
+    public string? Name { get; init; }
+
+    /// <summary>
```
sebastianburckhardt, line 21: I think we should add a `string?`  field `InstanceIdPrefix` because there have been situtations where we need more precise queries. (I have had users ask for this). Our backends support this already, so it will not create extra work for us, and it is not easy to change this later.

src/Client/Core/Entites/EntityQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// Gets entity instances which was last modified after the provided time.
```
sebastianburckhardt, line 22: which was -> which were

src/Client/Core/Entites/EntityQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public DateTimeOffset? LastModifiedFrom { get; init; }
+
+    /// <summary>
+    /// Gets entity instances which was last modified before the provided time.
```
sebastianburckhardt, line 27: which was -> which were

src/Client/Core/Entites/DurableEntityClient.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// Signals an entity to perform an operation.
+    /// </summary>
+    /// <param name="id">The ID of the entity to signal.</param>
+    /// <param name="operationName">The name of the operation.</param>
+    /// <param name="options">The options to signal the entity with.</param>
+    /// <param name="cancellation">The cancellation token to cancel enqueuing of the operation.</param>
+    /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+    /// <remarks>This does not wait for the operation to be processed by the receiving entity.</remarks>
+    public virtual Task SignalEntityAsync(
```
sebastianburckhardt, line 53: I am a bit concerned about this overload (not passing an input) creating a bad pitfall where users forget to pass an input to an operation that must have one specified (i.e. does not have a default). The problem is that they won't see an exception in the client because signals don't wait for a response. The exception will be quietly buried in some log somewhere.

sebastianburckhardt: IMO this pitfall may be worse than the utility provided by default arguments. I don't actually recall ever having a need for default arguments on entity operations so we should weigh the extra complexity and error modes of supporting this feature.

sebastianburckhardt (commented):
Some minor suggestions and questions, some of which I think warrant discussion.

src/Client/Core/Entites/CleanEntityStorage.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Client.Entites;
```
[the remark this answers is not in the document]
jviau: Entire folder is named wrong. Good catch

src/Client/Core/Entites/DurableEntityClient.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <param name="name">The name of the client.</param>
+    protected DurableEntityClient(string name)
+    {
+        this.Name = name;
+    }
+
+    /// <summary>
+    /// Gets the name of the client.
+    /// </summary>
+    public string Name { get; }
```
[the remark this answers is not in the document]
jviau: Either! We have to decide that based on how we configure the builder APIs (if any are added).

src/Client/Core/Entites/DurableEntityClient.cs
jviau: > creating a bad pitfall where users forget to pass an input to an operation that must have one specified

What alternative do you suggest? Users explicitly pass `null` for the input? I don't see this as any _new_ pitfall. Users can mistakenly pass `null` for input, or ignore it with optional params. Additionally, a similar pitfall already exists with passing in incompatible input for an operation.

To me, users already need to be aware of what input is needed - if any - for a given operation. So, I don't see this as a new pitfall - it already exists.

src/Client/Core/Entites/EntityQuery.cs
jviau: I didn't include `InstanceIdPrefix` because `EntityQuery.Name` already translates to an orchestration query instance ID prefix. Is the prefix you suggest different? Is it the prefix of the entity key?

Or is the idea to make `Name` more robust, and give it an "equals" or "startswith" capabilities?

src/Client/Core/DurableTaskClient.cs
```diff
@@ -41,6 +42,15 @@ protected DurableTaskClient(string name)
     /// </summary>
     public string Name { get; }
```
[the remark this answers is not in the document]
jviau: The name is because we allow for configuring and using of multiple named clients. This just exposes the name that client was built with.

src/Client/Core/Entites/EntityQuery.cs
sebastianburckhardt: The InstanceIdPrefix is more expressive because one can create prefixes of the entity name, or full entity names plus prefixes of the entity key.

For example,
1.  if the entities have instance ids where the entity name follows some convention, e.g. look something like @ModuleX_User@X  @ModuleX_Account@Z
then we can retrieve all entities whose name starts with ModuleX by specifying the prefix "@ModuleX"
2. if the entity keys have some internal structure e.g. @Task@a.b, ... @Task@a.c then we can retrieve a subset of the entities using the prefix @Task@a

sebastianburckhardt: > and give it an "equals" or "startswith" capabilities?

something also worth considering I suppose. Did not think that far.

src/Client/Core/Entites/DurableEntityClient.cs
sebastianburckhardt: > I don't see this as any new pitfall. 

true, the current implementation seems to have the same problem. And I can't really think of a better solution (short of using interfaces and proxies which is out of scope).

> The reason I include this overload is to avoid the accidental cancellation token as input - which IMO is an even worse pitfall. 

Thanks for pointing that out, I definitely did not see that. Makes sense.

sebastianburckhardt (approved):

PR #173, opened 2023-07-27
Add TaskOrchestrationContext entity abstractions

### LINKED ISSUE #167: Durable entities: Orchestration context abstractions
## Overview

To support interacting with entities from a client, we need to design & add appropriate APIs to do as such from a `DurableTaskClient`.

### Requirements

The client additions should cover the same functionality from [IDurableOrchestrationContext](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs) entity-related methods.

- Signal an entity
- Call an entity (waits for results)
- Lock entities (& releasing of locks)

### Non-Requirements

- Not all exact overloads from the in-proc `IDurableOrchestrationContext` will be supported.
- Invoking entity via interface proxy.

## Design

To avoid breaking changes but also avoid adding a complex plugin pattern (ala `HttpContext.Features`), we will simply design a new standalone abstract class `TaskOrchestrationEntityFeature`. We will then add a single property onto `TaskOrchestrationContext`:

``` CSharp
public abstract class TaskOrchestrationContext
{
    public virtual TaskOrchestrationEntityFeature Entities => throw new NotSupportedException($"The context {this.GetType()} does not support durable entities.");
}
```

jviau:
resolves #167 

This PR adds abstractions for interacting with entities from an orchestration.

## Supported scenarios:

1. Call entity
2. Signal entity (this is done via `CallEntityAsync`, but passing in an `CallEntityOptions` with `Signal` set on it).
3. Lock entities
4. Check for and retrieve current entity locks

### DIFF
--- a/src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
+++ b/src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
@@ -1,0 +1,97 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Entities;
+
+/// <summary>
+/// Feature for interacting with durable entities from an orchestration.
+/// </summary>
+public abstract class TaskOrchestrationEntityFeature
+{
+    /// <summary>
+    /// Calls an operation on an entity and waits for it to complete. Does not wait for completion if
+    /// <see cref="CallEntityOptions.Signal"/> is set on <paramref name="options"/>.
+    /// </summary>
+    /// <typeparam name="TResult">The result of the entity operation.</typeparam>
+    /// <param name="id">The target entity.</param>
+    /// <param name="operationName">The name of the operation.</param>
+    /// <param name="input">The operation input.</param>
+    /// <param name="options">The call options.</param>
+    /// <returns>
+    /// The result of the entity operation, or default(<typeparamref name="TResult"/>) in the signal-only case.
+    /// </returns>
+    public abstract Task<TResult> CallEntityAsync<TResult>(
+        EntityInstanceId id, string operationName, object? input = null, CallEntityOptions? options = null);
+
+    /// <summary>
+    /// Calls an operation on an entity and waits for it to complete. Does not wait for completion if
+    /// <see cref="CallEntityOptions.Signal"/> is set on <paramref name="options"/>.
+    /// </summary>
+    /// <typeparam name="TResult">The result of the entity operation.</typeparam>
+    /// <param name="id">The target entity.</param>
+    /// <param name="operationName">The name of the operation.</param>
+    /// <param name="options">The call options.</param>
+    /// <returns>
+    /// The result of the entity operation, or default(<typeparamref name="TResult"/>) in the signal-only case.
+    /// </returns>
+    public virtual Task<TResult> CallEntityAsync<TResult>(
+        EntityInstanceId id, string operationName, CallEntityOptions? options)
+        => this.CallEntityAsync<TResult>(id, operationName, null, options);
+
+    /// <summary>
+    /// Calls an operation on an entity and waits for it to complete. Does not wait for completion if
+    /// <see cref="CallEntityOptions.Signal"/> is set on <paramref name="options"/>.
+    /// </summary>
+    /// <param name="id">The target entity.</param>
+    /// <param name="operationName">The name of the operation.</param>
+    /// <param name="input">The operation input.</param>
+    /// <param name="options">The call options.</param>
+    /// <returns>
+    /// A task that completes when the operation has been completed. Or in the signal-only case, a task that is either
+    /// already complete or completes when the operation has been enqueued.
+    /// </returns>
+    public abstract Task CallEntityAsync(
+        EntityInstanceId id, string operationName, object? input = null, CallEntityOptions? options = null);
+
+    /// <summary>
+    /// Calls an operation on an entity and waits for it to complete. Does not wait for completion if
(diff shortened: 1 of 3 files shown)


src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">The result of the entity operation.</typeparam>
+    /// <param name="id">The target entity.</param>
+    /// <param name="operationName">The name of the operation.</param>
+    /// <param name="input">The operation input.</param>
+    /// <param name="options">The call options.</param>
+    /// <returns>
+    /// The result of the entity operation, or default(<typeparamref name="TResult"/>) in the signal-only case.
+    /// </returns>
+    public abstract Task<TResult> CallEntityAsync<TResult>(
```
sebastianburckhardt, line 23: Not sure I like the idea of using "CallEntityAsync" for both calls and signals. From a terminology perspective, we are not considering signals to be a special type of call, they are separate category of communication. Another issue which gets obscured here is that there are significant differences in the parameters and return values:
- there are no call options, only signal options. 
- signal never returns a TResult value.
- signal does not return a Task, it returns void, and is not async.

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// </summary>
+    /// <param name="entityIds">The list of locked entities.</param>
+    /// <returns>True if any locks are owned, false otherwise.</returns>
+    public abstract bool HasEntityLocks(out IReadOnlyList<EntityInstanceId> entityIds);
+
+    /// <summary>
+    /// Gets a value indicating whether any entity locks are owned by this instance.
+    /// </summary>
+    /// <returns>True if any locks are owned, false otherwise.</returns>
+    public virtual bool HasEntityLocks() => this.HasEntityLocks(out _);
```
sebastianburckhardt, line 96: this is not entirely right in terms of functionality.

These are the pieces of information we want to expose:

-  whether the orchestrator is currently inside a critical section
-  if inside a critical section, which entities are currently available for making calls

where a lock is available to make the call if (a) it was locked at the beginning of the critical section, and (b) there is not currently a call to that entity already pending. 

So, whether or not we are in a critical section cannot be determined simply by checking if the list is empty - it can be empty even if we are inside a critical section.

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// </returns>
+    public virtual Task CallEntityAsync(EntityInstanceId id, string operationName, CallEntityOptions? options)
+        => this.CallEntityAsync(id, operationName, null, options);
+
+    /// <summary>
+    /// Acquires one or more entity locks.
+    /// </summary>
+    /// <param name="entityIds">The entity IDs to lock.</param>
+    /// <returns>An async-disposable which can be disposed to release the lock.</returns>
+    public abstract Task<IAsyncDisposable> LockEntitiesAsync(IEnumerable<EntityInstanceId> entityIds);
```
sebastianburckhardt, line 75: I think we previously had a params [] here which I think is a bit more convenient because most of the time the number of elements is static.

jviau: the params overload exists below.

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
sebastianburckhardt: Keeping implementations flexible is probably a good idea. 
If you think it fits better with the other patterns, you can keep it. Personally I would still change it, also just for discoverability (it's not obvious that to find the signal you have to pass an option to call).

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
sebastianburckhardt: This is what I had before:

```
 /// <summary>
 /// Whether this orchestration is currently inside a critical section.
 /// </summary>
 public bool IsInsideCriticalSection => this.criticalSectionId != null;   

 /// <summary>
 /// Enumerate all the entities that are available for calling from within a critical section. 
 /// This set contains all the entities that were locked prior to entering the critical section,
 /// and for which there is not currently an operation call pending.
 /// </summary>
 /// <returns>An enumeration of all the currently available entities.</returns>
 public IEnumerable<(string entityName, string entityKey)> GetAvailableEntities()
 {
     if (this.IsInsideCriticalSection)
     {
         foreach(var e in this.availableLocks)
         {
             yield return (e.Name, e.Key);
         }
     }
 }
```

it is not perfect but it should give you an idea of how it is intended to work.
Another option would be to create 2 overloads of `IsInsideCriticalSection`, where one of them takes an out enumerable parameter. That is probably a little better than the above `GetAvailableEntities`.

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
sebastianburckhardt: ok, did not see it! Great.

src/Abstractions/Entities/TaskOrchestrationEntityFeature.cs
jviau: Would it be sufficient to rename `HasEntityLocks` to `InCriticalSection`?

sebastianburckhardt (approved):

PR #177, opened 2023-08-24
Fix instanceId not passed in when using retry

### LINKED ISSUE #174: TaskOrchestrationContextWrapper.CallSubOrchestratorAsync does not provide instance Id properly when RetryPolicy is specified
Hello there,

In TaskOrchestrationContextWrapper.CallSubOrchestratorAsync , we have:

https://github.com/microsoft/durabletask-dotnet/blob/9e8997ce8e961f13fe418ff0117e538b952a2b2c/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs#L117-L137

Where the `instanceId` extracted from the Option. However, when `options?.Retry?.Policy is RetryPolicy policy` is enabled, this `instanceId` is no longer used.

The function `CreateSubOrchestrationInstanceWithRetry` should support sending `instanceId` if using the following override:

https://github.com/Azure/durabletask/blob/1bb5c8f964f085bcdd6268c0fb5ec76a0701fb5a/src/DurableTask.Core/OrchestrationContext.cs#L249-L265 

I wonder if this needs to be changed

jviau:
Resolves #174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v1.0.3
 
+### Microsoft.DurableTask.Worker
+
+- Fix instance ID not being passed in when using retry policy (https://github.com/microsoft/durabletask-dotnet/issues/174)
+
 ### Microsoft.DurableTask.Worker.Grpc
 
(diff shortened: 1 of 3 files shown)


cgillum (approved):
LGTM!

PR #178, opened 2023-08-25
Add delete implicit operation

jviau:
This PR introduces an implicit `delete` operation when deriving from `TaskEntity<TState>`. This operation will null out the entities state, effectively deleting the entity.


sebastianburckhardt (approved):
Based on offline discussion the current design is sufficient, we may add another form of deletion (PurgeEntity) as a client-issued operation later.

PR #179, opened 2023-08-25
Separate entity state from TaskEntityContext onto its own object

jviau:
This PR separates entity state from `TaskEntityContext` into `TaskEntityState`, which is available off of `TaskEntityOperation`. The PR also removes binding of `TaskEntityOperation` from `TaskEntity<TState>`. The end result is that when using `TaskEntity<TState>`, users have no direct access to the operation or state. This is to prevent confusing behavior when trying to use direct entity state interactions within `TaskEntity<TState>`. If a user wants direct operation access, they should derive from `ITaskEntity`.


sebastianburckhardt (approved):
LGTM.

(I did not look at the tests but given the current priorities and timeline I think that is appropriate)

PR #185, opened 2023-09-06
Connect client entity SDK to gRPC service.

sebastianburckhardt:
Provides the glue to connect the entity client features to the gRPC service `TaskHubSidecarService`.

### DIFF
--- a/src/Client/Core/Entities/CleanEntityStorage.cs
+++ b/src/Client/Core/Entities/CleanEntityStorage.cs
@@ -41,5 +41,5 @@ public readonly record struct CleanEntityStorageResult
     /// <summary>
     /// Gets the number of orphaned locks that were removed.
     /// </summary>
-    public int OrphanedLocksRemoved { get; init; }
+    public int OrphanedLocksReleased { get; init; }
 }
--- a/src/Client/Core/Entities/CleanEntityStorage.cs
+++ b/src/Client/Core/Entities/CleanEntityStorage.cs
@@ -39,7 +39,7 @@ public readonly record struct CleanEntityStorageResult
     public int EmptyEntitiesRemoved { get; init; }
 
     /// <summary>
-    /// Gets the number of orphaned locks that were removed.
+    /// Gets the number of orphaned locks that were released.
     /// </summary>
     public int OrphanedLocksReleased { get; init; }
 }
--- a/src/Client/Core/Entities/CleanEntityStorage.cs
+++ b/src/Client/Core/Entities/CleanEntityStorage.cs
@@ -26,6 +26,11 @@ public readonly record struct CleanEntityStorageRequest
     /// explicitly purged.
     /// </remarks>
     public bool ReleaseOrphanedLocks { get; init; }
+
+    /// <summary>
+    /// Gets the continuation token to resume a previous <see cref="CleanEntityStorageRequest"/>.
+    /// </summary>
+    public string? ContinuationToken { get; init; }
 }
 
 /// <summary>
@@ -42,4 +47,9 @@ public readonly record struct CleanEntityStorageResult
     /// Gets the number of orphaned locks that were released.
     /// </summary>
     public int OrphanedLocksReleased { get; init; }
+
+    /// <summary>
+    /// Gets the continuation token to continue the <see cref="CleanEntityStorageRequest"/>, if not null.
+    /// </summary>
+    public string? ContinuationToken { get; init; }
 }
--- a/src/Client/Grpc/GrpcDurableEntityClient.cs
+++ b/src/Client/Grpc/GrpcDurableEntityClient.cs
@@ -141,6 +141,7 @@ class GrpcDurableEntityClient : DurableEntityClient
                 {
                     RemoveEmptyEntities = request.RemoveEmptyEntities,
                     ReleaseOrphanedLocks = request.ReleaseOrphanedLocks,
+                    ContinuationToken = request.ContinuationToken,
                 },
                 cancellationToken: cancellation);
 
@@ -148,6 +149,7 @@ class GrpcDurableEntityClient : DurableEntityClient
             {
                 EmptyEntitiesRemoved = response.EmptyEntitiesRemoved,
                 OrphanedLocksReleased = response.OrphanedLocksReleased,
+                ContinuationToken = response.ContinuationToken,
             };
         }
         catch (RpcException e) when (e.StatusCode == StatusCode.Cancelled)
--- a/src/Client/Grpc/GrpcDurableEntityClient.cs
+++ b/src/Client/Grpc/GrpcDurableEntityClient.cs
@@ -132,24 +132,36 @@ class GrpcDurableEntityClient : DurableEntityClient
     }
 
     /// <inheritdoc/>
-    public override async Task<CleanEntityStorageResult> CleanEntityStorageAsync(CleanEntityStorageRequest request = default, CancellationToken cancellation = default)
+    public override async Task<CleanEntityStorageResult> CleanEntityStorageAsync(CleanEntityStorageRequest request = default, bool continueUntilComplete = true, CancellationToken cancellation = default)
     {
+        string? continuationToken = request.ContinuationToken;
+        int emptyEntitiesRemoved = 0;
+        int orphanedLocksReleased = 0;
+
         try
         {
-            P.CleanEntityStorageResponse response = await this.sidecarClient.CleanEntityStorageAsync(
-                new P.CleanEntityStorageRequest
-                {
-                    RemoveEmptyEntities = request.RemoveEmptyEntities,
-                    ReleaseOrphanedLocks = request.ReleaseOrphanedLocks,
-                    ContinuationToken = request.ContinuationToken,
-                },
-                cancellationToken: cancellation);
+            do
+            {
+                P.CleanEntityStorageResponse response = await this.sidecarClient.CleanEntityStorageAsync(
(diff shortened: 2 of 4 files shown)


src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+        Guid requestId = Guid.NewGuid();
+        DateTimeOffset? scheduledTime = options?.SignalTime;
+
+        P.SignalEntityRequest request = new P.SignalEntityRequest()
+        {
+            InstanceId = id.ToString(),
+            RequestId = Google.Protobuf.ByteString.FromStream(new MemoryStream(requestId.ToByteArray())),
+            Name = operationName,
+            Input = this.durableTaskClient.DataConverter.Serialize(input),
+            ScheduledTime = scheduledTime.HasValue ? Timestamp.FromDateTimeOffset(scheduledTime.Value.ToUniversalTime()) : default,
```
jviau, line 52: I think we have an extension method for converting `DateTimeOffset?` to `Timestamp`
https://github.com/microsoft/durabletask-dotnet/blob/main/src/Shared/Grpc/ProtoUtils.cs#L216

```suggestion
            ScheduledTime = scheduledTime.ToTimestamp(),
```

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+    {
+        P.GetEntityRequest request = new P.GetEntityRequest()
+        {
+            InstanceId = id.ToString(),
+            IncludeState = includeState,
+        };
+
+        P.GetEntityResponse response = await this.sidecarClient.GetEntityAsync(request, cancellationToken: cancellation);
+
+        return response == null ? null : this.ToEntityMetadata(response.Entity, includeState);
```
jviau, line 70: I don't think this will ever be null, it will most likely throw before that. Or the server has a bad implementation, in which case we should just fail.

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <inheritdoc/>
+    public override async Task<EntityMetadata?> GetEntityAsync(EntityInstanceId id, bool includeState = false, CancellationToken cancellation = default)
+    {
+        P.GetEntityRequest request = new P.GetEntityRequest()
+        {
+            InstanceId = id.ToString(),
+            IncludeState = includeState,
+        };
+
+        P.GetEntityResponse response = await this.sidecarClient.GetEntityAsync(request, cancellationToken: cancellation);
```
jviau, line 68: Can you also capture and re-propagate `OperationCanceledException`, as we do in `GrpcDurableTaskClient`: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Client/Grpc/GrpcDurableTaskClient.cs#L139

I know some places are missing it in `GrpcDurableTaskClient` - will need to fix those another time.

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            EmptyEntitiesRemoved = response.EmptyEntitiesRemoved,
+            OrphanedLocksReleased = response.OrphanedLocksReleased,
+        };
+    }
+
+    EntityMetadata ToEntityMetadata(P.EntityMetadata metadata, bool includeState)
+    {
+        var coreEntityId = DTCore.Entities.EntityId.FromString(metadata.InstanceId);
+        var entityId = new EntityInstanceId(coreEntityId.Name, coreEntityId.Key);
```
jviau, line 137: nit:
```suggestion
        EntityInstanceId entityId = new(coreEntityId.Name, coreEntityId.Key);
```

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+    readonly GrpcDurableTaskClient durableTaskClient;
+    readonly ILogger logger;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="GrpcDurableEntityClient"/> class.
+    /// </summary>
+    /// <param name="durableTaskClient">The durable Task client.</param>
+    /// <param name="sidecarClient">The client for the GRPC connection to the sidecar.</param>
+    /// <param name="logger">The logger for logging client requests.</param>
+    public GrpcDurableEntityClient(GrpcDurableTaskClient durableTaskClient, TaskHubSidecarServiceClient sidecarClient, ILogger logger)
```
jviau, line 32: The only thing used from `GrpcDurableTaskClient` is the `DataConverter` - should we just take that in directly?

src/Client/Grpc/GrpcDurableEntityClient.cs
sebastianburckhardt: Good catch. This should instead be testing `response.Exists`.

src/Client/Grpc/GrpcDurableEntityClient.cs
sebastianburckhardt: Sure, why not.

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// <inheritdoc/>
+    public override async Task SignalEntityAsync(EntityInstanceId id, string operationName, object? input = null, SignalEntityOptions? options = null, CancellationToken cancellation = default)
+    {
+        Guid requestId = Guid.NewGuid();
+        DateTimeOffset? scheduledTime = options?.SignalTime;
+
+        P.SignalEntityRequest request = new P.SignalEntityRequest()
+        {
+            InstanceId = id.ToString(),
+            RequestId = Google.Protobuf.ByteString.FromStream(new MemoryStream(requestId.ToByteArray())),
```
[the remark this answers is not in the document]
sebastianburckhardt: A string encoding of a guid consumes a lot more space than its bytes.

src/Client/Grpc/GrpcDurableEntityClient.cs
sebastianburckhardt: Since it converts to `ScheduledTime?`, not `ScheduledTime`, I have to use `scheduledTime.ToTimestamp() ?? default`. Still looks nicer.

src/Client/Grpc/GrpcDurableEntityClient.cs
sebastianburckhardt: Yes. I am assuming this should be done consistently every time I call the sidecar.

src/Client/Core/Entities/CleanEntityStorage.cs
```diff
@@ -39,7 +44,12 @@ namespace Microsoft.DurableTask.Client.Entities;
...
-    /// Gets the number of orphaned locks that were removed.
+    /// Gets the number of orphaned locks that were released.
+    /// </summary>
+    public int OrphanedLocksReleased { get; init; }
+
+    /// <summary>
+    /// Gets the continuation token to continue the <see cref="CleanEntityStorageRequest"/>, if not null.
     /// </summary>
-    public int OrphanedLocksRemoved { get; init; }
+    public string? ContinuationToken { get; init; }
```
jviau, line 54: Do we want to expose this to customers? Or do we want to abstract away the continuation behavior? As in, the gRPC protocol and implementation will handle continuation under the hood, only returning when it has exhausted continuations.

If you feel continuation must be exposed, we should at least offer an extension method or something that will loop through continuation. I wouldn't want every time a customer uses this API for them to have to include a `while` loop.

sebastianburckhardt: There are some situations where exposing the continuation is desirable (e.g. user is running a durable orchestration to perform an extremely-long-running operation, so it can resume gracefully if interrupted).

But yeah, for the common case we do want to have the automatically-continuing one.

sebastianburckhardt: I will add an optional argument to control whether the query should always continue until complete, or return a partial result with a continuation token.

jviau: I'll approve for now, but still not super happy with this.

I wonder if a custom `IAsyncEnumerable<CleanEntityStorageResult>`-esque type is the appropriate way? Then going one by one is as simple as iteration over the enumerable. Or going over all of them is an `AggregateAsync()`

So it would be a type like:

``` CSharp
public abstract class AggregateRequest<TResult>
{
    public abstract Task<TResult?> NextAsync(); // make a single call

    public abstract Task<TResult?> AllAsync(); // invoke all of them
}
```

Just a thought, can be done at a later time (but before GA at least)

jviau (approved):

src/Client/Core/Entities/CleanEntityStorage.cs
sebastianburckhardt: Yeah, I was also first considering something like `AsyncPageable` or `AsyncEnumerable`. However, I decided against it because I can really only see two common ways in which I expect this would be used (1) do all the work at once and wait for the final result (I expect this to be the most common so it should be the easiest to use) (2) do only one page (if users want to persist intermediate progress). For neither of those it provides significant value to return a pageable or enumerable result.

jviau: This isn't blocking for the PR. We can think about it more later.

PR #186, opened 2023-09-06
Introduce entity work items and connect them to ITaskEntity definition.

sebastianburckhardt:
Defines entity work items as a third work item type, and implements the necessary shim classes to connect it to ITaskEntity and the user-facing entity SDK.


src/Worker/Core/Shims/TaskEntityShim.cs
jviau, line 1: It took me a bit to read and understand the commit/rollback logic here. At first I was surprised it was needed at all, but I get it now. Two suggestions for this:

1. Add a lengthy comment somewhere explaining what it is and why it is needed. (IE: on exception we need to drop any performed actions and state changes).
2. Code organization, consider the following:
    1. We want to roll back state & context, so lets add the rollback logic directly into those.
    2. The outer types `Commit()` and `Rollback()` methods simply call `context.Commit(); state.Commit();` and `context.Rollback(); state.Rollback();` respectively. No need for the outer class to understand what the exact commit/rollback logic is.

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -0,0 +1,240 @@
...
+        this.context = new ContextShim(this, dataConverter);
+        this.operation = new OperationShim(this);
+    }
+
+    /// <inheritdoc />
+    public override async Task<EntityBatchResult> ExecuteOperationBatchAsync(EntityBatchRequest operations)
+    {
+        // initialize/reset the state and action list
+        this.state.CurrentState = operations.EntityState;
+        this.context.Rollback(0);
```
jviau, line 54: If you go with the commit/rollback suggestion of mine, can make this a `Reset()`.

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -0,0 +1,240 @@
...
+                    Result = null,
+                    ErrorMessage = "exception in application code",
+                    FailureDetails = new FailureDetails(applicationException),
+                });
+
+                this.Rollback();
+            }
+        }
+
+        this.ClearCheckpoint(); // want to ensure the old state can be GC'd even if this shim is being cached
```
jviau, line 83: If we want to manage GC, maybe all the shim objects can be created at the start of `ExecuteOperationBatchAsync` and then let them go out of scope and be GC'd entirely.

I personally don't think optimizing to re-use these objects between batches is worth the added complexity.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -337,5 +345,69 @@ async Task OnRunActivityAsync(P.ActivityRequest request)
 
             await this.sidecar.CompleteActivityTaskAsync(response);
         }
+
+        async Task OnRunEntityBatchAsync(P.EntityBatchRequest request)
+        {
+            string instanceId = request.InstanceId;
+            var coreEntityId = DTCore.Entities.EntityId.FromString(request.InstanceId);
+            var entityId = new EntityId(coreEntityId.Name, coreEntityId.Key);
```
jviau, line 353: nit:
```suggestion
            EntityId entityId = new(coreEntityId.Name, coreEntityId.Key);
```

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -337,5 +345,69 @@ async Task OnRunActivityAsync(P.ActivityRequest request)
...
+                else
+                {
+                    // we could not find the entity. This is considered an application error,
+                    // so we return a non-retriable error-OperationResult for each operation in the batch.
+                    batchResult = new EntityBatchResult()
+                    {
+                        Actions = new List<OperationAction>(), // no actions
+                        EntityState = batchRequest.EntityState, // state is unmodified
+                        Results = Enumerable.Repeat(
+                            new OperationResult()
```
jviau, line 380: Small optimization: create the result object once, then add it N times.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -337,5 +345,69 @@ async Task OnRunActivityAsync(P.ActivityRequest request)
...
+                                    errorMessage: $"No entity task named '{name}' was found.",
+                                    stackTrace: null,
+                                    innerFailure: null,
+                                    isNonRetriable: true),
+                            },
+                            batchRequest.Operations!.Count).ToList(),
+                    };
+                }
+            }
+            catch (Exception frameworkException)
```
jviau, line 395: This could use a log statement

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -0,0 +1,240 @@
...
+        // initialize/reset the state and action list
+        this.state.CurrentState = operations.EntityState;
+        this.context.Rollback(0);
+
+        List<OperationResult> results = new List<OperationResult>();
+
+        for (int i = 0; i < operations.Operations!.Count; i++)
+        {
+            OperationRequest currentOperation = operations.Operations![i];
+            this.operation.SetNameAndInput(currentOperation.Operation!, currentOperation.Input);
```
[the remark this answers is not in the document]
sebastianburckhardt: This was a purposeful optimization to reduce allocations. From a code elegance perspective, using a new would obviously be nicer but I think we should try to execute the inner loop (operations in a batch) as efficiently as possible.

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -0,0 +1,240 @@
...
+                object? result = await this.taskEntity.RunAsync(this.operation);
+                string? serializedResult = this.dataConverter.Serialize(result);
+                results.Add(new OperationResult() { Result = serializedResult });
+            }
+            catch (Exception applicationException)
+            {
+                results.Add(new OperationResult()
+                {
+                    Result = null,
+                    ErrorMessage = "exception in application code",
```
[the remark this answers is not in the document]
sebastianburckhardt: ErrorMessage would not be there if not for legacy compatibility with `OperationResult` as defined in DT.Core (where FailureDetails may not always be present, based on configuration). So I basically had to just make up some non-null string to conform to DT.Core interpretation of `OperationResult`. 

Given that, I figured I may as well use a string that conveys information not already there.

I am not sure if it is worth the trouble to use Http or Grpc names... there is no intention of ever making this machine-interpretable, it only appears in trace messages, and making up my own strings is always going to be more descriptive.

src/Worker/Core/Shims/TaskEntityShim.cs
sebastianburckhardt: I think it is worth it if the batches are large and the operations are very short.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
sebastianburckhardt: Good point. Logging is actually one of the missing things all across these PRs. I wasn't sure how much / what logging to add in general.

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -0,0 +1,240 @@
...
+        this.dataConverter = Check.NotNull(dataConverter);
+        this.taskEntity = Check.NotNull(taskEntity);
+        this.entityId = new EntityInstanceId(entityId.Name, entityId.Key);
+        this.state = new StateShim(this, dataConverter);
+        this.context = new ContextShim(this, dataConverter);
+        this.operation = new OperationShim(this);
+    }
+
+    /// <inheritdoc />
+    public override async Task<EntityBatchResult> ExecuteOperationBatchAsync(EntityBatchRequest operations)
```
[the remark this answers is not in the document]
sebastianburckhardt: It doesn't quite work because it can't return the final entity state.

src/Worker/Core/Shims/TaskEntityShim.cs
sebastianburckhardt: FYI, this behavior is controllable via a parameter in DF (i.e. the rollback behavior can be turned off). I am wondering if we should try to honor this parameter here also. It adds complexity though so I would rather not.

src/Worker/Core/Shims/TaskEntityShim.cs
sebastianburckhardt: I have revised the Commit/Rollback/Reset.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
sebastianburckhardt: I think that is already what this does.

src/Worker/Core/IDurableTaskFactory.cs
```diff
@@ -34,4 +35,19 @@ public interface IDurableTaskFactory
...
+    /// <param name="name">The name of the orchestrator.</param>
+    /// <param name="serviceProvider">The service provider.</param>
+    /// <param name="entity">The entity or <c>null</c> if it does not exist.</param>
+    /// <returns>True if entity was created, false otherwise.</returns>
+    /// <remarks>
+    /// While <paramref name="serviceProvider" /> is provided here, it is not required to be used to construct
+    /// orchestrators. Individual implementations of this contract may use it in different ways. The default
+    /// implementation does not use it.
+    /// </remarks>
+    bool TryCreateEntity(
```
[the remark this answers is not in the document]
sebastianburckhardt: I have added an `IDurableTaskFactory2`.

Note that I haven't actually touched the TaskFactory implementation. This is one of the open work items.

src/Worker/Core/Shims/TaskEntityShim.cs
sebastianburckhardt: The nice part here is that the user does not really see these optimizations in any way, from the outside it looks clean.

src/Worker/Core/Shims/TaskEntityShim.cs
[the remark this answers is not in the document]
jviau: Ahh, good point.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
jviau: oh yeah, vacation brain 🤣

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
jviau: We can go through logging later I suppose

src/Worker/Core/Shims/TaskEntityShim.cs
jviau: If you want to keep it the way it is, that is fine. Not my style, but I'll live 😄

jviau (approved):

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -337,5 +345,68 @@ async Task OnRunActivityAsync(P.ActivityRequest request)
...
+                            {
+                                FailureDetails = new FailureDetails(
+                                    errorType: "EntityTaskNotFound",
+                                    errorMessage: $"No entity task named '{name}' was found.",
+                                    stackTrace: null,
+                                    innerFailure: null,
+                                    isNonRetriable: true),
+                            },
+                            batchRequest.Operations!.Count).ToList(),
+                        FailureDetails = null,
```
jviau, line 390: nit: indentation

```suggestion
                            FailureDetails = null,
```

jviau (approved):

PR #194, opened 2023-09-22
Allowing for multiple events of same name to be raised in parallel

RyanLettieri:
This change stems from the following issue discovered in dapr: https://github.com/dapr/dotnet-sdk/issues/1129


src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.DurableTask.Worker.Shims;
 /// </summary>
 sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
 {
+    // TODO: Change externalEventSources value from IEventSource to queue<IEventSource> to preserve FIFO in O(1)
```
cgillum, line 17: I'd prefer we do this as part of this PR since I'm a little worried about the implications of making this O(N). Is that something we can consider?

RyanLettieri: 100% I'll get that added in

eng/proto
tmacam, line 1: I don't think this change here was intentional, was it?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
tmacam: I would go 2 steps further and mark [`IEventSource.Next`](https://github.com/microsoft/durabletask-dotnet/blob/e2f329258d1e02a762911bd89274a9a584b6ec04/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.EventSource.cs#L17) as deprecated. It is only used in this project for this \$O(n)\$ FIFO implementation.  Using a Queue will make this code simpler to read as it would make the intent of this map explicit rather then implicit.

The only other place in this codebase it is used is at the event completion:

https://github.com/microsoft/durabletask-dotnet/blob/e2f329258d1e02a762911bd89274a9a584b6ec04/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs#L314-L322

I don't know if it is used externally, hence why marking it as deprecated is the way to go if this is a stable API.

cgillum: `IEventSource` is a private internal class, so we don't have to worry about making breaking changes to it. We can just go ahead and remove the `Next` property.

cgillum (approved):
LGTM!

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -195,19 +195,21 @@ public override Task<T> WaitForExternalEvent<T>(string eventName, CancellationTo
 
         // Create a task completion source that will be set when the external event arrives.
         EventTaskCompletionSource<T> eventSource = new();
-        if (this.externalEventSources.TryGetValue(eventName, out IEventSource? existing))
+        if (this.externalEventSources.TryGetValue(eventName, out Queue<IEventSource>? existing))
         {
-            if (existing.EventType != typeof(T))
+            if (existing.Peek().EventType != typeof(T))
```
tmacam, line 200: It should never be empty but, let's make this refactor+NPE-safe

```suggestion
            if (existing.Count > 0 && existing.Peek().EventType != typeof(T))
```

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -195,19 +195,21 @@ public override Task<T> WaitForExternalEvent<T>(string eventName, CancellationTo
...
             {
                 throw new ArgumentException("Events with the same name must have the same type argument. Expected"
-                    + $" {existing.EventType.FullName}.");
+                    + $" {existing.Peek().GetType().FullName}.");
```
tmacam, line 203:
```suggestion
                    + $" {existing.Peek().GetType().FullName} but was requested {typeof(T).FullName}");
```

tmacam (approved):
LGTM but some minor fixes

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.EventSource.cs
jviau, line 1: Need to replace the `<inheritdoc />` on line 36 with a `<summary></summary>` comment because you removed the property from the interface.

Although there was mentions of removing `Next`, can it be removed from `EventTaskCompletionSource<T>` as well?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.DurableTask.Worker.Shims;
 /// </summary>
 sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
 {
-    readonly Dictionary<string, IEventSource> externalEventSources = new(StringComparer.OrdinalIgnoreCase);
+    readonly Dictionary<string, Queue<IEventSource>> externalEventSources = new(StringComparer.OrdinalIgnoreCase);
```
jviau, line 17: I am not sure what the typical event usage is, but I am going to hazard a guess single events are common. In which case allocating a queue for it is not optimal. We will have to revisit this and see if there is a way to optimize for a few element count cases.

jviau (approved):
Please address the warnings before checking in.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
cgillum: Agreed we should revisit this. It turns out we do the same thing with the `NamedQueue<string>` class for buffered events, so we should consider optimizing both.

PR #195, opened 2023-09-22
Refactor EntityMetadata to improve state consumption

jviau:
This PR changes to following:

1. `includeState` is `true` by default for all get entity methods (including get all)
2. Added `EntityMetadata<TState>`, which has a strongly typed deserialized `TState State` property.
3. Extracted un-typed state into `SerializedData`, which has a `ReadAs<T>()` method.
4. `EntityMetadata` now implements `EntityMetadata<SerializedData>`
5. `EntityMetadata<TState>` now has a `bool IncludesState` which indicates if state was retrieved or not.
6. `EntityMetadata<TState>.State` now throws if `IncludeState==false`
7. Added `GetEntityAsync<T>` and `GetAllEntitiesAsync<T>` overloads to get `EntityMetadata<T>`

### DIFF
--- a/src/Client/Core/Entities/DurableEntityClient.cs
+++ b/src/Client/Core/Entities/DurableEntityClient.cs
@@ -77,7 +77,7 @@ public abstract class DurableEntityClient
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="includeState"><c>true</c> to include entity state in the response, <c>false</c> to not.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>A response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
     public abstract Task<EntityMetadata?> GetEntityAsync(
         EntityInstanceId id, bool includeState = true, CancellationToken cancellation = default);
 
@@ -86,7 +86,7 @@ public abstract class DurableEntityClient
     /// </summary>
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>A response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
     public virtual Task<EntityMetadata?> GetEntityAsync(
         EntityInstanceId id, CancellationToken cancellation)
             => this.GetEntityAsync(id, includeState: true, cancellation);
@@ -98,7 +98,7 @@ public abstract class DurableEntityClient
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="includeState"><c>true</c> to include entity state in the response, <c>false</c> to not.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>A response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
     public abstract Task<EntityMetadata<T>?> GetEntityAsync<T>(
         EntityInstanceId id, bool includeState = true, CancellationToken cancellation = default);
 
@@ -108,7 +108,7 @@ public abstract class DurableEntityClient
     /// <typeparam name="T">The type of the entity state.</typeparam>
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>A response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
     public virtual Task<EntityMetadata<T>?> GetEntityAsync<T>(
         EntityInstanceId id, CancellationToken cancellation)
             => this.GetEntityAsync<T>(id, includeState: true, cancellation);
--- a/src/Client/Core/Entities/DurableEntityClient.cs
+++ b/src/Client/Core/Entities/DurableEntityClient.cs
@@ -72,24 +72,46 @@ public abstract class DurableEntityClient
         => this.SignalEntityAsync(id, operationName, null, null, cancellation);
 
     /// <summary>
-    /// Tries to get the entity with ID of <paramref name="id"/>.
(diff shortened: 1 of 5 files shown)


src/Client/Core/Entities/DurableEntityClient.cs
```diff
@@ -72,24 +72,46 @@ protected DurableEntityClient(string name)
...
 
     /// <summary>
-    /// Tries to get the entity with ID of <paramref name="id"/>.
+    /// Tries to get the entity with ID of <paramref name="id"/>. Includes entity state by default.
     /// </summary>
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="includeState"><c>true</c> to include entity state in the response, <c>false</c> to not.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>a response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity.</returns>
```
sebastianburckhardt, line 80:
```suggestion
    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
```

src/Client/Core/Entities/DurableEntityClient.cs
```diff
@@ -72,24 +72,46 @@ protected DurableEntityClient(string name)
...
+        EntityInstanceId id, bool includeState = true, CancellationToken cancellation = default);
 
     /// <summary>
-    /// Tries to get the entity with ID of <paramref name="id"/>.
+    /// Tries to get the entity with ID of <paramref name="id"/>. Includes entity state.
     /// </summary>
     /// <param name="id">The ID of the entity to get.</param>
     /// <param name="cancellation">The cancellation token to cancel the operation.</param>
-    /// <returns>a response containing metadata describing the entity.</returns>
+    /// <returns>A response containing metadata describing the entity.</returns>
```
sebastianburckhardt, line 89:
```suggestion
    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
```

src/Client/Core/Entities/DurableEntityClient.cs
```diff
@@ -72,24 +72,46 @@ protected DurableEntityClient(string name)
...
+            => this.GetEntityAsync(id, includeState: true, cancellation);
+
+    /// <summary>
+    /// Tries to get the entity with ID of <paramref name="id"/>. Includes entity state by default.
+    /// </summary>
+    /// <typeparam name="T">The type of the entity state.</typeparam>
+    /// <param name="id">The ID of the entity to get.</param>
+    /// <param name="includeState"><c>true</c> to include entity state in the response, <c>false</c> to not.</param>
+    /// <param name="cancellation">The cancellation token to cancel the operation.</param>
+    /// <returns>A response containing metadata describing the entity.</returns>
```
sebastianburckhardt, line 101:
```suggestion
    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
```

src/Client/Core/Entities/DurableEntityClient.cs
```diff
@@ -72,24 +72,46 @@ protected DurableEntityClient(string name)
...
+    public abstract Task<EntityMetadata<T>?> GetEntityAsync<T>(
+        EntityInstanceId id, bool includeState = true, CancellationToken cancellation = default);
+
+    /// <summary>
+    /// Tries to get the entity with ID of <paramref name="id"/>. Includes entity state.
+    /// </summary>
+    /// <typeparam name="T">The type of the entity state.</typeparam>
+    /// <param name="id">The ID of the entity to get.</param>
+    /// <param name="cancellation">The cancellation token to cancel the operation.</param>
+    /// <returns>A response containing metadata describing the entity.</returns>
```
sebastianburckhardt, line 111:
```suggestion
    /// <returns>A response containing metadata describing the entity, or null if the entity does not exist.</returns>
```

src/Client/Core/Entities/EntityMetadata.cs
```diff
@@ -30,38 +47,41 @@ public EntityMetadata(EntityInstanceId id)
...
-    public DataConverter? DataConverter { get; init; }
-
-    /// <summary>
-    /// Gets the serialized state for this entity.
+    /// Gets a value indicating if entity metadata 
     /// </summary>
-    public string? SerializedState { get; init; }
+    [MemberNotNullWhen(true, "State")]
+    [MemberNotNullWhen(true, "state")]
+    public bool IncludesState { get; }
```
sebastianburckhardt, line 54: we could perhaps implement this more simply as 
```csharp
public bool IncludesState => this.state is not null;
```

src/Client/Core/Entities/EntityMetadata.cs
```diff
@@ -30,38 +47,41 @@ public EntityMetadata(EntityInstanceId id)
...
 
     /// <summary>
-    /// Deserializes the entity's state into an object of the specified type.
+    /// Gets the state for this entity.
     /// </summary>
```
sebastianburckhardt, line 58:
```suggestion
    /// </summary>
    /// <remarks>
    /// This method can only be used when <see cref="IncludeState"/>=true, meaning 
    /// that the entity state was included in the response returned by the query.
    /// </remarks>
    /// <exception cref="InvalidOperationException">
    /// Thrown if this metadata object was fetched without including the entity state.
    /// </exception>
```

src/Client/Core/Entities/EntityMetadata.cs
```diff
@@ -30,38 +47,41 @@ public EntityMetadata(EntityInstanceId id)
     public DateTimeOffset LastModifiedTime { get; init; }
 
     /// <summary>
-    /// Gets the data converter for this metadata.
-    /// </summary>
-    public DataConverter? DataConverter { get; init; }
-
-    /// <summary>
-    /// Gets the serialized state for this entity.
+    /// Gets a value indicating if entity metadata
```
sebastianburckhardt, line 50:
```suggestion
    /// Gets a value indicating whether this metadata response includes the entity state.
    /// Optionally, queries can exclude the state of the entity from the metadata that is retrieved.
```

sebastianburckhardt (approved):
some suggested changes to the comments, and one question about a possible simplification.
Otherwise looks good to me.

PR #200, opened 2023-09-25
Fix ADO build when binlog=false

jviau:
Ensures `$(build_binlog)` and `$(pkg_binlog)` are set to empty string when the build is enqueued with `binlog == false`. If we leave them unset then ADO does not replace them and places the raw value of "$(build_binlog)" when it is used, which fails the build due to unknown args.


cgillum:
@jviau Can you add a short description explaining what issue this PR fixes?

PR #204, opened 2023-10-04
Revise queries

sebastianburckhardt:
Several changes (they are separated by commits):

- rename StartOrchestration to ScheduleNewOrchestration
- specify different defaults for clean entity storage (should clean everything by default)
- support queries to (optionally) return stateless entities, and add more metadata to the returned entity metadata
- make the page size a nullable int and keep it null if not specified (allows backend to choose an appropriate page size instead of using an arbitrary default)

### DIFF
--- a/src/Client/Core/Entities/CleanEntityStorage.cs
+++ b/src/Client/Core/Entities/CleanEntityStorage.cs
@@ -6,18 +6,18 @@ namespace Microsoft.DurableTask.Client.Entities;
 /// <summary>
 /// Request struct for <see cref="DurableEntityClient.CleanEntityStorageAsync"/>.
 /// </summary>
-public readonly record struct CleanEntityStorageRequest
+public record CleanEntityStorageRequest
 {
     /// <summary>
-    /// Gets a value indicating whether to remove empty entities.
+    /// Gets a value indicating whether to remove empty entities. Defaults to true.
     /// </summary>
     /// <remarks>
     /// An entity is considered empty, and is removed, if it has no state, is not locked.
     /// </remarks>
-    public bool RemoveEmptyEntities { get; init; }
+    public bool RemoveEmptyEntities { get; init; } = true;
 
     /// <summary>
-    /// Gets a value indicating whether to release orphaned locks or not.
+    /// Gets a value indicating whether to release orphaned locks or not. Defaults to true.
     /// </summary>
     /// <remarks>
     /// Locks are considered orphaned, and are released, and if the orchestration that holds them is not in state
@@ -25,10 +25,10 @@ public readonly record struct CleanEntityStorageRequest
     /// occur if the orchestration instance holding the lock exhibits replay nondeterminism failures, or if it is
     /// explicitly purged.
     /// </remarks>
-    public bool ReleaseOrphanedLocks { get; init; }
+    public bool ReleaseOrphanedLocks { get; init; } = true;
 
     /// <summary>
-    /// Gets the continuation token to resume a previous <see cref="CleanEntityStorageRequest"/>.
+    /// Gets the continuation token to resume a previous <see cref="CleanEntityStorageRequest"/>. Defaults to null.
     /// </summary>
     public string? ContinuationToken { get; init; }
 }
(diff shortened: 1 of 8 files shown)


src/Client/Core/Entities/CleanEntityStorage.cs
```diff
@@ -6,29 +6,29 @@ namespace Microsoft.DurableTask.Client.Entities;
...
     /// </remarks>
-    public bool RemoveEmptyEntities { get; init; }
+    public bool RemoveEmptyEntities { get; init; } = true;
 
     /// <summary>
-    /// Gets a value indicating whether to release orphaned locks or not.
+    /// Gets a value indicating whether to release orphaned locks or not. Defaults to true.
     /// </summary>
     /// <remarks>
     /// Locks are considered orphaned, and are released, and if the orchestration that holds them is not in state
```
jviau, line 23:
```suggestion
    /// Locks are considered orphaned, and are released, if the orchestration that holds them is not in state
```

src/Client/Core/Entities/CleanEntityStorage.cs
```diff
@@ -6,29 +6,29 @@ namespace Microsoft.DurableTask.Client.Entities;
 /// <summary>
 /// Request struct for <see cref="DurableEntityClient.CleanEntityStorageAsync"/>.
 /// </summary>
-public readonly record struct CleanEntityStorageRequest
+public record CleanEntityStorageRequest
```
jviau, line 9: I am not a big fan of changing to a class here, but not sure what a good alternative would be.

jviau (commented):
Have some warnings to address due to the struct to class change.

src/Client/Core/Entities/EntityQuery.cs
```diff
@@ -61,9 +61,18 @@ public record EntityQuery
...
     /// <summary>
-    /// Gets the size of each page to return.
+    /// Gets a value indicating whether to include metadata about entities that have no user-defined state. Defaults to false.
     /// </summary>
-    public int PageSize { get; init; } = DefaultPageSize;
+    /// <remarks> Stateless entities occur when the storage provider is tracking metadata about an entity for synchronization purposes
+    /// even though the entity does not "logically" exist, in the sense that it has no application-defined state.
+    /// Stateless entities are usually transient. For example, they may be in the process of being created or deleted, or they may have been locked by a critical section.
+    /// </remarks>
+    public bool IncludeStateless { get; init; }
```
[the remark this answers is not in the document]
sebastianburckhardt: Not sure that I prefer an enum since the two are logically independent (all four combinations are possible), and one of them is going to be relevant for most queries (IncludeState) while the other is an obscure feature likely to never be touched by most users (IncludeStateless).

src/Client/Core/Entities/CleanEntityStorage.cs
sebastianburckhardt: Right, I agree. I couldn't figure out how to get the defaults specified without turning it into a class.

sebastianburckhardt: I suppose I could define the struct without defaults and then define a static member `Default` which gets used as the default for the call. Let me try that.

sebastianburckhardt: I have added a static `Default` factory method as a way to specify the defaults and still keep this a struct.

src/Client/Core/Entities/EntityQuery.cs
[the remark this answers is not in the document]
sebastianburckhardt: I have renamed IncludeStateless to IncludeTransient as discussed.

src/Client/Core/Entities/CleanEntityStorage.cs
```diff
@@ -9,15 +9,26 @@ namespace Microsoft.DurableTask.Client.Entities;
 public readonly record struct CleanEntityStorageRequest
 {
     /// <summary>
-    /// Gets a value indicating whether to remove empty entities.
+    /// Gets the default request parameters. The default is meant to represent
+    /// "maximal" cleaning that is safe to call at all times.
+    /// </summary>
+    public static CleanEntityStorageRequest Default => new()
```
jviau, line 15: You could also add a parameterless constructor and set all the bool properties to `true` there. This doesn't change `default` - but just an idea for you to consider.

jviau (approved):

PR #211, opened 2023-10-10
Update samples with new packages

jviau:
- Updates packages
- Combines `Counter.State` and `Counter.Entity` into single file with different start API.


samples/AzureFunctionsApp/Entities/Counter.cs
```diff
@@ -0,0 +1,157 @@
...
+{
+    public int Value { get; set; }
+
+    public int Add(int input) => this.Value += input;
+
+    public int OperationWithContext(int input, TaskEntityContext context)
+    {
+        // Get access to TaskEntityContext by adding it as a parameter. Can be with or without an input parameter.
+        // Order does not matter.
+        context.ScheduleNewOrchestration("SomeOrchestration", "SomeInput");
```
sebastianburckhardt, line 87: From what I can see, this operation is not actually called from anwhere, nor does it call an existing orchestration. It does serve as a syntax sample, but usually our samples demonstrate runnable scenarios. 

So it seems a bit out of place compared to the rest of the sample.

sebastianburckhardt (approved):

samples/AzureFunctionsApp/Entities/Counter.cs
jviau: Yeah, it was a syntax sample I threw together just to aid doc creation.. We can replace it with a runnable sample later.

PR #212, opened 2023-10-10
Use RawInput for ContinueAsNew

jviau:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/2589


src/Client/Core/RELEASENOTES.md
jviau, line 1: Empty files. Removing these so we only have non-empty release notes.

cgillum (approved):

PR #214, opened 2023-10-11
Update entity samples

jviau:
This PR expands upon the entity samples to better demonstrate entity APIs.


samples/AzureFunctionsApp/Greeting.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.Extensions.Logging;
+
+namespace AzureFunctionsApp;
+
+/// <summary>
+/// An example of performing the fibonacci sequence in Durable Functions. While this is both a (naive) recursive
```
sebastianburckhardt, line 13: this comment seems to be for a different sample?

samples/AzureFunctionsApp/Entities/User.cs
```diff
@@ -0,0 +1,171 @@
...
+    protected override User InitializeState(TaskEntityOperation entityOperation)
+    {
+        // No parameterless constructor, must initialize state manually.
+        return new(null!, -1);
+    }
+}
+
+public static class UserApis
+{
+    [Function("PutUser")]
```
sebastianburckhardt, line 87: It may be helpful to start this section with a set of examples for the REST invocations, e.g.

```
PUT    user/xyz?name=aang&age=10
GET    user/xyz
DELETE user/xyz
PATCH  user/xyz?age=12
POST   user/xyz/greet?message=hello
```

samples/AzureFunctionsApp/Entities/Lifetime.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.DurableTask.Client.Entities;
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace AzureFunctionsApp.Entities;
+
+/// <summary>
+/// Example showing the lifetime of an entity. An entity is initialized on the first operation it receives and then
+/// is considered deleted when <see cref="TaskEntity{TState}.State"/> is <c>null</c> at the end of an operation.
+/// </summary>
```
sebastianburckhardt, line 17: Possibly mention:
It is also possible to design an entity that remains stateless, by overriding InitializeState to return null, and never assigning a non-null state.

sebastianburckhardt (changes requested):
two small suggestions, and one mistake (wrong comment) that should be easy to fix.

samples/AzureFunctionsApp/Entities/Counter.cs
```diff
@@ -125,17 +102,27 @@ public static class CounterHelpers
...
 
-        string name = "counter";
-
-        // switch to StateCounter if ?mode=1 or ?mode=state is supplied.
+        // switch to Counter_State if ?mode=1 or ?mode=state is supplied.
+        // or to Counter_Alt if ?mode=2 or ?mode=static is supplied.
+        string name;
         string? mode = request.Query["mode"];
-        if (int.TryParse(mode, out int m) && m == 1)
+        if (int.TryParse(mode, out int m))
```
cgillum, line 109: This is a little confusing to me since I'm not sure (as a reader of samples) whether "mode" is some concept that's important for durable entities.  It looks like we're using it to specify which entity type to use. Any reason not to just accept a string value which is the entity type name directly? That would make the intent of this method more clear, I think.

samples/AzureFunctionsApp/Entities/Lifetime.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    readonly ILogger logger;
+
+    public Lifetime(ILogger<Lifetime> logger)
+    {
+        this.logger = logger;
+    }
+
+    /// <summary>
+    /// Optional property to override. When 'true', this will allow dispatching of operations to the TState object if
```
cgillum, line 28: What does "dispatching of operations to the TState object" mean? I don't really understand what this is describing.

jviau:
Will address comments in next PR

samples/AzureFunctionsApp/Entities/Lifetime.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        this.logger.LogInformation("Dispatching entity");
+        return dispatcher.DispatchAsync(this);
+    }
+
+    public void Init() { } // no op just to init this entity.
+
+    public void CustomDelete()
+    {
+        // Deleting an entity is done by null-ing out the state.
```
cgillum, line 46: Are we actually deleting the entity or just clearing its state? I wonder if we can be more explicit about what's happening.

samples/AzureFunctionsApp/Entities/Counter.cs
jviau: it is either/or:

mode=0 or mode=entity
mode=1 or mode=state
mode=2 or mode=static

I'll expand on the comments for this.

PR #217, opened 2023-10-12
Add nuget.config to ADO yml

jviau:
We are now depending on some packages from other ADO feeds as we stage our releases. This change unblocks the release build for those packages.

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -43,6 +43,8 @@ steps:
     command: restore
     verbosityRestore: Minimal
     projects: $(project)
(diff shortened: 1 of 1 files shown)


cgillum:
I might need some additional context on this. Won’t customers run into problems if we ship a package to nuget.org that depends on a package in some alternate feed?

jviau:
Oh - we won't publish to nuget.org until all dependencies are there. This change is to allow us to stage all our packages in our ADO feed ahead of time, then release them in one go.

cgillum (approved):

PR #218, opened 2023-10-12
Further refine entity samples, add .http files and launch task

jviau:
1. Addresses feedback from previous samples PR
2. Adds more HTTP endpoints for `counter` and `lifetime` examples
3. Adds `.http` files to show quick ways to call these examples.
4. Adds `launch.json` and `tasks.json` for running this sample in vscode


samples/AzureFunctionsApp/Entities/Lifetime.cs
```diff
@@ -25,8 +27,10 @@ public Lifetime(ILogger<Lifetime> logger)
     }
 
     /// <summary>
-    /// Optional property to override. When 'true', this will allow dispatching of operations to the TState object if
-    /// there is no matching method on the entity. Default is 'false'.
+    /// Optional property to override. When 'true', this will allow dispatching of operations to the <see cref="State">
+    /// object if there is no matching method on the entity. Default is 'false'.
+    /// Stated differently: if this is <c>true</c> and there is no matching method for a given operation on the entity
+    /// type, then the operation will attempt to find a matching method on <typeparamref name="TState"/> instead.
```
cgillum, line 33: I'm surprised to see this as a feature since I don't recall it ever being discussed. What are the use-cases? I'm also interested to understand the lifetime implications, etc.

samples/AzureFunctionsApp/Entities/counters.http
```diff
@@ -0,0 +1,26 @@
+@host = http://localhost:7071/api
```
cgillum, line 1: I assume these files require a VS Code extension to run? Should we put a comment in here making that clear, and perhaps pointing them to the download location?

jviau: added vscode recommended extensions.

samples/AzureFunctionsApp/Entities/Lifetime.cs
jviau: This was to help migration so customers don't _have_ to derive from `ITaskEntity` if they do not want to.

What lifetime implications are you concerned about?

cgillum: I'm wondering what happens if the state object is null, etc. I don't recall any such feature existing for the in-proc model. But maybe I'm misunderstanding what is being referred to by `TState`? Is that not the entity state?

cgillum (approved):

samples/AzureFunctionsApp/Entities/Lifetime.cs
jviau: In proc model does not have a concept of `ITaskEntity` base class.

If state is null we will use `Activator.CreateInstance`.

sebastianburckhardt: Trying to summarize the discussions and design a bit: 

For in-proc we had two models:
- **class-based**. Entity state must be defined by a class, and operations are dispatched to methods.
- **function-based**. Very low level but totally customizable, with entity state being an arbitrary type.

For isolated we have three (but only the first two are expected to be widely used):
- **class-based**. Pretty much the same as for in-proc. 
- **TaskEntity\<TState\>**. Similar to class-based but with one layer of indirection, which allows advanced scenarios (arbitrary TState types, DI injection, customizable code for constructing the state) without having to give up the convenience of the class-based approach.  
- **ITaskEntity**. Very similar to function-based. Very general but verbose. Not recommended unless you are building some kind of framework on top of entities.

PR #222, opened 2023-10-16
Add HasState and default value to GetState<T>

jviau:
This PR does the following:

1. Adds another sample on how to implement the `Counter` entity.
2. Adds `bool HasState` to `TaskEntityState`, so there is a quick way to know if an entity has any serialized state or not.
3. Adds a `T? defaultValue` to `.GetState<T>` to make the new entity scenario easier (won't have any state).
4. Reworks `TaskEntityShim.StateShim` a bit to be more robust with multiple state fetches, particularly when/if they change type.

### DIFF
--- a/samples/AzureFunctionsApp/Entities/Counter.cs
+++ b/samples/AzureFunctionsApp/Entities/Counter.cs
@@ -97,6 +97,40 @@ public class StateCounter
     }
 }
 
+public static class ManualCounter
+{
+    [Function("Counter_Manual")]
+    public static Task DispatchAsync([EntityTrigger] TaskEntityDispatcher dispatcher)
+    {
+        return dispatcher.DispatchAsync(operation =>
+        {
+            if (operation.State.GetState(typeof(int)) is null)
+            {
+                operation.State.SetState(0);
+            }
+
+            switch (operation.Name.ToLowerInvariant())
+            {
+                case "add":
+                    int state = operation.State.GetState<int>();
+                    state += operation.GetInput<int>();
+                    operation.State.SetState(state);
+                    return new(state);
+                case "reset":
+                    operation.State.SetState(0);
+                    break;
+                case "get":
+                    return new((operation.State.GetState(typeof(int)) as int?) ?? 0);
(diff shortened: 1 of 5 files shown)


src/Abstractions/Entities/TaskEntityState.cs
```diff
@@ -8,13 +8,32 @@ namespace Microsoft.DurableTask.Entities;
 /// </summary>
 public abstract class TaskEntityState
 {
+    /// <summary>
+    /// Gets a value indicating whether this entity has state or not yet.
```
sebastianburckhardt, line 12:
```suggestion
    /// Gets a value indicating whether this entity has state or not yet / anymore.
```

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -130,29 +131,30 @@ public void Commit()
         public void Rollback()
         {
             this.CurrentState = this.checkpointValue;
+            this.cachedValue = null;
```
sebastianburckhardt, line 134: I agree that clearing the cache explicitly seems like a good idea here.

(since the code may have modified the object without calling `SetState` and then the comparison in the setter does not catch that it was modified).

samples/AzureFunctionsApp/Entities/Counter.cs
```diff
@@ -97,6 +97,40 @@ public static Task DispatchAsync([EntityTrigger] TaskEntityDispatcher dispatcher
...
 }
 
+public static class ManualCounter
+{
+    [Function("Counter_Manual")]
+    public static Task DispatchAsync([EntityTrigger] TaskEntityDispatcher dispatcher)
+    {
+        return dispatcher.DispatchAsync(operation =>
+        {
+            if (operation.State.GetState(typeof(int)) is null)
```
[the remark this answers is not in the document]
sebastianburckhardt: Using the new APIs would make more sense here I think.

sebastianburckhardt (approved):
some minor suggestions, otherwise looks good to me.

PR #235, opened 2023-11-01
Unwrap original exception when dispatching entity

jviau:
Unwrap and rethrow the original exception from a failed entity operation.


sebastianburckhardt (approved):
Nice. Seems like a good idea to remove superfluous wrapping.

Just as an aside: is this something that should be done for functions other than entities? Or do you already do that?

jviau:
Only necessary for other functions if they are being invoked by reflection, which is what entities does. Orchestrations and activities will not have this issue.

PR #236, opened 2023-11-03
Add operation error log

jviau:
Adds an error log for when an entity operation has an unhandled exception.

### DIFF
--- a/src/Worker/Core/Logs.cs
+++ b/src/Worker/Core/Logs.cs
@@ -14,10 +14,10 @@ namespace Microsoft.DurableTask
     /// </remarks>
     static partial class Logs
     {
+        [LoggerMessage(EventId = 15, Level = LogLevel.Error, Message = "Unhandled exception in entity operation {entityInstanceId}/{operationName}.")]
+        public static partial void OperationError(this ILogger logger, Exception ex, EntityInstanceId entityInstanceId, string operationName);
+
         [LoggerMessage(EventId = 55, Level = LogLevel.Information, Message = "{instanceId}: Evaluating custom retry handler for failed '{name}' task. Attempt = {attempt}.")]
         public static partial void RetryingTask(this ILogger logger, string instanceId, string name, int attempt);
-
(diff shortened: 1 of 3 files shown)


src/Worker/Core/Logs.cs
```diff
@@ -15,5 +16,8 @@ static partial class Logs
     {
         [LoggerMessage(EventId = 55, Level = LogLevel.Information, Message = "{instanceId}: Evaluating custom retry handler for failed '{name}' task. Attempt = {attempt}.")]
         public static partial void RetryingTask(this ILogger logger, string instanceId, string name, int attempt);
+
+        [LoggerMessage(EventId = 80, Level = LogLevel.Error, Message = "Unhandled exception in entity operation {EntityInstanceId}/{OperationName}.")]
+        public static partial void OperationError(this ILogger logger, Exception ex, EntityInstanceId entityInstanceId, string operationName);
```
davidmrdavid, line 21: wow, we jumped a fair bit: from 55 to 80. Why?

src/Worker/Core/Shims/TaskEntityShim.cs
```diff
@@ -68,6 +72,7 @@ public override async Task<EntityBatchResult> ExecuteOperationBatchAsync(EntityB
             }
             catch (Exception applicationException)
             {
+                this.logger.OperationError(applicationException, this.entityId, current.Operation!);
```
davidmrdavid, line 75: what does the `!` operator do at the end of this expression? My guess is something related to nullability

davidmrdavid (approved):
Approved w/ questions

src/Worker/Core/Logs.cs
jviau: Our event IDs are all over the place - I was trying to avoid the IDs in DurableTask.Core

src/Worker/Core/Shims/TaskEntityShim.cs
jviau: Yeah, it suppresses a nullability warning, saying "I know this isn't null here"

src/Worker/Core/Logs.cs
jviau: Went with an ID of 15 as it is in line with the other eventIds we have (from the gRPC worker)

PR #244, opened 2023-11-14
Refactor to use primary constructors from .net8

jviau:
This PR refactors to use primary constructors when possible. There are several instances of removing the field and `this.` accessors as I am switching to rely on primary ctor closure (let the compiler decide how to capture and store those fields given by the ctor when possible). I only do this when arg checking is not needed (will already be checked before ctor or is not production code). 

Some additional refactor changes in this PR: removing unused fields/properties (we can add them back in when/if needed).


src/Abstractions/Entities/EntityOperationFailedException.cs
```diff
@@ -10,36 +10,30 @@ namespace Microsoft.DurableTask.Entities;
 /// Detailed information associated with a particular operation failure, including exception details, can be found in the
 /// <see cref="FailureDetails"/> property.
 /// </remarks>
-public sealed class EntityOperationFailedException : Exception
+/// <remarks>
+/// Initializes a new instance of the <see cref="EntityOperationFailedException"/> class.
+/// </remarks>
```
cgillum, line 15: This now has two `<remarks>` sections?

src/Abstractions/Converters/JsonDataConverter.cs
```diff
@@ -8,24 +8,19 @@ namespace Microsoft.DurableTask.Converters;
 /// <summary>
 /// An implementation of <see cref="DataConverter"/> that uses System.Text.Json APIs for data serialization.
 /// </summary>
-public class JsonDataConverter : DataConverter
+/// <remarks>
+/// Initializes a new instance of the <see cref="JsonDataConverter"/> class.
+/// </remarks>
```
cgillum, line 13: Is `<remarks>` the recommended way to describe the constructor now? How are the class remarks disambiguated from constructor remarks in the generated .NET documentation?

cgillum (commented):
Haven't reviewed everything yet (there's a ton of files 😢) but had some high-level questions.

src/Abstractions/Converters/JsonDataConverter.cs
jviau: Removed all those `<remarks>`

src/Abstractions/Entities/EntityOperationFailedException.cs
jviau: Removed all those `<remarks>`

Issue #255, opened 2023-12-12
Support orchestration ID reuse policy


kaibocai:
Reference: 
1. https://github.com/microsoft/durabletask-go/pull/46
2. https://github.com/microsoft/durabletask-java/pull/188

Issue #183, opened 2023-09-05
Allow dedupe statuses when scheduling new orchestration instance


sagimar:
Hi, we wish to achieve idempotence when calling ScheduleNewOrchestrationInstanceAsync with the same orchestration id, so if the orchestration id is already existed (in pending, running or completion state)- the orchestration won't be re-triggered.

In the current implementation, the orchestration is actually being triggered twice, even if the orchestration is in Pending/Running state, and the documentation has no real promises in this matter:

> "If an orchestration with the specified instance ID already exists and is in a non-terminal state (Pending, Running, etc.), then this operation may fail silently. However, if an orchestration instance with this ID already exists in a terminal state (Completed, Terminated, Failed, etc.) then the instance may be recreated automatically, depending on the configuration of the backend instance store."

Suggested solution:
To schedule a new orchestration instance, one should call:
```
    public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
        TaskName orchestratorName,
        object? input = null,
        StartOrchestrationOptions? options = null,
        CancellationToken cancellation = default)
```

The implementation of this method is in [ShimDurableTaskClient.cs](https://github.com/microsoft/durabletask-dotnet/blob/e2f329258d1e02a762911bd89274a9a584b6ec04/src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs#L129) and in the end of the methods it calls:
`await this.Client.CreateTaskOrchestrationAsync(message);`

This method is from azure/durabletask, and it has another overload, with the following signature:
`public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[]? dedupeStatuses)`
When one sending the dedupeStatuses to this method, if an orchestration is existed in one of the given statuses, durabletask throws InvalidOperationException, that could be caught to handle idempotence.

Then, in my opinion, allowing receiving OrchestrationStatus[]? dedupeStatuses in ScheduleNewOrchestrationInstanceAsync() should help to handle the idempotence (or any other solution you have in mind will be appreciated).

Thanks,
Sagi

davidmrdavid:
Hi @sagimar: thanks for reaching out.
I think I need more context here: when you say that "In the current implementation, the orchestration is actually being triggered twice", do you mean that you're seeing the orchestration being triggered twice one after the other, or do you mean in parallel? Just trying to get clarity around the current behavior you're seeing. Any traces that you may share will help me understand your requirements.

[...]

PR #258, opened 2024-01-05
Initial push for adding orchestration reuse ID

RyanLettieri:
Related Issue: https://github.com/microsoft/durabletask-dotnet/issues/255

### DIFF
--- a/src/Abstractions/Internal/IOrchestrationSubmitter.cs
+++ b/src/Abstractions/Internal/IOrchestrationSubmitter.cs
@@ -27,6 +27,7 @@ public interface IOrchestrationSubmitter
     /// The optional input to pass to the scheduled orchestration instance. This must be a serializable value.
     /// </param>
     /// <param name="options">The options to start the new orchestration with.</param>
+    /// <param name="orchestrationIdReusePolicy">The policy for reusing an orchestration ID.</param>
     /// <param name="cancellation">
     /// The cancellation token. This only cancels enqueueing the new orchestration to the backend. Does not cancel the
     /// orchestration once enqueued.
@@ -40,5 +41,6 @@ public interface IOrchestrationSubmitter
         TaskName orchestratorName,
         object? input = null,
         StartOrchestrationOptions? options = null,
+        HashSet<string>? orchestrationIdReusePolicy = null,
         CancellationToken cancellation = default);
 }
--- a/src/Abstractions/Internal/IOrchestrationSubmitter.cs
+++ b/src/Abstractions/Internal/IOrchestrationSubmitter.cs
@@ -27,7 +27,6 @@ public interface IOrchestrationSubmitter
     /// The optional input to pass to the scheduled orchestration instance. This must be a serializable value.
     /// </param>
     /// <param name="options">The options to start the new orchestration with.</param>
-    /// <param name="orchestrationIdReusePolicy">The policy for reusing an orchestration ID.</param>
     /// <param name="cancellation">
     /// The cancellation token. This only cancels enqueueing the new orchestration to the backend. Does not cancel the
     /// orchestration once enqueued.
@@ -41,6 +40,5 @@ public interface IOrchestrationSubmitter
         TaskName orchestratorName,
         object? input = null,
         StartOrchestrationOptions? options = null,
-        HashSet<string>? orchestrationIdReusePolicy = null,
         CancellationToken cancellation = default);
 }
--- a/src/Abstractions/OrchestrationOptions.cs
+++ b/src/Abstractions/OrchestrationOptions.cs
@@ -4,84 +4,49 @@
 namespace Microsoft.DurableTask;
 
 /// <summary>
-/// Orchestration status options.
+/// Enum describing the runtime status of the orchestration.
 /// </summary>
-///
-public sealed class OrchestrationOptions
+public enum OrchestrationRuntimeStatus
 {
     /// <summary>
-    /// Initializes a new instance of the <see cref="OrchestrationOptions"/> class.
+    /// The orchestration started running.
     /// </summary>
-    public OrchestrationOptions()
-    {
-    }
+    Running,
 
     /// <summary>
-    /// Enum describing the return value when attempting to reuse a previously used instance ID.
+    /// The orchestration completed normally.
     /// </summary>
-    public enum InstanceIdReuseAction
-    {
-        /// <summary>
-        /// An error will be returned when attempting to reuse a previously used instance ID.
-        /// </summary>
-        ERROR,
-
-        /// <summary>
-        /// The request to reuse the previously used instanceID will be ignored.
-        /// </summary>
-        IGNORE,
-
-        /// <summary>
-        /// The currently running orchestration will be terminated and a new instance will be started.
-        /// </summary>
-        TERMINATE,
-    }
+    Completed,
 
     /// <summary>
-    /// Enum describing the runtime status of the orchestration.
+    /// The orchestration is transitioning into a new instance.
     /// </summary>
-    public enum OrchestrationRuntimeStatus
-    {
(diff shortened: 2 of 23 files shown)


src/Abstractions/Internal/IOrchestrationSubmitter.cs
```diff
@@ -40,5 +41,6 @@ public interface IOrchestrationSubmitter
         TaskName orchestratorName,
         object? input = null,
         StartOrchestrationOptions? options = null,
+        HashSet<string>? orchestrationIdReusePolicy = null,
```
jviau, line 44: Please add this to `StartOrchestrationOptions` instead of its own parameter. All non-essential parameters for affecting orchestration start should go there.

jviau (changes requested):
I think you may have already updating the protobuf submodule, but was there merge issues in that repo? Looks like all the entity work is gone.

src/Abstractions/Internal/IOrchestrationSubmitter.cs
jviau, line 44: "ReusePolicy" isn't that clean to me. Was the name `InstanceIdConflictBehavior` considered?

jviau: Also, should this be an enum with the available behaviors as values in it?

jviau:
This will also address #183

jviau (changes requested):
I suggest the following approach:

1. Create a new enum in `Abstractions.csproj` to represent the possible states for this feature. This is the type the customer will interact with.
2. Add this as an _optional_ value to `StartOrchestrationOptions` **and** ~`SubOrchestrationOptions`~ (actually only worry about Start options in this PR, we can tackle sub-orchestration options later)
4. Add a configurable default value to `DurableTaskClientOptions` and `DurableTaskWorkerOptions`. This default value will be used when a customer does not explicitly provide it on the options from point 2.
5. Update clients and workers to consume this as appropriate. The gRPC client/worker implementation would convert the public facing enum to the gRPC enum internally.

src/Abstractions/TaskOptions.cs
```diff
@@ -100,4 +103,7 @@ public SubOrchestrationOptions(TaskOptions options, string? instanceId = null)
 /// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
 /// is specified, the orchestration instance will be scheduled immediately.
 /// </param>
-public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null);
+/// <param name="OrchestrationIdReusePolicy">The orchestration reuse policy. This allows for the reuse of an instance ID
+/// as well as the options for it.</param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null,
+ Dictionary<List<OrchestrationOptions.OrchestrationRuntimeStatus>, OrchestrationOptions.InstanceIdReuseAction>? OrchestrationIdReusePolicy = null);
```
jviau, line 109: I see what you are trying to solve here: making it more granular on what the behavior is based on the state of the orchestration. Which I do think is worth considering.

I see the following problems with this approach:

1. This information does not align with the protobuf spec
2. It is a bit unwieldy for customers to need to specify this dictionary.

Instead, here is what I recommend:

1. Lets first evaluate if this is needed? Or can we decide and document what the behavior of the 3 individual policies are based on the orchestration state? For example, we may consider applying conflict resolution for non-terminal orchestrations only.
2. If not, then we can expand the protobuf and our enum to have more states that better clarify what happens. IE: `ErrorAlways`, `ErrorIfRunning` eg.

The end result should be customers need to only provide a single enum to this options object.

src/Abstractions/OrchestrationOptions.cs
```diff
@@ -0,0 +1,87 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Orchestration status options.
+/// </summary>
+///
+public sealed class OrchestrationOptions
```
jviau, line 10: We don't need a top level class here. Enums can be top level.

jviau (changes requested):

src/Abstractions/OrchestrationOptions.cs
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Enum describing the runtime status of the orchestration.
+/// </summary>
+public enum OrchestrationRuntimeStatus
```
jviau, line 9: I think we will need to have type-forwarding from this type in the client package.

src/Abstractions/TaskOptions.cs
```diff
@@ -100,4 +100,7 @@ public SubOrchestrationOptions(TaskOptions options, string? instanceId = null)
 /// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
 /// is specified, the orchestration instance will be scheduled immediately.
 /// </param>
-public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null);
+/// <param name="OrchestrationIdReusePolicy">The orchestration reuse policy. This allows for the reuse of an instance ID
+/// if the instance ID referenced is in any of the states supplied in this parameter.</param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null,
+ OrchestrationRuntimeStatus[]? OrchestrationIdReusePolicy = null);
```
jviau, line 106: Can we wrap this in a type: `OrchestrationIdReuse`, so we can have helpers:

``` CSharp
OrchestrationIdReuse.IfTerminal; // has all terminal statuses included.
OrchestrationIdReuse.Always; // has all statuses included.
OrchestrationIdReuse.IfStatus(params OrchestrationRuntimeStatus[] statuses);
OrchestrationIdReuse.IfStatus(IEnumerable<OrchestrationRuntimeStatus> statuses);
OrchestrationIdReuse.Never; // equal to default.
```

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -83,6 +83,7 @@ public override ValueTask DisposeAsync()
             Version = orchestratorName.Version,
             InstanceId = options?.InstanceId ?? Guid.NewGuid().ToString("N"),
             Input = this.DataConverter.Serialize(input),
+            OrchestrationIdReusePolicy = { },
```
jviau, line 86: You need to map this from `options` correct?

src/Client/OrchestrationServiceClientShim/Client.OrchestrationServiceClientShim.csproj
```diff
@@ -17,11 +22,13 @@ The client is responsible for interacting with orchestrations from outside the w
 
   <ItemGroup>
     <ProjectReference Include="../Core/Client.csproj" />
+    <ProjectReference Include="../../Grpc/Grpc.csproj" />
```
jviau, line 25: this shouldn't be depending on gRPC

src/Client/OrchestrationServiceClientShim/Client.OrchestrationServiceClientShim.csproj
```diff
@@ -8,6 +8,11 @@ The client is responsible for interacting with orchestrations from outside the w
     <EnableStyleCop>true</EnableStyleCop>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="6.0.0" />
```
jviau, line 12: What are these packages needed for?

Issue #260, opened 2024-01-17
Support Cascade Terminate/Purge


shivamkm07:
durabletask-go now supports cascade terminate/purge. It can be enabled by passing `recursive = True` in the Terminate/Purge request. Also cascade feature(recursive=true) should be the default behavior which can be disabled by passing `recursive = False`.

The same should be enabled in `GrpcDurableTaskClient` for use.

Reference: https://github.com/microsoft/durabletask-go/pull/47

PR #262, opened 2024-01-24
Cascade terminate/purge support in GrpcDurableTaskClient

shivamkm07:
This PR adds support to recursively terminate/purge sub-orchestrations in GrpcDurableTaskClient. It also sets the recursive behavior to be true by default. 

Closes: #260

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -391,8 +391,11 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <see cref="PurgeResult.PurgedInstanceCount"/> value of <c>1</c> or <c>0</c>, depending on whether the target
     /// instance was successfully purged.
     /// </returns>
-    public abstract Task<PurgeResult> PurgeInstanceAsync(
-        string instanceId, bool recursive = true, CancellationToken cancellation = default);
+    public virtual Task<PurgeResult> PurgeInstanceAsync(
+        string instanceId, bool recursive = true, CancellationToken cancellation = default)
+    {
+        throw new NotSupportedException($"{this.GetType()} does not support purging of orchestration instances.");
+    }
 
     /// <inheritdoc cref="PurgeAllInstancesAsync(PurgeInstancesFilter, bool, CancellationToken)"/>
     public virtual Task<PurgeResult> PurgeAllInstancesAsync(PurgeInstancesFilter filter, CancellationToken cancellation)
@@ -410,8 +413,11 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// This method returns a <see cref="PurgeResult"/> object after the operation has completed with a
     /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged.
     /// </returns>
-    public abstract Task<PurgeResult> PurgeAllInstancesAsync(
-        PurgeInstancesFilter filter, bool recursive = true, CancellationToken cancellation = default);
+    public virtual Task<PurgeResult> PurgeAllInstancesAsync(
+        PurgeInstancesFilter filter, bool recursive = true, CancellationToken cancellation = default)
+    {
+        throw new NotSupportedException($"{this.GetType()} does not support purging of orchestration instances.");
+    }
 
     // TODO: Create task hub
 
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -214,30 +214,9 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
         => this.TerminateInstanceAsync(instanceId, null, cancellation);
 
     /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
-    public virtual Task TerminateInstanceAsync(string instanceId, object? output)
+    public virtual Task TerminateInstanceAsync(string instanceId, object? output, CancellationToken cancellation = default)
     {
         TerminateInstanceOptions? options = output is null ? null : new() { Output = output };
-        return this.TerminateInstanceAsync(instanceId, options);
-    }
-
-    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
-    public virtual Task TerminateInstanceAsync(string instanceId, object? output, CancellationToken cancellation)
-    {
-        TerminateInstanceOptions? options = output is null ? null : new() { Output = output };
-        return this.TerminateInstanceAsync(instanceId, options, cancellation);
-    }
-
-    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
-    public virtual Task TerminateInstanceAsync(string instanceId, bool recursive)
-    {
-        TerminateInstanceOptions options = new() { Recursive = recursive };
-        return this.TerminateInstanceAsync(instanceId, options);
-    }
-
-    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
-    public virtual Task TerminateInstanceAsync(string instanceId, bool recursive, CancellationToken cancellation)
-    {
-        TerminateInstanceOptions options = new() { Recursive = recursive };
         return this.TerminateInstanceAsync(instanceId, options, cancellation);
     }
 
@@ -378,7 +357,7 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// If <paramref name="instanceId"/> is not found in the data store, or if the instance is found but not in a
     /// terminal state, then the returned <see cref="PurgeResult"/> object will have a
     /// <see cref="PurgeResult.PurgedInstanceCount"/> value of <c>0</c>. Otherwise, the existing data will be purged and
-    /// <see cref="PurgeResult.PurgedInstanceCount"/> will be <c>countOfInstancesPurged</c>.
+    /// <see cref="PurgeResult.PurgedInstanceCount"/> will be the count of purged instances.
     /// </para>
     /// </remarks>
     /// <param name="instanceId">The unique ID of the orchestration instance to purge.</param>
@@ -388,8 +367,8 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// </param>
     /// <returns>
     /// This method returns a <see cref="PurgeResult"/> object after the operation has completed with a
-    /// <see cref="PurgeResult.PurgedInstanceCount"/> value of <c>1</c> or <c>0</c>, depending on whether the target
-    /// instance was successfully purged.
+    /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged,
+    /// including the count of sub-orchestrations purged if any.
     /// </returns>
     public virtual Task<PurgeResult> PurgeInstanceAsync(
         string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
@@ -411,7 +390,8 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// </param>
     /// <returns>
     /// This method returns a <see cref="PurgeResult"/> object after the operation has completed with a
-    /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged.
+    /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged,
+    /// including the count of sub-orchestrations purged if any.
     /// </returns>
     public virtual Task<PurgeResult> PurgeAllInstancesAsync(
         PurgeInstancesFilter filter, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -338,6 +338,10 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <returns>An async pageable of the query results.</returns>
     public abstract AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(OrchestrationQuery? filter = null);
 
+    /// <inheritdoc cref="PurgeInstanceAsync(string, bool, CancellationToken)"/>
+    public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation)
+        => this.PurgeInstanceAsync(instanceId, true, cancellation);
+
     /// <summary>
     /// Purges orchestration instance metadata from the durable store.
     /// </summary>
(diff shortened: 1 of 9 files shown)


jviau (commented):
As we now have multiple parameters for terminate, I recommend we do the following:

1. Add `TerminateInstanceOptions`
2. Make the current `abstract` method `virtual`. Default implementation will call one of the new methods we add.
3. Add `TerminateInstanceOptions` accepting `virtual` methods
4. Update our clients to override the new options-receiving method.

``` CSharp
/// add xmldoc
public record TerminateInstanceOptions
{
    public string? Output { get; init; }

    public bool Recursive { get; init; }
}
```

new methods:

``` CSharp
public virtual Task TerminateInstanceAsync(string instanceId, TerminateInstanceOptions? options, CancellationToken cancellation)
    => throw new NotSupportedException($"{this.GetType()} does not support orchestration termination.");

// this is the formerly abstract method
public virtual Task TerminateInstanceAsync(string instanceId, object? output = null, CancellationToken cancellation = default)
{
    TerminateInstanceOptions? options = output is null ? null : new() { Output = output };
    return this.TerminateInstanceAsync(instanceId, options, cancellation);
}
```

src/Client/Core/DurableTaskClient.cs
```diff
@@ -238,13 +248,14 @@ public virtual Task TerminateInstanceAsync(string instanceId, CancellationToken
...
     /// <param name="output">The optional output to set for the terminated orchestration instance.</param>
+    /// <param name="recursive">The optional parameter to recursively termintate sub-orchestrations, true by default.</param>
     /// <param name="cancellation">
     /// The cancellation token. This only cancels enqueueing the termination request to the backend. Does not abort
     /// termination of the orchestration once enqueued.
     /// </param>
     /// <returns>A task that completes when the terminate message is enqueued.</returns>
     public abstract Task TerminateInstanceAsync(
-        string instanceId, object? output = null, CancellationToken cancellation = default);
+        string instanceId, object? output = null, bool recursive = true, CancellationToken cancellation = default);
```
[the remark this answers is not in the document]
shivamkm07: A separate virtual method is added, with same definition as previous abstract method. So anyone using this abstract method should work still now right? How is this a breaking change then?
 ```
public virtual Task TerminateInstanceAsync(string instanceId, object? output, CancellationToken cancellation)
        => this.TerminateInstanceAsync(instanceId, output, true, cancellation);

```

**Comment by shivamkm07** [comment]
> As we now have multiple parameters for terminate, I recommend we do the following:
> 
> 1. Add `TerminateInstanceOptions`
> 2. Make the current `abstract` method `virtual`. Default implementation will call one of the new methods we add.
> 3. Add `TerminateInstanceOptions` accepting `virtual` methods
> 4. Update our clients to override the new options-receiving method.
> 
> ```cs
> /// add xmldoc
> public record TerminateInstanceOptions
> {
>     public object? Output { get; init; }
> 
>     public bool Recursive { get; init; }
> }
> ```
> 
> new methods:
> 
> ```cs
> public virtual Task TerminateInstanceAsync(string instanceId, TerminateInstanceOptions? options, CancellationToken cancellation)
>     => throw new NotSupportedException($"{this.GetType()} does not support orchestration termination.");
> 
> // this is the formerly abstract method
> public virtual Task TerminateInstanceAsync(string instanceId, object? output = null, CancellationToken cancellation = default)
> {
>     TerminateInstanceOptions? options = output is null ? null : new() { Output = output };
>     return this.TerminateInstanceAsync(instanceId, options, cancellation);
> }
> ```

I am okay with this change. Just should not the new `TerminateInstanceOptions` method be an abstract method and both `GrpcDurableTaskClient` and `ShimDurableTaskClient` implement this method? This is what seems to be followed in other `DurableTaskClient` methods..

File: src/Client/Core/DurableTaskClient.cs (hunk 1 above)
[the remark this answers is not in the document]
Comment (jviau) [reply src/Client/Core/DurableTaskClient.cs:258]: Correct on the `virtual` method meaning the old abstract method is covered. But now there is a _new_ abstract method which is a breaking change on its own.

File: src/Client/Core/DurableTaskClient.cs [hunk 2]
```diff
@@ -368,7 +392,7 @@ public virtual Task ResumeInstanceAsync(string instanceId, CancellationToken can
     /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged.
     /// </returns>
     public abstract Task<PurgeResult> PurgeAllInstancesAsync(
-        PurgeInstancesFilter filter, CancellationToken cancellation = default);
+        PurgeInstancesFilter filter, bool recursive = true, CancellationToken cancellation = default);
```
jviau, line 395: Same issue with the other comment: adding a net-new abstract method is a breaking change.

jviau (changes requested):

src/Client/Core/DurableTaskClient.cs
shivamkm07: Changed to virtual

src/Client/Core/DurableTaskClient.cs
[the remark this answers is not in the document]
shivamkm07: Done

shivamkm07:
> As we now have multiple parameters for terminate, I recommend we do the following:
> 
> 1. Add `TerminateInstanceOptions`
> 2. Make the current `abstract` method `virtual`. Default implementation will call one of the new methods we add.
> 3. Add `TerminateInstanceOptions` accepting `virtual` methods
> 4. Update our clients to override the new options-receiving method.
> 
> ```cs
> /// add xmldoc
> public record TerminateInstanceOptions
> {
>     public object? Output { get; init; }
> 
>     public bool Recursive { get; init; }
> }
> ```
> 
> new methods:
> 
> ```cs
> public virtual Task TerminateInstanceAsync(string instanceId, TerminateInstanceOptions? options, CancellationToken cancellation)
>     => throw new NotSupportedException($"{this.GetType()} does not support orchestration termination.");
> 
> // this is the formerly abstract method
> public virtual Task TerminateInstanceAsync(string instanceId, object? output = null, CancellationToken cancellation = default)
> {
>     TerminateInstanceOptions? options = output is null ? null : new() { Output = output };
>     return this.TerminateInstanceAsync(instanceId, options, cancellation);
> }
> ```

@jviau Pushed the suggested changes. Please review.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -124,19 +124,25 @@ public override ValueTask DisposeAsync()
 
     /// <inheritdoc/>
     public override async Task TerminateInstanceAsync(
-        string instanceId, object? output = null, CancellationToken cancellation = default)
+        string instanceId, TerminateInstanceOptions? options = null, CancellationToken cancellation = default)
     {
+        if (options is null)
+        {
+            options = new TerminateInstanceOptions();
```
jviau, line 131: I'd prefer we avoid allocating unnecessary objects like this. Instead you can either:

1. Have a `static` empty instance of `TerminateInstanceOptions`
2. OR split into defaults in this method:

``` CSharp
string? output = options?.Output;
bool recursive = options?.Recursive ?? false;

// rest of function
```

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -169,11 +169,16 @@ public override AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(Orches
 
     /// <inheritdoc/>
     public override Task TerminateInstanceAsync(
-        string instanceId, object? output = null, CancellationToken cancellation = default)
+        string instanceId, TerminateInstanceOptions? options = null, CancellationToken cancellation = default)
     {
+        if (options is null)
+        {
+            options = new TerminateInstanceOptions();
```
jviau, line 176: Same here as the other comment: avoid unnecessary object allocation

jviau (changes requested):

src/Client/Core/DurableTaskClient.cs
```diff
@@ -353,22 +391,33 @@ public virtual Task ResumeInstanceAsync(string instanceId, CancellationToken can
...
     /// A <see cref="CancellationToken"/> that can be used to cancel the purge operation.
     /// </param>
     /// <returns>
     /// This method returns a <see cref="PurgeResult"/> object after the operation has completed with a
     /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged.
     /// </returns>
-    public abstract Task<PurgeResult> PurgeAllInstancesAsync(
-        PurgeInstancesFilter filter, CancellationToken cancellation = default);
+    public virtual Task<PurgeResult> PurgeAllInstancesAsync(
+        PurgeInstancesFilter filter, bool recursive = true, CancellationToken cancellation = default)
```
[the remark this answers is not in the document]
shivamkm07: Added `PurgeInstanceOptions` But haven't inherited from `PurgeInstancesFilter` because it is being used by both `PurgeInstanceAsync` and `PurgeAllInstancesAsync` while `PurgeInstancesFilter` is just used in `PurgeAllInstancesAsync`.

src/Client/Core/DurableTaskClient.cs
```diff
@@ -338,13 +372,17 @@ public virtual Task ResumeInstanceAsync(string instanceId, CancellationToken can
...
     /// <see cref="OrchestrationRuntimeStatus.Terminated"/> state can be purged.
     /// </para><para>
+    /// Purging an orchestration will by default purge all of the child sub-orchestrations that were started by the
+    /// orchetration instance. If you don't want to purge sub-orchestration instances, you can set `recursive` flag to
+    /// false which will disable purging of child sub-orchestration instances.
     /// If <paramref name="instanceId"/> is not found in the data store, or if the instance is found but not in a
     /// terminal state, then the returned <see cref="PurgeResult"/> object will have a
     /// <see cref="PurgeResult.PurgedInstanceCount"/> value of <c>0</c>. Otherwise, the existing data will be purged and
-    /// <see cref="PurgeResult.PurgedInstanceCount"/> will be <c>1</c>.
+    /// <see cref="PurgeResult.PurgedInstanceCount"/> will be <c>countOfInstancesPurged</c>.
```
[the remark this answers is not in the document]
shivamkm07: Done

src/Client/Core/TerminateInstanceOptions.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Client;
+
+/// <summary>
+/// Options to terminate an orchestration.
+/// </summary>
+public class TerminateInstanceOptions
```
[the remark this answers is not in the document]
shivamkm07: Done

src/Client/Core/DurableTaskClient.cs
```diff
@@ -209,12 +209,40 @@ protected DurableTaskClient(string name)
...
+
+    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
+    public virtual Task TerminateInstanceAsync(string instanceId, bool recursive)
+    {
+        TerminateInstanceOptions options = new() { Recursive = recursive };
+        return this.TerminateInstanceAsync(instanceId, options);
+    }
+
+    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
+    public virtual Task TerminateInstanceAsync(string instanceId, bool recursive, CancellationToken cancellation)
```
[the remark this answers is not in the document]
shivamkm07: Removed

src/Client/Core/DurableTaskClient.cs
```diff
@@ -209,12 +209,40 @@ protected DurableTaskClient(string name)
     public abstract Task<OrchestrationMetadata> WaitForInstanceCompletionAsync(
         string instanceId, bool getInputsAndOutputs = false, CancellationToken cancellation = default);
 
-    /// <inheritdoc cref="TerminateInstanceAsync(string, object, CancellationToken)"/>
+    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
     public virtual Task TerminateInstanceAsync(string instanceId, CancellationToken cancellation)
         => this.TerminateInstanceAsync(instanceId, null, cancellation);
 
+    /// <inheritdoc cref="TerminateInstanceAsync(string, TerminateInstanceOptions, CancellationToken)"/>
+    public virtual Task TerminateInstanceAsync(string instanceId, object? output)
```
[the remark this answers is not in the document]
shivamkm07: Done

src/Client/Grpc/GrpcDurableTaskClient.cs
shivamkm07: Done

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
shivamkm07: Done

src/Client/Core/DurableTaskClient.cs
```diff
@@ -353,22 +391,33 @@ public virtual Task ResumeInstanceAsync(string instanceId, CancellationToken can
     /// <see cref="PurgeResult.PurgedInstanceCount"/> value of <c>1</c> or <c>0</c>, depending on whether the target
     /// instance was successfully purged.
     /// </returns>
-    public abstract Task<PurgeResult> PurgeInstanceAsync(
-        string instanceId, CancellationToken cancellation = default);
+    public virtual Task<PurgeResult> PurgeInstanceAsync(
+        string instanceId, bool recursive = true, CancellationToken cancellation = default)
```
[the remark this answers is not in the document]
shivamkm07: Added

src/Client/Core/DurableTaskClient.cs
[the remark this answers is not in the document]
jviau: Oh I meant the other way around:

`PurgeInstanceFilter : PurgeInstanceOptions`

src/Client/Core/PurgeInstanceOptions.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Client;
+
+/// <summary>
+/// Options to purge an orchestration.
+/// </summary>
+public record PurgeInstanceOptions
```
jviau, line 9: You can simplify a record type:

``` CSharp
public record PurgeInstanceOptions(bool Recursive = false);
```

Also, we need to keep recursive as default `false`. Otherwise, the default is a breaking behavior change from what we have today (which is non-recursive).

src/Client/Core/TerminateInstanceOptions.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Client;
+
+/// <summary>
+/// Options to terminate an orchestration.
+/// </summary>
+public record TerminateInstanceOptions
```
jviau, line 9: Can simplify this type as well:

``` CSharp
public record TerminateInstanceOptions(object? Output = null, bool Recursive = false)
```

Also, we need to keep recursive as default `false`. Otherwise, the default is a breaking behavior change from what we have today (which is non-recursive).

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -49,7 +49,6 @@ internal static HistoryEvent ConvertHistoryEvent(P.HistoryEvent proto)
                         OrchestrationInstance = proto.ExecutionStarted.ParentInstance.OrchestrationInstance.ToCore(),
                         TaskScheduleId = proto.ExecutionStarted.ParentInstance.TaskScheduledId,
                     },
-                    Correlation = proto.ExecutionStarted.CorrelationData,
```
jviau, line 52: Why is this removed? Was it removed from the proto?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -321,18 +325,20 @@ public override AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(Orches
 
     /// <inheritdoc/>
     public override Task<PurgeResult> PurgeInstanceAsync(
-        string instanceId, CancellationToken cancellation = default)
+        string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
     {
+        bool recursive = options?.Recursive ?? true;
```
jviau, line 330: default to `false` to avoid breaking behavior change. Please also address this in all other locations for purge and terminate.

If we want the behavior to be `true` by default we will need to do that in a major version rev. You can add a TODO comment for that in the options class.

jviau (changes requested):
Only a few small comments of recursive default behavior.

src/Client/Core/PurgeInstanceOptions.cs
shivamkm07: Actually default value is intentionally kept as `true` here, but I agree it would be a breaking change. The thing is it is already set as true at other places like `durabletask-python` and even in `Dapr`, so setting it to false here would make it inconsistent. thoughts here @cgillum ?

shivamkm07: Okay, makes sense.

shivamkm07: Done

src/Client/Core/TerminateInstanceOptions.cs
shivamkm07: Done

src/Shared/Grpc/ProtoUtils.cs
shivamkm07: Yes. The build was failing once I updated the proto

src/Client/Grpc/GrpcDurableTaskClient.cs
shivamkm07: Done

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
jviau, line 1: Please add a TODO comment to the purge and terminate functions here to support recursive.

jviau (approved):

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
shivamkm07: Done

PR #273, opened 2024-03-14
Fix TaskFailureDetails.IsCausedBy to support custom exceptions and 3rd party exceptions

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/2697

### Problem

The issue was a misunderstanding of how `Type.GetType()` works. The `ErrorType` property in `TaskFailureDetails` has the type name and the namespace, but not the assembly name. Calling `Type.GetType()` without the assembly name only works for types defined in mscorelib and types defined in the entrypoint assembly. However, it doesn't work with types defined externally, including in the app assembly when running in a host like Azure Functions (or a test host process).

### Solution

The solution is to enumerate the assemblies in the current assembly load context to look for the specified exception types.

### Additionally

- This PR also adds a new overload to `IsCausedBy` which takes a `Type` parameter, as an alternative to a `T` type parameter.
- This PR also enhances the existing tests to cover the previously missed scenarios.

### DIFF
--- a/src/Abstractions/TaskFailureDetails.cs
+++ b/src/Abstractions/TaskFailureDetails.cs
@@ -35,6 +35,7 @@ public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? 
     /// for any reason, this method will return <c>false</c>. Base types are supported.
     /// </remarks>
     /// <typeparam name="T">The type of exception to test against.</typeparam>
+    /// <exception cref="AmbiguousMatchException">If multiple exception types with the same name are found.</exception>
     /// <returns>
     /// <c>true</c> if the <see cref="ErrorType"/> value matches <typeparamref name="T"/>; <c>false</c> otherwise.
     /// </returns>
@@ -50,21 +51,44 @@ public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? 
     /// <inheritdoc cref="IsCausedBy{T}"/>
     public bool IsCausedBy(Type targetBaseExceptionType)
     {
+        Check.NotNull(targetBaseExceptionType);
+
         if (!typeof(Exception).IsAssignableFrom(targetBaseExceptionType))
         {
             throw new ArgumentException($"The type {targetBaseExceptionType} is not an exception type.", nameof(targetBaseExceptionType));
         }
 
+        if (string.IsNullOrEmpty(this.ErrorType))
+        {
+            return false;
+        }
+
         // This check works for .NET exception types defined in System.Core.PrivateLib (aka mscorelib.dll)
         this.loadedExceptionType ??= Type.GetType(this.ErrorType, throwOnError: false);
 
-        // This check works for custom exception types defined in the app's assembly
+        // For exception types defined in the same assembly as the target exception type.
+        this.loadedExceptionType ??= targetBaseExceptionType.Assembly.GetType(this.ErrorType, throwOnError: false);
+
+        // For custom exception types defined in the app's assembly
         this.loadedExceptionType ??= Assembly.GetCallingAssembly().GetType(this.ErrorType);
 
-        // This last check works for exception types defined in any loaded assembly (e.g. NuGet packages, etc.)
-        this.loadedExceptionType ??= AppDomain.CurrentDomain.GetAssemblies()
-            .Select(a => a.GetType(this.ErrorType, throwOnError: false))
-            .FirstOrDefault(t => t is not null);
+        if (this.loadedExceptionType is null)
+        {
+            // This last check works for exception types defined in any loaded assembly (e.g. NuGet packages, etc.).
+            // This is a fallback that should rarely be needed except in obscure cases.
+            List<Type> matchingExceptionTypes = AppDomain.CurrentDomain.GetAssemblies()
+                .Select(a => a.GetType(this.ErrorType, throwOnError: false))
+                .Where(t => t is not null)
+                .ToList();
+            if (matchingExceptionTypes.Count == 1)
+            {
+                this.loadedExceptionType = matchingExceptionTypes[0];
+            }
+            else if (matchingExceptionTypes.Count > 1)
+            {
+                throw new AmbiguousMatchException($"Multiple exception types with the name '{this.ErrorType}' were found.");
+            }
+        }
 
         if (this.loadedExceptionType is null)
         {
--- a/src/Abstractions/TaskFailureDetails.cs
+++ b/src/Abstractions/TaskFailureDetails.cs
@@ -15,7 +15,7 @@ namespace Microsoft.DurableTask;
 /// <param name="InnerFailure">The inner cause of the task failure.</param>
 public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? StackTrace, TaskFailureDetails? InnerFailure)
 {
-    Type? exceptionType;
+    Type? loadedExceptionType;
 
     /// <summary>
     /// Gets a debug-friendly description of the failure information.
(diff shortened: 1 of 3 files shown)


src/Abstractions/TaskFailureDetails.cs
```diff
@@ -35,12 +36,37 @@ public override string ToString()
...
+    /// <exception cref="ArgumentException">If <paramref name="type"/> is not an exception type.</exception>
+    /// <inheritdoc cref="IsCausedBy{T}"/>
+    public bool IsCausedBy(Type type)
+    {
+        if (!typeof(Exception).IsAssignableFrom(type))
+        {
+            throw new ArgumentException($"The type {type} is not an exception type.", nameof(type));
+        }
+
+        // This check works for core .NET exception types
```
davidmrdavid, line 58: ultra nit: I suspect this may be more precise? .NET core / core .NET is overloaded (or at least I switched them in my head)
```suggestion
        // This check works for built-in .NET exception types
```

src/Abstractions/TaskFailureDetails.cs
```diff
@@ -35,12 +36,37 @@ public override string ToString()
...
         this.exceptionType ??= Type.GetType(this.ErrorType, throwOnError: false);
-        return this.exceptionType != null && typeof(T).IsAssignableFrom(this.exceptionType);
+
+        // This check works for custom exception types defined in the app's assembly
+        this.exceptionType ??= Assembly.GetCallingAssembly().GetType(this.ErrorType);
+
+        // This last check works for exception types defined in any loaded assembly (e.g. NuGet packages, etc.)
+        this.exceptionType ??= AppDomain.CurrentDomain.GetAssemblies()
+            .Select(a => a.GetType(this.ErrorType, throwOnError: false))
```
jviau, line 66: How will this behave if two assemblies have an exception with the same full name? Will we get false positives?

I wonder if instead we should be using `Type.AssemblyQualifiedName`, which would include assembly and public key. I think `Type.GetType(..)` with that value will then work.

src/Abstractions/TaskFailureDetails.cs
```diff
@@ -35,12 +36,37 @@ public override string ToString()
...
+            .FirstOrDefault(t => t is not null);
+
+        return this.exceptionType is not null && type.IsAssignableFrom(this.exceptionType);
```
davidmrdavid, line 69: does `this.exceptionType` being not `null` not guarantee, at this point, that it's also assignable?

davidmrdavid (commented):
I feel like I'm seeing some .NET magic in here.

Quick question before approving: seems we have to go through a fair bit of trouble due to not knowing the specific assembly where a type originates from. If we could capture the assembly ID, would that not simplify our logic here?

src/Abstractions/TaskFailureDetails.cs
cgillum: I wanted to avoid "built-in" because the definition of "built-in" is arguably even less precise. The technical explanation is that only exceptions in System.Private.CoreLib are correctly resolved. Exceptions shipped by the .NET team that live in other assemblies, however, are not covered. This is why I chose to use "core" (with a lowercase "c" so as not to confuse with .NET Core....which clearly didn't work 😅 ).

Let me know how strongly you feel about this potential point of clarification.

src/Abstractions/TaskFailureDetails.cs
cgillum: What you're suggesting (or some variant of it) would be ideal, but it's unfortunately a breaking change, which is why I didn't go that route. Yes, there's a possibility of false positives, but I think it's extremely unlikely, to the point where I wouldn't block this PR for that. From my view, no app should ever have the same exception type name + namespace in two different assemblies with the expectation that they be treated differently. In many cases you wouldn't even be able to compile such code.

jviau: `AssemblyQualifiedName` includes versions, which might cause an issue if an app is redeployed with a higher dependency. Local testing shows `Type.GetType` works with just type name + assembly name:

``` CSharp
Type t1 = Type.GetType("System.Text.Json.JsonException"); // t1 is null
Type t2 = Type.GetType("System.Text.Json.JsonException, System.Text.Json"); // t2 is not null
```

src/Abstractions/TaskFailureDetails.cs
cgillum: The double-negatives question might be throwing me off, but I think the answer to your question is that `this.exceptionType` not being null simply means that we successfully loaded the type definition. Once loaded, we still need to check to see if what we loaded is compatible with what the user is trying to check for.

Perhaps renaming these variables would make their purposes more clear?

cgillum:
> If we could capture the assembly ID, would that not simplify our logic here?

Yes, that would simplify the logic. The problem is that we'd have to make changes across at least three different repos to capture this additional information: the `Azure/durabletask` repo, the `microsoft/durabletask-protobuf`, and this `microsoft/durabletask-dotnet` repo. We unfortunately can't just add it to the existing `ErrorType` field because that would be a breaking change, as I noted in my response to Jacob's comment.

I feel like the current solution is a reasonable compromise that we can ship immediately.

src/Abstractions/TaskFailureDetails.cs
jviau: > What you're suggesting (or some variant of it) would be ideal, but it's unfortunately a breaking change, which is why I didn't go that route. Yes, there's a possibility of false positives, but I think it's extremely unlikely, to the point where I wouldn't block this PR for that. From my view, no app should ever have the same exception type name + namespace in two different assemblies with the expectation that they be treated differently. In many cases you wouldn't even be able to compile such code.

Would it be a breaking change? We aren't adding any extra fields, just appending it to `ErrorType` field. But it is still backwards compatible: no assembly being there still works. So both of these will 'work':

`ErrorType == "System.Text.Json.JsonException"` -> `IsCausedBy<T>()` continues to not work.
`ErrorType == "System.Text.Json.JsonException, System.Text.Json"` -> `IsCausedBy<T>()` now works.

cgillum: Part of me also worries a bit about the implications of trying to load a type using a specified assembly name. My biggest fear is that we suddenly expose ourselves to an assembly versioning issue where, for example, we save v1 of an assembly name into the history, but then the app is later updated to use v2 of that assembly. In such cases, I expect we would no longer be able to load that exception type, causing the orchestration to go down an unexpected path resulting in a non-determinism exception (A paranoid part of me also wonders if there could be negative security implications of loading arbitrary assemblies into memory...).

jviau: For your two points:

1. If that type is now gone from the assembly, that assembly has had a breaking change. I don't think we can really prevent anything regarding that. It is not a new or unique issue either - any polymorphic deserialization has the same issue.
2. We won't be loading arbitrary assemblies into memory. This would load only assemblies that are known to the application (I believe it will be limited to assemblies in the deps.json).

src/Abstractions/TaskFailureDetails.cs
davidmrdavid: The new clarified comments work for me, thanks!

src/Abstractions/TaskFailureDetails.cs
davidmrdavid: Makes sense - my misunderstanding was in that I did not fully understand the usage of `IsCausedBy`. I re-read the linked issue and this checks out.

davidmrdavid (approved):
I believe I understand the implementation better now, and while it has trade-offs, I'm convinced it is a sufficient improvement to warrant shipping it and unblocking users as-is.

I believe we have an item scheduled to revise the design of exception de-/serialization in .NET. I suggest that we add this area as part of that discussion and to defer those decisions to then :-) .

Since there's still a relevant discussion going on in the comments though, let's reach a resolution on that discussion first, but I'm good with the current solution.

src/Abstractions/TaskFailureDetails.cs
davidmrdavid: Just out of curiosity - 
@jviau are you saying we'd call "Type.AssemblyQualifiedName" and perform a string manipulation to remove the assembly version and everything that comes after it?

So to go from:

> Type.AssemblyQualifiedName: System.Text.Json.JsonException, System.Text.Json, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51

to:

> "System.Text.Json.JsonException, System.Text.Json"

perhaps with a "split by delimeter" operation? Would be nice if no string manipulation was needed at all.

jviau: I think `$"{t.FullName}, {t.Assembly.GetName().Name}"` will be sufficient.

cgillum:
A fix is needed now, even if only a short term fix. We can have a follow up fix to do the work of plumbing through the assembly name in a non-breaking way through all the layers, but we should not block this PR on that longer term fix.

jviau:
> A fix is needed now, even if only a short term fix. We can have a follow up fix to do the work of plumbing through the assembly name in a non-breaking way through all the layers, but we should not block this PR on that longer term fix.

How is updating `FromException` to go from `ErrorType = type.ToString();` --> `ErrorType = $"{type.FullName}, {type.Assembly.GetName().Name}";` breaking?

cgillum:
> > A fix is needed now, even if only a short term fix. We can have a follow up fix to do the work of plumbing through the assembly name in a non-breaking way through all the layers, but we should not block this PR on that longer term fix.
> 
> How is updating `FromException` to go from `ErrorType = type.ToString();` --> `ErrorType = $"{type.FullName}, {type.Assembly.GetName().Name}";` breaking?

Because users are working around the current buggy behavior by doing this: https://github.com/Azure/azure-functions-durable-extension/issues/2697#issuecomment-1997553578, which will break as soon as we make this change to the format of ErrorType.

cgillum:
Jacob and I discussed this offline further and came up with a better solution that improves robustness and performance without introducing breaking changes. I’ll update the PR accordingly.

jviau (approved):

PR #276, opened 2024-03-25
Re-introduce default cancellationToken for `PurgeInstanceAsync` for backwards compatibility

davidmrdavid:
Follow-up to: https://github.com/microsoft/durabletask-dotnet/pull/262

It would appear that in the refactoring above we accidentally made a breaking change to our `PurgeInstancesAsync` API, dropping the declaration of a `PurgeInstanceAsync(string instanceId, CancellationToken cancellation = default)` method, which is overriden [here](https://github.com/Azure/azure-functions-durable-extension/blob/b227f215a0b0d8ba3daee2cd06f791cf4e259c1e/src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs#L54).

At a glance, I suspect this is the reason for our smoke test failures in the `worker.extensions.durabletask` package: https://github.com/Azure/azure-functions-durable-extension/actions/runs/8428205010/job/23080260061?pr=2772

Failure trace:

> #10 7.776 /root/src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs(54,39): error CS0115: 'FunctionsDurableTaskClient.PurgeInstanceAsync(string, CancellationToken)': no suitable method found to override [/root/src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj]

This PR is a naive attempt at trying to fix them.

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -337,7 +337,7 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     public abstract AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(OrchestrationQuery? filter = null);
 
     /// <inheritdoc cref="PurgeInstanceAsync(string, PurgeInstanceOptions, CancellationToken)"/>
-    public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation)
+    public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation = default)
         => this.PurgeInstanceAsync(instanceId, null, cancellation);
 
     /// <summary>


src/Client/Core/DurableTaskClient.cs
```diff
@@ -337,7 +337,7 @@ public virtual Task ResumeInstanceAsync(string instanceId, CancellationToken can
     public abstract AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(OrchestrationQuery? filter = null);
 
     /// <inheritdoc cref="PurgeInstanceAsync(string, PurgeInstanceOptions, CancellationToken)"/>
-    public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation)
+    public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation = default)
```
cgillum, line 340: From a binary compatibility perspective, this is still a breaking change. We need to introduce a virtual method that doesn't include the `CancellationToken` parameter at all.

```csharp
public virtual Task<PurgeResult> PurgeInstancesAsync(string instanceId, CancellationToken cancellation)
    => this.PurgeInstanceAsync(instanceId, CancellationToken.None);
```

cgillum: Wait, maybe my suggestion isn't the right one. I'd need to double check where the failure is coming from.

davidmrdavid: I'll be honest, I haven't had the chance to validate my diff yet either. Just FYI

davidmrdavid:
Before I forget - let's validate that there weren't other similar breaking changes (in other client methods) that the test didn't catch but still need resolving.

src/Client/Core/DurableTaskClient.cs
jviau, line 340: Adding the default param back will cause ambiguous overload resolution with
``` CSharp
public virtual Task<PurgeResult> PurgeInstanceAsync(
        string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
```

when calling `.PurgeInstanceAsync(string)` (which of the 2 default param methods is it to use?)

We will also need to remove the default param from `options` in that other method.

jviau, line 340: Oh! the error message you see of no overload resolution is because of the typo in #267

davidmrdavid:
Closing this as the expected bug fix was merged.

Issue #279, opened 2024-04-03
Ability to Set Both RetryPolicy? Policy and AsyncRetryHandler? Handler in TaskRetryOptions


tomseida:
I am confused why `HandleAsync` was made obsolete in `RetryPolicy` via the [pull request](https://github.com/microsoft/durabletask-dotnet/commit/78300ab0e7dcd0e92847fa74d7e3f0278521c512) tied to issue #275 without providing a means to have both a  retry policy and retry handler.

Per the current implementation `TaskRetryOptions` one can have either `RetryPolicy? Policy` or `AsyncRetryHandler? Handler`, but not both. If one chooses the former, they can create sleep periods between retries, but will retry all exceptions generated by the activity function. If one chooses the later, they can control which exceptions are retried, but not the sleep period between retries.

The orchestrations I implemented prior to moving to isolated functions often used the retry policy and custom handler in tandem. Below is a simplified Activity function that throws transient and non-transient exceptions. 

```
public class AF_Example(IWidgetRepository widgetRepository, IMyApiClientService myApiClientService)
{
    [Function(nameof(AF_Example))]
    public async Task<string> HandleAsync(
        [ActivityTrigger] Guid widgetUID,
        FunctionContext executionContext,
        CancellationToken cancellationToken)
    {
        // If widget is not found, this activity function will never succeed.
        Widget widget = await widgetRepository.GetAsync(widgetUID, cancellationToken)
            ?? throw new NotFoundException(nameof(Widget), widgetUID.ToString());

        // Register widget via an API service and obtain confirmation ID. This
        // operation may experience transient exceptions (5XX status code).
        string confirmationID = await myApiClientService.RegisterWidgetAsync(widget, cancellationToken);

        return confirmationID;
    }
}
```

The first action in this activity function attempts to get a widget from a database. If the `Widget` is not found, there is no point in the orchestration retrying. I would set `AsyncRetryHandler? Handler` delegate to not retry on `NotFoundException`. My orchestrations would have a `try/catch` block to capture these exceptions for handling and I had confidence that no subsequent activity functions would be executed because of the exception.

[...]

PR #281, opened 2024-04-04
Add ability to set both RetryPolicy and AsyncRetryHandler

tomseida:
Add ability to set both RetryPolicy and AsyncRetryHandler for activity functions and suborchestrations. 

See Issue #279


tomseida:
@microsoft-github-policy-service agree



________________________________

Fazer01:
I'm also awaiting this functionality as we had this in place using the in-process variant. As we use .NET8 and isolated we are unable to use both the policy and retryhandler. It's either the policy or the handler, not both at the moment.

stevie1706:
We are also waiting for this PR. Can it be reviewed and merged?

cgillum:
Hi - thanks for this PR!

Question: do you think the `RetryPolicy` and `AsyncRetryHandler` need to be combined, or is there a simpler solution where we just have a simple `Func<Exception, bool>` delegate (non-async) that can be applied to the `RetryPolicy`?

`AsyncRetryHandler` is really supposed to be used for advanced scheduling purposes and allows you to do everything a `RetryPolicy` can do and more. I'd like to keep these separate, if possible, and was hoping maybe you didn't need the full power of `AsyncRetryHandler` for your use case.

tomseida:
@cgillum, I have been using orchestrations for about 4 years and using RetryPolicy most of that time. A simple `Func<Exception, bool>` delegate applied to the `RetryPolicy` has always sufficed. I would be fine with keep this separate from `AsyncRetryHandler`.

stevie1706:
Hi All,   Are we able to get this reviewed and merged?

cgillum:
I'd like us to go with the simpler approach of adding a `Func<Exception, bool>` property to the `RetryPolicy` class to handle exception filtering. That can be done either in this PR or we can close this one and open a separate one.

tomseida:
@cgillum, below is my understanding of your suggestion. This looks exactly like how retries worked for in-process. I am in full support of this since it simplifies function app upgrades from in-process to isolated. Please confirm that the below modification to `RetryPolicy` meets expectations. I am going to assume it does and alter this PR to make this work.
```
    /// <summary>
    /// Gets or sets a delegate to call on exception to determine if retries should proceed.
    /// </summary>
    public Func<Exception, bool>? Handle { get; set; }

#pragma warning disable SA1623 // Property summary documentation should match accessors
    /// <summary>
    /// This functionality is not implemented. Will be removed in the future. Use RetryPolicy.Handle or TaskOptions.FromRetryHandler instead.
    /// </summary>
    [Obsolete("This functionality is not implemented. Will be removed in the future. Use RetryPolicy.Handle or TaskOptions.FromRetryHandler instead.")]
    public Func<Exception, Task<bool>>? HandleAsync { get; set; }
#pragma warning restore SA1623 // Property summary documentation should match accessors
```

tomseida:
Closing this pull request and replaced with PR #314 .

PR #284, opened 2024-04-08
Non-deterministic System.DateTime usage Roslyn Analyzer

allantargino:
This analyzer produces the diagnostic `DURABLE0001`, is the equivalent of [`DF0101`](https://github.com/Azure/azure-functions-durable-extension/blob/cf9f4c228d9d6dd7b8d2ac3f3abb38415f7d6e11/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/DateTimeAnalyzer.cs#L12) in the in-proc version of the analyzer. It looks for usages of System.DateTime.(Now|UtcNow|Today) in orchestration methods (or their invocations).

It used a stateful analyzer ([example](https://github.com/dotnet/roslyn-sdk/blob/main/samples/CSharp/Analyzers/Analyzers.Implementation/StatefulAnalyzers/CompilationStartedAnalyzerWithCompilationWideAnalysis.cs)) in a base class to gather information about existent orchestration methods and their invocations and, in parallel, the concrete implementation looks for the DateTime usages. Last, after the compilation has ended, it matches if the violations happen in orchestrations and reports them as diagnostics.

I expect to use this abstraction/pattern to look for other usages to implement next rules.

Besides, this PR used the [Release Tracking Analyzers](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md) and included two additional files to help keep tracking of released diagnostics.


allantargino:
hm that's weird - the first commit build failed because it failed to compile and find xUnit references. I expected it to just use the common [Directory.Build.targets](https://github.com/microsoft/durabletask-dotnet/blob/d9438aadc5c82c6dabb66779370d9fef8b694dd5/test/Directory.Build.targets#L16).

Then, I manually added the package reference and the build step worked, but now tests are failing as well 🙃

allantargino:
fixed :)
The reference to xUnit was not being imported because the test project had the suffix `Test` instead of `Tests` 😄 ([this check](https://github.com/microsoft/durabletask-dotnet/blob/d9438aadc5c82c6dabb66779370d9fef8b694dd5/test/Directory.Build.targets#L6) + this [variable](https://github.com/microsoft/durabletask-dotnet/blob/d9438aadc5c82c6dabb66779370d9fef8b694dd5/test/Directory.Build.props#L4) were part of it). 

If necessary, I can take the fix and merge it though another PR. Just please let me know 👍

src/Analyzers/Helpers/KnownTypeSymbols.cs
```diff
@@ -0,0 +1,44 @@
...
+/// Provides a set of well-known types that are used by the analyzers.
+/// Inspired by KnownTypeSymbols class in
+/// <see href="https://github.com/dotnet/runtime/blob/2a846acb1a92e811427babe3ff3f047f98c5df02/src/libraries/System.Text.Json/gen/Helpers/KnownTypeSymbols.cs">System.Text.Json.SourceGeneration</see> source code.
+/// Lazy initialization is used to avoid the the initialization of all types during class construction, since not all symbols are used by all analyzers.
+/// </summary>
+sealed class KnownTypeSymbols(Compilation compilation)
+{
+    readonly Compilation compilation = compilation;
+
+    Cached<INamedTypeSymbol?> orchestrationTriggerAttribute;
```
jviau, line 18: is the `Cached<T>` wrapper necessary? Can this work off of just `INamedTypeSymbol?` as the field directly and testing for `null`?

src/Analyzers/Helpers/RoslynExtensions.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DurableTask.Analyzers.Helpers;
```
jviau, line 8: You don't _have_ to follow folder structure as your namespaces. If you want to put this in `Microsoft.DurableTask.Analyzers` that is fine. Then you don't need to add a `using` statement for it each time.

If you want that is - you can also leave it in the `Helpers` namespace if you prefer.

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.CodeAnalysis;
+using Microsoft.DurableTask.Analyzers.Helpers;
+
+namespace Microsoft.DurableTask.Analyzers.Orchestration;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class DateTimeOrchestrationAnalyzer : OrchestrationAnalyzer
```
jviau, line 14: Will this also flag usage of `ITimeProvider`? Or will that be a separate analyzer?

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public const string DiagnosticId = "DURABLE0001";
+
+    static readonly DiagnosticDescriptor Rule = new(
+        DiagnosticId,
+        "DateTime calls must be deterministic inside an orchestration function",
```
jviau, line 20: Do we have plans to add these strings to a resx?

jviau (commented):
Will this analyzer work for non-function triggers as well? IE:

``` CSharp
public class MyOrchestrator : TaskOrchestrator<Input, Output>
{
    public async Task<Output> RunAsync(TaskOrchestrationContext context, Input input)
    {
        var time = DateTime.UtcNow; // will this be checked?
    }
}

```

allantargino:
As it is it doesn't support orchestrators implemented through `ITaskOrchestrator`. I am going to investigate this scenario a bit right and get back to you.

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
allantargino: I believe it can be a different diagnostic/analyzer. I created an issue so we can track it: https://github.com/microsoft/durabletask-dotnet/issues/285

jviau:
> As it is it doesn't support orchestrators implemented through `ITaskOrchestrator`. I am going to investigate this scenario a bit right and get back to you.

That is fine if it doesn't support it right now - we can review this PR for just the DF scenario. Adding support for the contract in another PR will help keep the PRs small

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
bachuv: I think that we should add them to a resx file like in the In Proc Analyzer - https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask.Analyzers/Resources.resx

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true);
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => [Rule];
+
+    protected override void RegisterAdditionalCompilationStartAction(CompilationStartAnalysisContext context, OrchestrationAnalysisResult orchestrationAnalysisResult)
+    {
+        INamedTypeSymbol systemDateTimeSymbol = context.Compilation.GetSpecialType(SpecialType.System_DateTime);
+
+        ConcurrentBag<(ISymbol symbol, IPropertyReferenceOperation operation)> dateTimeUsage = [];
```
bachuv, line 32: Can we add a comment here describing how `dateTimeUsage` is used?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.DurableTask.Analyzers.Helpers;
+
+namespace Microsoft.DurableTask.Analyzers.Orchestration;
+
+public abstract class OrchestrationAnalyzer : DiagnosticAnalyzer
+{
+    public override void Initialize(AnalysisContext context)
```
bachuv, line 17: Can you add some more comments in this method explaining what each `context.xxx()` line is doing?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,131 @@
...
+            // iterating over multiple syntax references is needed because the same method can be declared in multiple places (e.g. partial classes)
+            IEnumerable<MethodDeclarationSyntax> calleeSyntaxes = calleeMethodSymbol.DeclaringSyntaxReferences.Select(r => r.GetSyntax()).OfType<MethodDeclarationSyntax>();
+            foreach (MethodDeclarationSyntax calleeSyntax in calleeSyntaxes)
+            {
+                FindInvokedMethods(semanticModel, calleeSyntax, calleeMethodSymbol, rootOrchestration, result);
+            }
+        }
+    }
+
+    static bool TryTrackMethod(SemanticModel semanticModel, MethodDeclarationSyntax callerSyntax, IMethodSymbol callerSymbol,
```
bachuv, line 91: Can you add a comment explaining what this method does?

bachuv (commented):
Thanks for working on this analyzer!

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
allantargino: done :)

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: I've added comments to the RegisterCompilationStartAction context usage, especially now that I added more registrations :)

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: In the most recent commit, I cleaned up and removed the method. 
The code is the same, just moved to the caller location. Essentially, it adds the visited method to the dictionary.

src/Analyzers/Helpers/RoslynExtensions.cs
allantargino: I moved to `Microsoft.DurableTask.Analyzers`, much better than before - I really liked the flatten structure. Thanks!

src/Analyzers/Helpers/KnownTypeSymbols.cs
allantargino: not necessary :) I removed it and used as suggested! Thanks

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
allantargino: Moved strings to resources file ✅

allantargino:
Hi @jviau and @bachuv, I just fixed the code based on the comments above. Thanks for the great suggestions. Now the analyzer supports Durable Functions, TaskOrchestrators and Func based orchestrations. Besides, I moved the strings to a resource file.

src/Analyzers/Analyzers.csproj
```diff
@@ -31,6 +31,18 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="Resources.resx">
```
jviau, line 35: Lets use the new resx source-gen package: https://www.nuget.org/packages/Microsoft.CodeAnalysis.ResxSourceGenerator/3.11.0-beta1.24122.1

See this blog for guidance: https://blog.nerdbank.net/2023/04/get-rid-of-your-designer-cs-code-behind-for-resx (you can skip the part where it asks to add .NET8 nuget feed source)

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var operation = (IPropertyReferenceOperation)ctx.Operation;
+            IPropertySymbol property = operation.Property;
+
+            if (!property.ContainingSymbol.Equals(systemDateTimeSymbol, SymbolEqualityComparer.Default))
+            {
+                return;
+            }
+
+            if (property.Name is nameof(DateTime.Now) or nameof(DateTime.UtcNow) or nameof(DateTime.Today))
```
jviau, line 50: love the pattern matching!

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,249 @@
...
+    public override void Initialize(AnalysisContext context)
+    {
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+        context.EnableConcurrentExecution();
+
+        context.RegisterCompilationStartAction(context =>
+        {
+            var knownSymbols = new KnownTypeSymbols(context.Compilation);
+
+            if (knownSymbols.FunctionOrchestrationAttribute == null || knownSymbols.FunctionNameAttribute == null ||
```
jviau, line 24: Can you use pattern matching here as well?

not sure if this syntax is correct:

`if (knownSymbols is { FunctionOrchestrationAttribute is null or FunctionNameAttribute is null ... })`

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,249 @@
...
+                knownSymbols.TaskOrchestratorInterface == null || knownSymbols.TaskOrchestratorBaseClass == null ||
+                knownSymbols.DurableTaskRegistry == null)
+            {
+                // symbols not available in this compilation, skip analysis
+                return;
+            }
+
+            IMethodSymbol runAsyncTaskOrchestratorInterface = knownSymbols.TaskOrchestratorInterface.GetMembers("RunAsync").OfType<IMethodSymbol>().First();
```
jviau, line 32: Would it be safer to use `FirstOrDefault()` and then return if its not present? I am under the impression analyzers should avoid throwing.

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,249 @@
...
+                {
+                    return;
+                }
+
+                if (!methodSymbol.TryGetSingleValueFromAttribute(knownSymbols.FunctionNameAttribute, out string functionName))
+                {
+                    return;
+                }
+
+                var orchestration = new AnalyzedOrchestration(functionName);
```
jviau, line 57: nit: prefer this style for ctors:

`AnalyzedOrchestration orchestration = new(functionName);`

src/Analyzers/RoslynExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+            .SelectMany(p => p.GetAttributes())
+            .Any(a => attributeSymbol.Equals(a.AttributeClass, SymbolEqualityComparer.Default));
+    }
+
+    public static bool ImplementsInterface(this INamedTypeSymbol symbol, ISymbol interfaceSymbol)
+    {
+        return symbol.AllInterfaces.Any(i => interfaceSymbol.Equals(i, SymbolEqualityComparer.Default));
+    }
+
+    public static bool InheritsFromOpenGeneric(this INamedTypeSymbol symbol, ITypeSymbol type)
```
jviau, line 59: Is the "OpenGeneric` naming accurate? The implementation doesn't look like it necessarily has to be an open generic.

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
```diff
@@ -0,0 +1,331 @@
...
+    [Fact]
+    public async Task DurableFunctionOrchestrationInvokingRecursiveMethodsHasSingleDiag()
+    {
+        string code = WrapDurableFunctionOrchestration(@"
+[Function(""Run"")]
+long Run([OrchestrationTrigger] TaskOrchestrationContext context) => RecursiveMethod(0);
+
+long RecursiveMethod(int i){
+    if (i == 10) return 1;
+    DateTime date = {|#0:DateTime.Now|};
```
jviau, line 95: Interesting we will examine called methods. How far do we look? How complex of a call chain can this follow? What happens if it is an interface or abstraction method being called?

jviau (commented):
I am not experienced in writing analyzers so can't comment much on the implementation details. I'll look at it further but probably won't hold up the PR on it (as this will be preview).

I think it may be valuable for us to formally detail how our analyzers will follow method call chains that originate in an orchestration. How will this behave with static, instance, interfaces, abstract, etc methods?

src/Analyzers/RoslynExtensions.cs
allantargino: You are right, I renamed it to: `BaseTypeIsConstructedFrom`. Then its usage becomes:
```cs
if (!classSymbol.BaseTypeIsConstructedFrom(knownSymbols.TaskOrchestratorBaseClass))
{
    return;
}
```

Thoughts?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: done :)

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: I originally used `First()` because there is a single member on that interface (and I assumed we won't break that contract by renaming/removing that method). But I do agree it is safer and will protect new versions of the contract, also preventing the same analyzer to break. I just change it, thanks!

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: I could make it work by:

```cs
if (knownSymbols is
    { FunctionOrchestrationAttribute: null } or { FunctionNameAttribute: null } or
    { TaskOrchestratorInterface: null } or { TaskOrchestratorBaseClass: null } or
    { DurableTaskRegistry: null })
{
    // symbols not available in this compilation, skip analysis
    return;
}
```

Giving the additional `} {` tokens per check, which syntax do you prefer? I am fine with either.

>  I experiment a bit with the syntax and code generated in [SharpLab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABAzAAgwJhwBEcBvAXwFgAoG7PQgYTJpzeIH4cEcBeHAHYBXADYiA3K3ZEuATz6DREqWxV5cGAIxIcASwEAXHAFlNACgCULau1s4AbgEMoOMAoEwA7jkaXJNuzZdADMzMAA6Hl5+YTEcNDRXcPloxTELNUC2DAB2HAAGfyy8PJR8IvYqANVqnAB6OpwYAEchXScRGEM1ei0dfSNjfEtrYqcXNxivHz9M9hCwvQBnMgQQNJFyHGgyWXXYzYza4tyCisDTsvOcKts1BpwAExDgmFhu2t7tPUMTLBHSHM2ONXO5pr4LNdbAs3LoVqQ1hsADQ4PYbchWB4QAQieQhHCOOIABygECJK2cMA2BKMBgAFlSlo4ALZUgy6VlAuynQpcy7lNRVKpAA=).

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
allantargino: > What is the expected behavior of the below code:

The `DateTime.Now` property reference operation would have the following diagnostic:
`"The method 'HelperMethod' uses 'System.Date.Now' that may cause non-deterministic behavior when invoked from orchestration 'MyOrchestrator'"`

Within the diagnostic message, I tried to explicitly state that the helper method is not the problem per say, but instead, **its invocation from an orchestration**. That's why for all types of orchestration I try to locate the function name and add it to the diagnostic message. As soon as you remove the call from the orchestration, the diagnostic will disappear.

> How far do we look? How complex of a call chain can this follow? What happens if it is an interface or abstraction method being called?

I am glad you brought this case up, I haven't thought about it before😁. I just checked and we can follow the call chain of every method that has an implementation that isn't hidden behind an interface, abstract class or a library which we don't have access to the actual syntax nodes. For instance, let's say you have:

```cs
interface IHelper {
    DateTime Method();
}

class Helper : IHelper {
    public DateTime Method() => DateTime.Now;
}

class Orchestrator
{
    // this triggers the diagnostic
    [Function(""Run"")]
    void Run([OrchestrationTrigger] TaskOrchestrationContext context)
    {
        var helper = new Helper();
        _ = helper.Method();
    }

    // this does not trigger the diagnostic
    [Function(""Run"")]
    void Run([OrchestrationTrigger] TaskOrchestrationContext context)
    {
        IHelper helper = new Helper();
        _ = helper.Method();
    }
}
```

> I think it may be valuable for us to formally detail how our analyzers will follow method call chains that originate in an orchestration. How will this behave with static, instance, interfaces, abstract, etc methods? And as we create this spec we should consider if this could impact non-orchestration scenarios.

src/Analyzers/Analyzers.csproj
allantargino: That's so cool!! I didn't know that package/code generator. And the generated code is really nice:
image
Thanks for the great tip 👍

Ah, I ended up using version `3.11.0-beta1.24165.2`, released 1h ago 😄

src/Analyzers/RoslynExtensions.cs
jviau: Sounds good - `InheritsFrom` is also a good name,

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
jviau: >  By formally detail, do you mean writing about that behavior in our future analyzer docs?

Write about it in docs, but also write it for us to follow ourselves as we add new analyzers. The method probing behavior should be consistent between all our analyzers.

> I just checked and we can follow the call chain of every method...

Is there a perf concern here? What happens when there is a recursive method?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
jviau: I am good with either. Another option is to just replace `== null` with `is null` in the existing `||` style.

allantargino: Ok, I will keep at it is for now.

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
allantargino: > Write about it in docs, but also write it for us to follow ourselves as we add new analyzers.

Great, I will make sure to document it before we release it: https://github.com/microsoft/durabletask-dotnet/issues/286

> Is there a perf concern here? What happens when there is a recursive method?

Not actually, there is a guard against it: https://github.com/microsoft/durabletask-dotnet/blob/4c97c965db857128f1e737f6888d7513c60d4e82/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs#L198-L203

allantargino: Performance wise, I am going to benchmark it soon 👍

src/Analyzers/Analyzers.csproj
```diff
@@ -29,6 +29,11 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.ResxSourceGenerator" Version="3.11.0-beta1.24165.2" PrivateAssets="all" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="*.resx" />
```
jviau, line 36: This can either be removed.

Unless you need the format methods, in which case add `EmitFormatMethods="true"`. Also change to `**/*.resx` so it will apply to resx files recursively.

``` csharp
    <EmbeddedResource Update="**/*.resx" EmitFormatMethods="true" />
```

But seeing how you pass to `LocalizableString` type, I don't think you need the format methods.

src/Analyzers/Orchestration/DateTimeOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DurableTask.Analyzers.Orchestration;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class DateTimeOrchestrationAnalyzer : OrchestrationAnalyzer
+{
+    public const string DiagnosticId = "DURABLE0001";
+
+    static readonly LocalizableString title = new LocalizableResourceString(nameof(Resources.DateTimeOrchestrationAnalyzerTitle), Resources.ResourceManager, typeof(Resources));
```
jviau, line 17: Is all lower case the naming scheme we have for static readonly?

src/Analyzers/Analyzers.csproj
jviau, line 1: If you add `<EnableStyleCop>true</EnableStyleCop>` to a property group here, you will get more style analyzers for naming conventions, etc.

Example: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/Abstractions.csproj#L5C90-L6C42

jviau (approved):

src/Analyzers/Analyzers.csproj
allantargino: nice, removed :)

src/Analyzers/Analyzers.csproj
allantargino: Cool - it helps a lot with style conventions, added 👍

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,285 @@
...
+                }
+
+                // try to get the name of the orchestration from the method call, otherwise use the containing type name
+                IArgumentOperation nameArgument = invocation.Arguments.First(a => a.Parameter!.Name == "name");
+                Optional<object?> name = nameArgument.GetConstantValueFromAttribute(ctx.Operation.SemanticModel!, ctx.CancellationToken);
+                string orchestrationName = name.Value?.ToString() ?? methodSymbol.Name;
+
+                AnalyzedOrchestration orchestration = new(orchestrationName);
+
+                SyntaxNode funcRootSyntax = delegateCreationOperation.Syntax;
```
bachuv, line 193: How is `delegateCreationOperation` used? What's an example value for it?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,285 @@
...
+            // allows concrete implementations to register specific actions/analysis and then check if they happen in any of the orchestration methods
+            this.RegisterAdditionalCompilationStartAction(context, result);
+        });
+    }
+
+    /// <summary>
+    /// Register additional actions to be executed after the compilation has started.
+    /// It is expected from a concrete implementation of <see cref="OrchestrationAnalyzer"/> to register a
+    /// <see cref="CompilationStartAnalysisContext.RegisterCompilationEndAction"/>
+    /// and then compare that any discovered violations happened in any of the symbols in <paramref name="orchestrationAnalysisResult"/>.
```
bachuv, line 208:
```suggestion
    /// and then compare that to any discovered violations happened in any of the symbols in <paramref name="orchestrationAnalysisResult"/>.
```

Let me know if you mean something different!

test/Analyzers.Tests/Verifiers/CSharpAnalyzerVerifier.Durable.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing;
+
+namespace Microsoft.DurableTask.Analyzers.Tests.Verifiers;
+
+public static partial class CSharpAnalyzerVerifier<TAnalyzer>
```
bachuv, line 9: Can we add a comment here explaining `CSharpAnalyzerVerifier`?

src/Analyzers/Analyzers.csproj
```diff
@@ -29,6 +30,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.4" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.ResxSourceGenerator" Version="3.11.0-beta1.24165.2" PrivateAssets="all" />
```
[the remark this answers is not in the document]
allantargino: Yes, there is no other stable package: https://www.nuget.org/packages/Microsoft.CodeAnalysis.ResxSourceGenerator.

src/Analyzers/KnownTypeSymbols.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Gets ITaskOrchestrator type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskOrchestratorInterface => this.GetOrResolveFullyQualifiedType("Microsoft.DurableTask.ITaskOrchestrator", ref this.taskOrchestratorInterface);
+
+    /// <summary>
+    /// Gets TaskOrchestrator type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskOrchestratorBaseClass => this.GetOrResolveFullyQualifiedType("Microsoft.DurableTask.TaskOrchestrator`2", ref this.taskOrchestratorBaseClass);
```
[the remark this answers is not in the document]
allantargino: That's how you specify fully qualified names in dotnet: https://learn.microsoft.com/en-us/dotnet/fundamentals/reflection/specifying-fully-qualified-type-names#specify-generic-types

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: The framework allows orchestrators to be created using extensions methods of `DurableTaskRegistry`:
image

So, after checking whether an `AddOrchestratorFunc` function was invoked, the code looks for the parameter named `orchestrator`, which as per screenshot above is either an Action or a Func, in other words, the value is an `IDelegateCreationOperation` here.

src/Analyzers/KnownTypeSymbols.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// <summary>
+    /// Gets DurableTaskRegistry type symbol.
+    /// </summary>
+    public INamedTypeSymbol? DurableTaskRegistry => this.GetOrResolveFullyQualifiedType("Microsoft.DurableTask.DurableTaskRegistry", ref this.durableTaskRegistry);
```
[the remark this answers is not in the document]
allantargino: Please see https://github.com/microsoft/durabletask-dotnet/pull/284#discussion_r1568665423

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
allantargino: For examples, please check those unit tests: https://github.com/allantargino/durabletask-dotnet/blob/917542335988f17a4b035b869e7d86a9708c03e4/test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs#L225-L272

src/Analyzers/Analyzers.csproj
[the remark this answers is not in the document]
bachuv: How come we don't need this package in the In proc analyzer?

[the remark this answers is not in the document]
allantargino: This is a new approach to avoid having generated code being committed to the repository. This was a (really really nice) suggestion/idea from Jacob in this very same PR discussion: https://github.com/microsoft/durabletask-dotnet/pull/284#discussion_r1560063036.

The reason why this wasn't needed it was because it used the classical approach of having the generated code by VS and committing it to git. (instead of generating it at compile time).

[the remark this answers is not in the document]
bachuv: I see, thanks for the explanation!

bachuv (approved):

PR #289, opened 2024-04-18
Fixed issue 288: the filter passed to PurgeAllInstancesAsync is now used, instead of creating a new fitler.

JohanKlijn:
the filter passed to PurgeAllInstancesAsync is now used, instead of creating a new fitler.


jviau (commented):
Thanks for the contribution and catching this issue. Seems like we have a testing gap here.

JohanKlijn:
@jviau: I am new to contribute to this project, so I don't know the full process, but do I need to do something before someone reviews the pull request?

jviau (approved):

PR #290, opened 2024-04-18
Fix inner exception propagation

cgillum:
Part of fixing https://github.com/Azure/azure-functions-durable-extension/issues/2476

There were various places where we weren't correctly setting the inner failure details for `TaskFailureDetails`. This PR fixes those and adds a new test to cover this scenario specifically.

Note that this PR depends on updates to DurableTask.Core (https://github.com/Azure/durabletask/pull/1067) and DurableTask.Sidecar (https://github.com/microsoft/durabletask-sidecar/pull/24). I'm not expecting this PR to build until we have the new versions of those packages available.

### DIFF
--- a/src/Abstractions/TaskFailureDetails.cs
+++ b/src/Abstractions/TaskFailureDetails.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using CoreFailureDetails = DurableTask.Core.FailureDetails;
 using CoreOrchestrationException = DurableTask.Core.Exceptions.OrchestrationException;
 
 namespace Microsoft.DurableTask;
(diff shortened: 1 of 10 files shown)


src/Abstractions/TaskFailureDetails.cs
```diff
@@ -107,16 +109,78 @@ public bool IsCausedBy(Type targetBaseExceptionType)
...
                 coreEx.FailureDetails?.ErrorMessage ?? "(unknown)",
                 coreEx.FailureDetails?.StackTrace,
-                null /* InnerFailure */);
+                FromCoreFailureDetailsRecursive(coreEx.FailureDetails?.InnerFailure) ?? FromExceptionRecursive(coreEx.InnerException));
+        }
+
+        return new TaskFailureDetails(
+            exception.GetType().ToString(),
+            exception.Message,
+            exception.StackTrace,
```
jviau, line 135: I am a bit concerned that stack trace will bloat the size of this payload. Should we consider this now or something to think about later?

jviau (approved):

src/Abstractions/TaskFailureDetails.cs
cgillum: Would it make sense to have some kind of configuration flag to exclude the stack trace?

jviau: Probably yeah. Or we could opt to not include stack trace at all for inner exceptions?

davidmrdavid:
Should this PR be marked as "ready for review" now that the other PR is merged?

src/Abstractions/TaskFailureDetails.cs
davidmrdavid: I think there's an argument to be made that, just like we claim inputs and outputs should be small (and we don't tamper with that payload on the user's behalf) that if there's a very large exception trace then it's on the user to trim it / be aware of the perf cost.

I worry that by trying to optimize this case, we'll just get more customer questions about exception serialization is "incomplete". My recommendation would be to include the full stack trace by default, and to possibly add the feature flag to disable it if we find that, in practice, this is a problem.

davidmrdavid (approved):

src/Abstractions/TaskFailureDetails.cs
cgillum: Let's go with what we have for now, but happy to follow up in a separate set of PRs.

PR #295, opened 2024-04-22
Add versioning support in durablttask-dotnet(phase1)

skyao:
Code updates:

- TaskName

  Add a new constructor that sets the version expected by the client.

  The existing code has a version field and associated implementation, but no constructor to set the version is provided. Adding this constructor makes it possible to do so in client-side code:

  ```c#
  string instanceId = await client.ScheduleNewOrchestrationInstanceAsync(
    new TaskName(nameof(HelloOrchestration), "1.2.0"));
  ```
    

- TaskOrchestrationContext

  Add new abstract method to get instance version.

  So that in the orchestration code, we get the instance version by this context:

  ```c#
      public static async Task<List<string>> RunOrchestrator(
          [OrchestrationTrigger] TaskOrchestrationContext context)
      {
  			string instanceVersion = context.InstanceVersion;
        ......
      }
  ```

- TaskOrchestrationContextWrapper

  Implemented new added abstract method to get instance version.

  This method will get the value of instance version from OrchestrationContext which is updated to add InstanceVersion filed and pass the value in another PR in durable repo.

### DIFF
--- a/src/Abstractions/TaskName.cs
+++ b/src/Abstractions/TaskName.cs
@@ -30,6 +30,26 @@ public readonly struct TaskName : IEquatable<TaskName>
     }
 
     /// <summary>
+    /// Initializes a new instance of the <see cref="TaskName"/> struct.
+    /// </summary>
+    /// <param name="name">The name of the task. Providing <c>null</c> will yield the default struct.</param>
+    /// <param name="version">The version of the task.</param>
+    public TaskName(string name, string version)
+    {
+        if (name is null)
+        {
+            // Force the default struct when null is passed in.
+            this.Name = null!;
+            this.Version = version!;
+        }
+        else
+        {
+            this.Name = name;
+            this.Version = version;
+        }
+    }
+
+    /// <summary>
     /// Gets the name of the task without the version.
     /// </summary>
     /// <value>
--- a/src/Abstractions/TaskOrchestrationContext.cs
+++ b/src/Abstractions/TaskOrchestrationContext.cs
@@ -23,6 +23,11 @@ public abstract class TaskOrchestrationContext
     public abstract string InstanceId { get; }
 
     /// <summary>
+    /// Gets the version of the current orchestration instance.
+    /// </summary>
+    public abstract string InstanceVersion { get; }
+
+    /// <summary>
     /// Gets the parent instance or <c>null</c> if there is no parent orchestration.
     /// </summary>
     public abstract ParentOrchestrationInstance? Parent { get; }
--- a/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
@@ -52,6 +52,9 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
     public override string InstanceId => this.innerContext.OrchestrationInstance.InstanceId;
 
     /// <inheritdoc/>
+    public override string InstanceVersion => this.innerContext.OrchestrationInstance.InstanceVersion;
+
+    /// <inheritdoc/>
     public override ParentOrchestrationInstance? Parent => this.invocationContext.Parent;
 
     /// <inheritdoc/>


skyao:
This PR is depended on the PR of https://github.com/Azure/durabletask/pull/1071   to build.

src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -22,6 +22,11 @@ public abstract class TaskOrchestrationContext
     /// </summary>
     public abstract string InstanceId { get; }
 
+    /// <summary>
+    /// Gets the version of the current orchestration instance.
+    /// </summary>
+    public abstract string InstanceVersion { get; }
```
jviau, line 28: Adding a new abstract member is a source breaking change. May need to be `virtual` and return `null` by default.

jviau (changes requested):

src/Abstractions/TaskName.cs
```diff
@@ -29,6 +29,26 @@ public TaskName(string name)
...
+    /// </summary>
+    /// <param name="name">The name of the task. Providing <c>null</c> will yield the default struct.</param>
+    /// <param name="version">The version of the task.</param>
+    public TaskName(string name, string version)
+    {
+        if (name is null)
+        {
+            // Force the default struct when null is passed in.
+            this.Name = null!;
+            this.Version = version!;
```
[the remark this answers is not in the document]
skyao: Thank you, @jviau, I follow your above two suggestions, but there is warning:

```c#
        if (name is null && version is null)
        {
            // Force the default struct when null is passed in.
            this.Name = null!;
            this.Version = null!;
        }
        else
        {
            this.Name = name;    // 'name' may be null here.
            this.Version = version;
        }
```

src/Abstractions/TaskOrchestrationContext.cs
skyao: Good suggestion, code updated, please help to review and confirm it.

And there is a similar problem in another PR but a little more complex, please have a look at: 

https://github.com/Azure/azure-functions-durable-extension/pull/2799#discussion_r1576627815

src/Abstractions/TaskName.cs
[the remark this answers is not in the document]
jviau: hmm yeah need to think through the behavior a bit: what to do when `name` is `null`, but `version` isn't? That is probably an error case. Maybe `ArgumentException` explaining `name` must not be `null` when `version` is non-null.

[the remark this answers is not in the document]
skyao: There are four case about name or version is null or not:

1. name is null & version is null

    ```c#
    this.Name = null!;
    this.Version = null!;
    ```
2. name is null & version is not null

    ```c#
    throw ArgumentException("name must not be null when version is non-null")
    ```

3. name is not null and version is null

    ```c#
    this.Name = name;
    this.Version = null!;
    ```

4. name is not null and version is null

    ```c#
    this.Name = name;
    this.Version =version;
    ```

Please confirm if above are right.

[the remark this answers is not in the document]
jviau: Almost correct - just some nuance around `Version`. It is not null, so we should default to `string.Empty` (as we do currently).

- If version is null and name is not null, throw argument exception as well. (Users should pass in string.Empty as default).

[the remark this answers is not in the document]
skyao: Thank, I had updated the code, please help to review. 

For the case that version is null and name is not null, I suggest not throw argument exception but fallback to the constructor without version parameter. 

I think it would be more friendly. Of cause, it depends on how we co "version is null and name is not null",  This shouldn't be considered an error. 

Of course, it depends on how we look at the case "version is null and name is not null", whether we think of version as an required parameter and should not be null, or whether we understand version as a non-required parameter that null value is acceptable and we will fallback to the constructor without version parameter.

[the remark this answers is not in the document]
skyao: @jviau @cgillum please help to review it and let us make a decision, this PR is almost ready to merge except this.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,162 @@ namespace Microsoft.DurableTask;
...
+
+  /// <summary>
+  /// Gets a value indicating whether to <see cref="TaskName"/> objects
+  /// are equal using value semantics.
+  /// </summary>
+  /// <param name="other">The other object to compare to.</param>
+  /// <returns><c>true</c> if the two objects are equal using value semantics; otherwise <c>false</c>.</returns>
+  public bool Equals(TaskName other)
+  {
+    return string.Equals(this.Name, other.Name, StringComparison.OrdinalIgnoreCase);
```
jviau, line 126: We will need to update this to include `Version` in the comparison.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,162 @@ namespace Microsoft.DurableTask;
...
 
-        return this.Equals(other);
-    }
+  /// <summary>
+  /// Calculates a hash code value for the current <see cref="TaskName"/> instance.
+  /// </summary>
+  /// <returns>A 32-bit hash code value.</returns>
+  public override int GetHashCode()
+  {
+    return StringComparer.OrdinalIgnoreCase.GetHashCode(this.Name);
```
jviau, line 151: We will need to update this to include `Version` in the hash code.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,162 @@ namespace Microsoft.DurableTask;
...
     {
-        if (string.IsNullOrEmpty(this.Version))
-        {
-            return this.Name;
-        }
-        else
-        {
-            return this.Name + ":" + this.Version;
-        }
+      return this.Name + ":" + this.Version;
```
jviau, line 166: We should consider adding FromString semantics. But I wonder if that will be a breaking change? We have not restricted what characters can be used in a task name before this.

This change is going to introduce a confusing behavior:

``` CSharp
TaskName name1 = new("MyTask", "2"); // Name = "MyTask", Version = "2'
TaskName name2 = name1.ToString(); // implicit operator convers. Name = "MyTask:2", Version = ""
```

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,162 @@ namespace Microsoft.DurableTask;
...
-    /// <summary>
-    /// Implicitly converts a <see cref="string"/> into a <see cref="TaskName"/> value.
-    /// </summary>
-    /// <param name="value">The string to convert into a <see cref="TaskName"/>.</param>
-    public static implicit operator TaskName(string value) => string.IsNullOrEmpty(value) ? default : new(value);
+  /// <summary>
+  /// Implicitly converts a <see cref="TaskName"/> into a <see cref="string"/> of the <see cref="Name"/> property value.
+  /// </summary>
+  /// <param name="value">The <see cref="TaskName"/> to be converted into a string.</param>
+  public static implicit operator string(TaskName value) => value.Name;
```
jviau, line 88: This doesn't call `value.ToString()`? (Shocked I missed that previously). Should definitely fix that.

jviau (changes requested):
I am a bit concerned about the usability of this API. Both behavior we had before this PR, but also behavior afterwards.

Right now my biggest concern is we have a semantic of using `:` as the separator for name & version when going to string. We never restricted characters in the task name before, but we have a need for it now (restricting `:`) which is a breaking change. Wonder how we can work through this.

src/Abstractions/TaskName.cs
skyao: Updated as you suggestion.

src/Abstractions/TaskName.cs
skyao: Updated as your suggestion.

src/Abstractions/TaskName.cs
skyao: Updated as your suggestion.

skyao:
In this PR I just want to pass version field correctly so that it can used to support versioning feature. But I don't know if I should update the related method like string() / Equals() / GetHashCode() because before this PR these methods ignored version field. 

I'm afraid that after I updated these methods (string() / Equals() / GetHashCode()) to include version field but not only name field, there are some broken changes to exist code. 

@jviau @cgillum please help to make a decision that should I update these methods to include version field.

src/Abstractions/TaskName.cs
```diff
@@ -148,7 +149,13 @@ public override bool Equals(object? obj)
   /// <returns>A 32-bit hash code value.</returns>
   public override int GetHashCode()
   {
-    return StringComparer.OrdinalIgnoreCase.GetHashCode(this.Name);
+    if (string.IsNullOrEmpty(this.Version))
+    {
+      return StringComparer.OrdinalIgnoreCase.GetHashCode(this.Name);
+    }
+
+    return StringComparer.OrdinalIgnoreCase.GetHashCode(this.Name) * 31
```
jviau, line 157: I think you can add a reference to [Microsoft.Bcl.HashCode](https://www.nuget.org/packages/Microsoft.Bcl.HashCode). You can then use `HashCode.Combine(this.Name, this.Version);`

src/Abstractions/TaskName.cs
jviau, line 1: Can you change this file back to 4 space indentation

jviau:
> In this PR I just want to pass version field correctly so that it can used to support versioning feature. But I don't know if I should update the related method like string() / Equals() / GetHashCode() because before this PR these methods ignored version field.
> 
> I'm afraid that after I updated these methods (string() / Equals() / GetHashCode()) to include version field but not only name field, there are some broken changes to exist code.
> 
> @jviau @cgillum please help to make a decision that should I update these methods to include version field.

The from string conversion is the only breaking change I see. We need to parse the version out of the string, otherwise round-tripping between string and back will give different results. But if we choose a delimiter a customer already uses, it can break their orchestrations. But I am not sure what other choice we have if we want to introduce `Version`.

@cgillum what do you think? Do we introduce a behavior breaking change?

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,169 @@ namespace Microsoft.DurableTask;
...
+  /// The name of the activity task without the version.
+  /// </value>
+  public string Name { get; }
+
+  /// <summary>
+  /// Gets the version of the task.
+  /// </summary>
+  /// <remarks>
+  /// Task versions is currently locked to <see cref="string.Empty" /> as it is not yet integrated into task
+  /// identification. This is being left here as we intend to support it soon.
```
cgillum, line 80: Do we need to update these remaks?

src/Abstractions/TaskName.cs
cgillum: Should the implicit operator be updated to parse out the version?

I'm not super concerned about the breaking change since 99+% of .NET users should be relying on the function name, which already can't have special characters. However, we can take a look at Kusto data to confirm.

cgillum (commented):
Added some comments. One thing that's odd to me is that we don't roundtrip the version information when converting between `TaskName` and `string`. I left a comment about this.

The breaking change part might be something we can confirm with telemetry. I can help with this.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,169 @@ namespace Microsoft.DurableTask;
...
-    /// </summary>
-    /// <param name="a">The first <see cref="TaskName"/> to compare.</param>
-    /// <param name="b">The second <see cref="TaskName"/> to compare.</param>
-    /// <returns><c>true</c> if the two <see cref="TaskName"/> objects are not equal; otherwise <c>false</c>.</returns>
-    public static bool operator !=(TaskName a, TaskName b)
+  /// <summary>
+  /// Initializes a new instance of the <see cref="TaskName"/> struct.
+  /// </summary>
+  /// <param name="name">The name of the task. Providing <c>null</c> will yield the default struct.</param>
+  /// <param name="version">The version of the task.</param>
```
cgillum, line 36: Should we enforce a convention for the version? For example, it must either by an `Int32` or something that can be parsed by `System.Version.TryParse`? This will make it easier for us to implement automatic version selection on the client side, as well as make it easier for customers to do version comparisons in their code.

jviau:
> One thing that's odd to me is that we don't roundtrip the version information when converting between `TaskName` and `string`.

Yes, we definitely need to add this. @skyao can you also add `public static TaskName FromString(string taskName)` and have the string conversion operator call this. It should parse out name and version based on the convention we settle on.

src/Abstractions/TaskName.cs
jviau: I had thought about this before and I settled on not enforcing a convention, as I could see customers wanting a wide variety of versioning schemes. For example, Azure resources often version with a date format for their public API and they might want to use the same versioning scheme for durable.

src/Abstractions/TaskName.cs
jviau: Yes the implicit operator would be changed to call `FromString`.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,169 @@ namespace Microsoft.DurableTask;
...
     {
-        if (string.IsNullOrEmpty(this.Version))
-        {
-            return this.Name;
-        }
-        else
-        {
-            return this.Name + ":" + this.Version;
-        }
+      return this.Name + ":" + this.Version;
```
jviau, line 173: We should have a discussion on what separator to use. I like `@`, as that is what we have used for distributed tracing to indicate the version - and ":" was used to separate the task type: `{task_type}:{name}@{version}`

However, we already use `@` to separate entity name and key - will that be an issue? Or is that a good thing to use `@` as a separator in both for consistency.

src/Abstractions/TaskName.cs
skyao: I referenced the code in src\Shared\Core\CSharpLangCompat.cs 

```c#
        public override int GetHashCode()
        {
            return (this.Start.GetHashCode() * 31) + this.End.GetHashCode();
        }
```

I agree with you that [Microsoft.Bcl.HashCode](https://www.nuget.org/packages/Microsoft.Bcl.HashCode) is a good  choose, but it means that we need to introduce a new dependency. 

If you and Chris agree that introducing new dependency in this PR is acceptable, I would like to update the code.

skyao: Done. Added dependency and update code as your suggestion.

src/Abstractions/TaskName.cs
```diff
@@ -8,127 +8,169 @@ namespace Microsoft.DurableTask;
...
-    /// <param name="other">The other object to compare to.</param>
-    /// <returns><c>true</c> if the two objects are equal using value semantics; otherwise <c>false</c>.</returns>
-    public bool Equals(TaskName other)
+    else
     {
-        return string.Equals(this.Name, other.Name, StringComparison.OrdinalIgnoreCase);
+      if (version is null)
+      {
+        this.Name = name;
+        this.Version = string.Empty; // fallback to the contructor without version parameter
```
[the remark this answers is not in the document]
skyao: Typo fixed

src/Abstractions/TaskName.cs
cgillum: I think using `@` could break entity ID parsing. Looking at [this code](https://github.com/Azure/durabletask/blob/d4b98d46158e470c6b769601cd76a86d9bb6d096/src/DurableTask.Core/Entities/EntityId.cs#L54-L78), I can see that entity IDs take the form of `{name}@{key}` and the code that parses this looks for the first instance of `@` to separate the name from the key. If we inject a version into the name, we make it `{name}@{version}@{key}`, and the key for entities will be parsed as `{version}@{key}` instead of `{key}`.

jviau: @cgillum entity Ids is the _instance_ ID - this is the _name_. They are not the same field.

src/Abstractions/TaskName.cs
skyao: Done.

skyao:
> > One thing that's odd to me is that we don't roundtrip the version information when converting between `TaskName` and `string`.
> 
> Yes, we definitely need to add this. @skyao can you also add `public static TaskName FromString(string taskName)` and have the string conversion operator call this. It should parse out name and version based on the convention we settle on.

Done, code updated. Please have a look.

src/Abstractions/TaskName.cs
jviau: @cgillum yeah I am trying to think through how that behaves. One thing is that it is a separate type for entities. We have a few options:

1. use `@` for separator. Entity instance IDs can then look like `@{entity}@{version}@{key}` (or we can do key first, then version). **RISK**: keys could already contain `@`. This could break those instances.
    - @sebastianburckhardt suggested we could prefix with double `@@` to indicate this entity has a version field. So `@myEntity@some@key` -> no version field. `@@myEntity@version@some@key` -> version field present.
3. use a different character. `@{entity}?{version}@{key}` (or key first). **RISK**: this character could already be used by customers in name _or_ key
4. Alternatively, we could opt to _not_ support versioning of entities just yet

jviau: Want to clarify that I am not opposed to a different separator char for version. We haven't officially GA'd distributed tracing v2, so we could update that spec with the new separator. But what one exactly?

Some that I have ruled out myself in the past:

- `/` - avoided this so it wasn't confused with HTTP routes in distributed tracing
- `%` - often used for encoding special chars, so wanted to avoid that
- `*` - wild card, wanted to avoid that.
- `!` - used in netherite for partition targeting. We are looking to also support that in other backends.
- `:` - used in distributed tracing to separate the task-type. ie: `orchestration:name@version` or `activity:name@version`

skyao: Can we merge this PR first and then open a new issue to trace and update the separator after we have a final decision? 

Because today is my last day, I'm afraid that I can't continue to update the code in this PR because of permissions.

skyao: @cgillum @jviau please have a look at this PR and let's make the final decision.

YunchuWang:
versioning support is released, do we still keep the pr? @skyao

cgillum:
Sky has moved on to a new role so we can go ahead and close this.

PR #296, opened 2024-04-22
Guid.NewGuid Roslyn Analyzer

allantargino:
This analyzer produces the diagnostic DURABLE0002, is the equivalent of [DF0102](https://github.com/Azure/azure-functions-durable-extension/blob/cf9f4c228d9d6dd7b8d2ac3f3abb38415f7d6e11/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/GuidAnalyzer.cs#L14) in the in-proc version of the analyzer. It looks for usages of `Guid.NewGuid()` in orchestrations (and their method invocations).


bachuv (approved):
LGTM!

PR #300, opened 2024-04-26
using a visitor to decouple the orchestrations discovery from analysis

allantargino:
I realized that `OrchestrationAnalyzer` base class was performing two responsibilities:

1. Discovering orchestrations
2. Recursively finding methods invocations

While this was ok for the first 3 concrete implementations (DateTime, Guid and Delay analyzers), I couldn't properly use the base class to inspect discovered orchestrations and inspect their method signatures (when looking for Cancellation Tokens, checking bindings, etc).
The discovery behavior was tightly coupled to the recursive method invocation finding. 

In order to decouple it, I propose we introduce `OrchestrationVisitor`, a class that will contain the logic to be executed after an orchestration is found. `MethodProbeOrchestrationVisitor` represents the existent recursive method invocation behavior it had before.

This abstraction greatly simplified the code, allowing us to even remove the usage of concurrent collections and compilation end actions.

### DIFF
--- a/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
+++ b/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
@@ -205,7 +205,11 @@ public abstract class OrchestrationAnalyzer<TOrchestrationVisitor> : DiagnosticA
 }
 
 /// <summary>
-/// Base class for visitors that analyze orchestrations.
+/// An Orchestration Visitor allows a concrete implementation of an analyzer to visit different types of orchestrations,
+/// such as Durable Functions, TaskOrchestrator, ITaskOrchestrator, and OrchestratorFunc.
+/// It provides a set of methods that can be overridden to inspect different types of orchestrations.
+/// Besides, it provides a method to initialize the visitor members, checking for available symbols and
+/// return whether the concrete implementation visitor should continue running and perform its analysis.
 /// </summary>
 public abstract class OrchestrationVisitor
 {
--- a/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
+++ b/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
@@ -214,7 +214,7 @@ public abstract class OrchestrationAnalyzer<TOrchestrationVisitor> : DiagnosticA
 public abstract class OrchestrationVisitor
 {
     /// <summary>
-    /// Gets the compilation adquired from analyzer context.
+    /// Gets the Compilation instance acquired from the analyzer context, including syntax trees, semantic models, and other information.
     /// </summary>
     protected Compilation Compilation { get; private set; } = null!;
 
--- a/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
+++ b/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
@@ -11,90 +11,6 @@ using Microsoft.CodeAnalysis.Operations;
 namespace Microsoft.DurableTask.Analyzers.Orchestration;
 
 /// <summary>
-/// Base class for visitors that analyze orchestrations.
-/// </summary>
-public abstract class OrchestrationVisitor
-{
-    /// <summary>
-    /// Gets the compilation adquired from analyzer context.
-    /// </summary>
-    protected Compilation Compilation { get; private set; } = null!;
-
-    /// <summary>
-    /// Gets the set of well-known type symbols.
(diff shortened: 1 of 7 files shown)


src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
 
 namespace Microsoft.DurableTask.Analyzers.Orchestration;
 
+/// <summary>
+/// Base class for visitors that analyze orchestrations.
+/// </summary>
+public abstract class OrchestrationVisitor
```
allantargino, line 16: For reference, using this abstraction, here's how easy it is to build new analyzers that don't care about method probing, such as the ones only interested in inspecting method's signatures: https://github.com/allantargino/durabletask-dotnet/commit/162a5e04dc73bba678ed11121244279d90383274

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
...
+{
+    /// <summary>
+    /// Gets the compilation adquired from analyzer context.
```
bachuv, line 19:
```suggestion
    /// Gets the compilation acquired from the analyzer context to compile a set of syntax trees.
```

Just a suggestion!

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
...
+
+    /// <summary>
+    /// Gets the set of well-known type symbols.
+    /// </summary>
+    protected KnownTypeSymbols KnownTypeSymbols { get; private set; } = null!;
+
+    /// <summary>
+    /// Initializes the visitor members and returns whether the concrete implementation visitor was initialized.
+    /// </summary>
+    /// <param name="compilation">The compilation adquired from analyzer context.</param>
```
bachuv, line 31:
```suggestion
    /// <param name="compilation">The compilation acquired from analyzer context.</param>
```

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
...
+    {
+    }
+}
+
 /// <summary>
 /// Base class for analyzers that analyze orchestrations.
 /// </summary>
-public abstract class OrchestrationAnalyzer : DiagnosticAnalyzer
+/// <typeparam name="TOrchestrationVisitor">Orchestration Visitor to be used during Orchestrations discovery.</typeparam>
+public abstract class OrchestrationAnalyzer<TOrchestrationVisitor> : DiagnosticAnalyzer where TOrchestrationVisitor : OrchestrationVisitor, new()
```
bachuv, line 101: Can we move this class to be first in this file? I think it makes more sense since the file name is `OrchestrationAnalyzer`.

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
...
+    /// <param name="methodSyntax">Method Syntax Node.</param>
+    /// <param name="methodSymbol">Method Symbol.</param>
+    /// <param name="orchestrationName">Durable Function name.</param>
+    /// <param name="reportDiagnostic">Function that can be used to report diagnostics.</param>
+    public virtual void VisitDurableFunction(SemanticModel semanticModel, MethodDeclarationSyntax methodSyntax, IMethodSymbol methodSymbol, string orchestrationName, Action<Diagnostic> reportDiagnostic)
+    {
+    }
+
+    /// <summary>
+    /// Visits a TaskOrchestrator&lt;T1,T2&gt; orchestration.
```
bachuv, line 61: Can we fix this comment?

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -10,10 +10,95 @@
 
 namespace Microsoft.DurableTask.Analyzers.Orchestration;
 
+/// <summary>
+/// Base class for visitors that analyze orchestrations.
```
bachuv, line 14: Can you add a longer description in this summary of how the `OrchestrationVisitor` class is used. A high level design description would be helpful for maintainability in the future. Thanks!

bachuv (approved):
LGTM!

PR #301, opened 2024-05-02
Attribute Binding Roslyn Analyzers

allantargino:
This PR adds 3 Roslyn analyzers related to attribute binding:
- OrchestrationTriggerBindingAnalyzer: Analyzer that reports when 'OrchestrationTriggerAttribute' is not used with a 'TaskOrchestrationContext'.
- DurableClientBindingAnalyzer: Analyzer that reports when 'DurableClientAttribute' is not used with a 'DurableTaskClient'.
- EntityTriggerBindingAnalyzer: Analyzer that reports when 'EntityTriggerAttribute' is not used with a 'TaskEntityDispatcher'.

### DIFF
--- a/src/Analyzers/AttributeBinding/DurableClientBindingAnalyzer.cs
+++ b/src/Analyzers/AttributeBinding/DurableClientBindingAnalyzer.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.DurableTask.Analyzers.AttributeBinding;
+
+/// <summary>
+/// Analyzer that matches 'DurableClientAttribute' with 'DurableTaskClient' parameters.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class DurableClientBindingAnalyzer : MatchingAttributeBindingAnalyzer
+{
+    /// <summary>
+    /// Diagnostic ID supported for the analyzer.
+    /// </summary>
+    public const string DiagnosticId = "DURABLE1002";
+
+    static readonly LocalizableString Title = new LocalizableResourceString(nameof(Resources.DurableClientBindingAnalyzerTitle), Resources.ResourceManager, typeof(Resources));
+    static readonly LocalizableString MessageFormat = new LocalizableResourceString(nameof(Resources.DurableClientBindingAnalyzerMessageFormat), Resources.ResourceManager, typeof(Resources));
+
+    static readonly DiagnosticDescriptor Rule = new(
+            DiagnosticId,
+            Title,
+            MessageFormat,
+            AnalyzersCategories.AttributeBinding,
+            DiagnosticSeverity.Error,
+            isEnabledByDefault: true);
+
+    /// <inheritdoc/>
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => [Rule];
+
+    /// <inheritdoc/>
+    protected override ExpectedBinding GetExpectedBinding(KnownTypeSymbols knownTypeSymbols)
+    {
+        return new ExpectedBinding()
+        {
+            Attribute = knownTypeSymbols.DurableClientAttribute,
+            Type = knownTypeSymbols.DurableTaskClient,
+        };
+    }
+
+    /// <inheritdoc/>
+    protected override void ReportDiagnostic(SymbolAnalysisContext ctx, ExpectedBinding expected, IParameterSymbol parameter)
+    {
+        string wrongType = parameter.Type.ToDisplayString(SymbolDisplayFormat.CSharpShortErrorMessageFormat);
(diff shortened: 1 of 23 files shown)


src/Analyzers/KnownTypeSymbols.cs
```diff
@@ -50,6 +55,31 @@ public sealed class KnownTypeSymbols(Compilation compilation)
...
 
+    /// <summary>
+    /// Gets a TaskOrchestrationContext type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskOrchestrationContext => this.GetOrResolveFullyQualifiedType("Microsoft.DurableTask.TaskOrchestrationContext", ref this.taskOrchestrationContext);
+
+    /// <summary>
+    /// Gets a DurableClientAttribute type symbol.
+    /// </summary>
+    public INamedTypeSymbol? DurableClientAttribute => this.GetOrResolveFullyQualifiedType("Microsoft.Azure.Functions.Worker.DurableClientAttribute", ref this.durableClientAttribute);
```
jviau, line 66: Might be worth splitting functions specific types into a sub class `KnownTypeSymbols.Functions.{Symbol}`. Can also then organize into partial-files.

Are there other groupings of known types we can split into sub classes?

src/Analyzers/AttributeBinding/MatchingAttributeBindingAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// <inheritdoc/>
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.None);
+
+        context.RegisterCompilationStartAction(
+            ctx =>
+            {
+                KnownTypeSymbols knownTypeSymbols = new(ctx.Compilation);
```
jviau, line 40: I am not super familiar with source generators APIs, but is there a way to share state between our analyzers? Can then init `KnownTypeSymbols` once and re-use it.

Not high priority, just curious if it is possible.

jviau (approved):

jviau:
Can we also organize all functions-specific analyzers under a `Functions` folder?

src/Analyzers/AttributeBinding/MatchingAttributeBindingAnalyzer.cs
allantargino: I was reading https://github.com/dotnet/roslyn/issues/6324 and seems there is a way by using the `TryGetValue` methods, available throughout the different analysis contexts: https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostics.compilationstartanalysiscontext.trygetvalue?view=roslyn-dotnet-4.7.0. I am going to create an issue to investigate it later, that's a good suggestion, although I would love to measure whether the performance impact is worth the state sharing. Thanks!

Issue: https://github.com/microsoft/durabletask-dotnet/issues/302

src/Analyzers/AttributeBinding/MatchingAttributeBindingAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// </summary>
+    /// <param name="knownTypeSymbols">The set of well-known types.</param>
+    /// <returns>The expected binding for this analyzer.</returns>
+    protected abstract ExpectedBinding GetExpectedBinding(KnownTypeSymbols knownTypeSymbols);
+
+    /// <summary>
+    /// After an incorrect attribute/type matching is found, this method is called so the concrete implementation can report a diagnostic.
+    /// </summary>
+    /// <param name="ctx">Context for a symbol action. Allows reporting a diagnostic.</param>
+    /// <param name="expected">Expected binding for a attribute/type.</param>
```
bachuv, line 63:
```suggestion
    /// <param name="expected">Expected binding for an attribute/type.</param>
```

bachuv (approved):

src/Analyzers/AttributeBinding/DurableClientBindingAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+/// <summary>
+/// Analyzer that matches 'DurableClientAttribute' with 'DurableTaskClient' parameters.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public sealed class DurableClientBindingAnalyzer : MatchingAttributeBindingAnalyzer
+{
+    /// <summary>
+    /// Diagnostic ID supported for the analyzer.
+    /// </summary>
+    public const string DiagnosticId = "DURABLE1002";
```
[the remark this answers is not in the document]
allantargino: We are partitioning diagnostics ids by category, and `1xxx` was intended to be associated with "Attribute Bindings" category. We have documented it in our internal Design Document, but I am going to move it to a public documentation after the other planned analyzers are added :)

src/Analyzers/KnownTypeSymbols.cs
allantargino: Great suggestion, thanks!
For now, I did a quick refactor into partial classes. In another PR, I am going to refactor it to group into subclasses, I loved the suggestion.

PR #303, opened 2024-05-06
CancellationToken roslyn analyzer

allantargino:
Currently, CancellationToken binding is not supported (ignored) during [input conversion](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs#L79). This analyzer will prevent users from using it inside Azure Functions Orchestrators.

### DIFF
--- a/src/Analyzers/Functions/Orchestration/CancellationTokenOrchestrationAnalyzer.cs
+++ b/src/Analyzers/Functions/Orchestration/CancellationTokenOrchestrationAnalyzer.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.DurableTask.Analyzers.Orchestration;
+using static Microsoft.DurableTask.Analyzers.Functions.Orchestration.CancellationTokenOrchestrationAnalyzer;
+
+namespace Microsoft.DurableTask.Analyzers.Functions.Orchestration;
+
+/// <summary>
+/// Analyzer that reports a warning when CancellationToken is used in an Azure Function Orchestration.
+/// </summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class CancellationTokenOrchestrationAnalyzer : OrchestrationAnalyzer<CancellationTokenOrchestrationVisitor>
+{
+    /// <summary>
+    /// Diagnostic ID supported for the analyzer.
+    /// </summary>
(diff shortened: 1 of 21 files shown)


src/Analyzers/Resources.resx
```diff
@@ -153,4 +153,10 @@
   <data name="DurableClientBindingAnalyzerTitle" xml:space="preserve">
     <value>[DurableClient] must be used with DurableTaskClient</value>
   </data>
+  <data name="CancellationTokenOrchestrationAnalyzerMessageFormat" xml:space="preserve">
+    <value>Orchestration '{0}' is using a non-supported CancellationToken binding</value>
+  </data>
+  <data name="CancellationTokenOrchestrationAnalyzerTitle" xml:space="preserve">
+    <value>Function CancellationToken binding is not supported for OrchestrationTrigger</value>
```
bachuv, line 160: Should we use the description that's in the in proc analyzer (pasted below) to keep it consistent? This description might be simpler.

> CancellationToken should not be used as an orchestrator function parameter.

src/Analyzers/Functions/Orchestration/CancellationTokenOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.DurableTask.Analyzers.Orchestration;
+using static Microsoft.DurableTask.Analyzers.Functions.Orchestration.CancellationTokenOrchestrationAnalyzer;
+
+namespace Microsoft.DurableTask.Analyzers.Functions.Orchestration;
+
+/// <summary>
+/// Analyzer that reports a warning when CancellationToken is used in an Azure Function Orchestration.
```
bachuv, line 14:
```suggestion
/// Analyzer that reports a warning when CancellationToken is used in a Durable Functions Orchestration.
```

src/Analyzers/Resources.resx
allantargino: Updated :)1

src/Analyzers/Resources.resx
```diff
@@ -153,4 +153,10 @@
   <data name="DurableClientBindingAnalyzerTitle" xml:space="preserve">
     <value>[DurableClient] must be used with DurableTaskClient</value>
   </data>
+  <data name="CancellationTokenOrchestrationAnalyzerMessageFormat" xml:space="preserve">
+    <value>Orchestration '{0}' is using a non-supported CancellationToken binding</value>
```
bachuv, line 157:
```suggestion
    <value>Orchestration '{0}' is using a CancellationToken binding, which is not supported</value>
```

bachuv (approved):
Just left a small suggestion, but otherwise looks good!

PR #305, opened 2024-05-06
Simplifying ITaskOrchestration detection logic

allantargino:
When detecting class based orchestrations, instead of checking individually for `TaskOrchestrator<T1,T2>` and `ITaskOrchestrator`, we can just whether check if a type implements `ITaskOrchestrator` and has a `TaskOrchestrationContext` parameter.


jviau (approved):
Looks great - you do have a couple build warnings.

PR #307, opened 2024-05-06
fix analyzer exception 'Syntax node is not within syntax tree'

allantargino:
During orchestration method analysis, since the syntax tree of the callee method might be different from the caller method, we need to get the correct semantic model when they differ.

### DIFF
--- a/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
+++ b/src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
@@ -316,7 +316,12 @@ public class MethodProbeOrchestrationVisitor : OrchestrationVisitor
             IEnumerable<MethodDeclarationSyntax> calleeSyntaxes = calleeMethodSymbol.GetSyntaxNodes();
             foreach (MethodDeclarationSyntax calleeSyntax in calleeSyntaxes)
             {
-                this.FindInvokedMethods(semanticModel, calleeSyntax, calleeMethodSymbol, rootOrchestration, reportDiagnostic);
+                // Since the syntax tree of the callee method might be different from the caller method, we need to get the correct semantic model,
+                // avoiding the exception 'Syntax node is not within syntax tree'.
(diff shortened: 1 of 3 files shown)


allantargino:
The exception was:
```
AzureFunctionsApp.cs(26,38): warning DURABLE0001: The method 'RunOrchestrator' uses 'System.DateTime.Now' that may cause non-deterministic behavior when invoked from orchestration 'AzureFunctionsApp' 
CSC : warning AD0001: Analyzer 'Microsoft.DurableTask.Analyzers.Orchestration.DateTimeOrchestrationAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. 
```

For reference, helped with the fix: https://github.com/dotnet/roslyn/issues/18730

src/Analyzers/Orchestration/OrchestrationAnalyzer.cs
```diff
@@ -316,7 +316,12 @@ protected virtual void VisitMethod(SemanticModel semanticModel, SyntaxNode metho
             IEnumerable<MethodDeclarationSyntax> calleeSyntaxes = calleeMethodSymbol.GetSyntaxNodes();
             foreach (MethodDeclarationSyntax calleeSyntax in calleeSyntaxes)
             {
-                this.FindInvokedMethods(semanticModel, calleeSyntax, calleeMethodSymbol, rootOrchestration, reportDiagnostic);
+                // Since the syntax tree of the callee method might be different from the caller method, we need to get the correct semantic model,
+                // avoiding the exception 'Syntax node is not within syntax tree'.
```
bachuv, line 320: Do we have a test that covers this scenario?

allantargino: Now I have haha

bachuv (approved):
Thanks for adding the test!

PR #309, opened 2024-05-10
Orchestration Banned APIs Roslyn analyzer

allantargino:
This PR adds 3 analyzers that verify whether orchestrations are using banned APIs through:

- Task.Run/Thread.Start/TaskFactory (`DURABLE0004`)
- Input/Output types (such as HttpClient, etc) (`DURABLE0005`)
- Environment Variables (`DURABLE0006`)

The list of I/O types is by no means comprehensive or complete, but rather an initial set of well-known types used in Azure Functions - we should augment that list soon. I tried to have feature parity with the [types covered by our previous analyzer](https://github.com/Azure/azure-functions-durable-extension/blob/eee1aee25d4609e00e113b73a711b262105945a3/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/IOTypesAnalyzer.cs#L53).

### DIFF
--- a/src/Analyzers/KnownTypeSymbols.Net.cs
+++ b/src/Analyzers/KnownTypeSymbols.Net.cs
@@ -17,7 +17,12 @@ public sealed partial class KnownTypeSymbols
     INamedTypeSymbol? thread;
     INamedTypeSymbol? task;
     INamedTypeSymbol? taskT;
+    INamedTypeSymbol? taskFactory;
+    INamedTypeSymbol? taskContinuationOptions;
+    INamedTypeSymbol? taskFactoryT;
     INamedTypeSymbol? cancellationToken;
+    INamedTypeSymbol? environment;
+    INamedTypeSymbol? httpClient;
 
     /// <summary>
     /// Gets a Guid type symbol.
@@ -40,7 +45,33 @@ public sealed partial class KnownTypeSymbols
     public INamedTypeSymbol? TaskT => this.GetOrResolveFullyQualifiedType(typeof(Task<>).FullName, ref this.taskT);
 
     /// <summary>
+    /// Gets a TaskFactory type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskFactory => this.GetOrResolveFullyQualifiedType(typeof(TaskFactory).FullName, ref this.taskFactory);
+
+    /// <summary>
+    /// Gets a TaskFactory&lt;T&gt; type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskFactoryT => this.GetOrResolveFullyQualifiedType(typeof(TaskFactory<>).FullName, ref this.taskFactoryT);
+
+    /// <summary>
+    /// Gets a TaskContinuationOptions type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskContinuationOptions => this.GetOrResolveFullyQualifiedType(typeof(TaskContinuationOptions).FullName, ref this.taskContinuationOptions);
+
+    /// <summary>
     /// Gets a CancellationToken type symbol.
     /// </summary>
     public INamedTypeSymbol? CancellationToken => this.GetOrResolveFullyQualifiedType(typeof(CancellationToken).FullName, ref this.cancellationToken);
+
+#pragma warning disable RS1035 // Environment Variables are not supposed to be used in Analyzers, but here we just reference the API, never using it.
+    /// <summary>
+    /// Gets an Environment type symbol.
+    /// </summary>
+    public INamedTypeSymbol? Environment => this.GetOrResolveFullyQualifiedType(typeof(Environment).FullName, ref this.environment);
+
+    /// <summary>
+    /// Gets an HttpClient type symbol.
+    /// </summary>
+    public INamedTypeSymbol? HttpClient => this.GetOrResolveFullyQualifiedType(typeof(HttpClient).FullName, ref this.httpClient);
 }
--- a/src/Analyzers/KnownTypeSymbols.Net.cs
+++ b/src/Analyzers/KnownTypeSymbols.Net.cs
@@ -69,6 +69,7 @@ public sealed partial class KnownTypeSymbols
     /// Gets an Environment type symbol.
     /// </summary>
     public INamedTypeSymbol? Environment => this.GetOrResolveFullyQualifiedType(typeof(Environment).FullName, ref this.environment);
+#pragma warning restore RS1035
 
     /// <summary>
     /// Gets an HttpClient type symbol.
--- a/src/Analyzers/Orchestration/EnvironmentOrchestrationAnalyzer.cs
+++ b/src/Analyzers/Orchestration/EnvironmentOrchestrationAnalyzer.cs
@@ -1,0 +1,74 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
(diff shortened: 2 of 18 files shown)


jviau (approved):
Some other types to consider: `IConfiguration`, `IOptions`, `IOptionsSnapshot`, `IOptionsMonitor`

jviau:
I do have a general question: what is the plan for analyzers when we make a feature which invalidates an analyzer? Not saying we have anything planned, but what if we do something that supplies some special idempotent clients?

src/Analyzers/Orchestration/IOOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// Visitor that inspects the method body for I/O operations by searching for specific types.
+    /// </summary>
+    public sealed class IOOrchestrationVisitor : MethodProbeOrchestrationVisitor
+    {
+        ImmutableArray<INamedTypeSymbol> bannedTypes;
+
+        /// <inheritdoc/>
+        public override bool Initialize()
+        {
+            List<INamedTypeSymbol> candidateSymbols = [
```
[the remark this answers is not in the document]
allantargino: hm when I tried, the compiler issues [`CS9176: There is no target type for the collection literal.` ](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/array-declaration-errors?f1url=%3FappId%3Droslyn%26k%3Dk(CS9176)#invalid-collection-initializer)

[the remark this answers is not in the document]
allantargino: Great suggestion on another types, I created an issue: #311

allantargino:
Regarding the invalidation of released analyzers, I think that depends whether the new feature is in a major or minor release.
- In a new major release, I think we can just completely erase the analyzer, releasing a new major of the analyzer too.
- In a new minor release, I researched a bit and there is really neat mechanism called [DiagnosticSuppressor](https://github.com/dotnet/roslyn/blob/main/docs/analyzers/DiagnosticSuppressorDesign.md) which I think we can use:
  > Provide an ability for platform/library authors to author simple, context-aware compiler extensions to programmatically suppress specific instances of reported analyzer and/or compiler diagnostics, which are always known to be false positives in the context of the platform/library.
  
  For instance, I created the following Supressor and `DURABLE0001` (DateTime diagnostic) stopped being shown:
  ```cs
  [DiagnosticAnalyzer(LanguageNames.CSharp)]
  public sealed class MySupressor : DiagnosticSuppressor
  {
      static SuppressionDescriptor Rule = new SuppressionDescriptor("SOMETHING123", "DURABLE0001", "no longer make sense");
      public override ImmutableArray<SuppressionDescriptor> SupportedSuppressions => [Rule];
  
      public override void ReportSuppressions(SuppressionAnalysisContext context)
      {
          foreach (var diagnostic in context.ReportedDiagnostics)
          {
              context.ReportSuppression(Suppression.Create(Rule, diagnostic));
          }
      }
  }
  ```

I am not sure if that's the best process available, but could work 👍

src/Analyzers/KnownTypeSymbols.Net.cs
```diff
@@ -39,8 +44,34 @@ public sealed partial class KnownTypeSymbols
...
+    /// Gets a TaskContinuationOptions type symbol.
+    /// </summary>
+    public INamedTypeSymbol? TaskContinuationOptions => this.GetOrResolveFullyQualifiedType(typeof(TaskContinuationOptions).FullName, ref this.taskContinuationOptions);
+
     /// <summary>
     /// Gets a CancellationToken type symbol.
     /// </summary>
     public INamedTypeSymbol? CancellationToken => this.GetOrResolveFullyQualifiedType(typeof(CancellationToken).FullName, ref this.cancellationToken);
+
+#pragma warning disable RS1035 // Environment Variables are not supposed to be used in Analyzers, but here we just reference the API, never using it.
```
bachuv, line 67: Do we need a corresponding `#pragma warning restore RS1035` after line 71?

src/Analyzers/Orchestration/EnvironmentOrchestrationAnalyzer.cs
```diff
@@ -0,0 +1,74 @@
...
+        MessageFormat,
+        AnalyzersCategories.Orchestration,
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true);
+
+    /// <inheritdoc/>
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => [Rule];
+
+    /// <summary>
+    /// Visitor that inspects the method body for retrivals of Environment Variables through the <see cref="System.Environment" /> type.
```
bachuv, line 40:
```suggestion
    /// Visitor that inspects the method body for retrievals of Environment Variables through the <see cref="System.Environment" /> type.
```

src/Analyzers/Orchestration/IOOrchestrationAnalyzer.cs
bachuv, line 1: Can you take a look at the possible null reference warnings in this file?

src/Analyzers/RoslynExtensions.cs
```diff
@@ -118,7 +118,8 @@ public static IEnumerable<MethodDeclarationSyntax> GetSyntaxNodes(this IMethodSy
     /// <returns>True if the method symbol is contained in the type symbol and has the method name, false otherwise.</returns>
     public static bool IsEqualTo(this IMethodSymbol methodSymbol, INamedTypeSymbol? typeSymbol, string methodName)
     {
-        return methodSymbol.ContainingType.Equals(typeSymbol, SymbolEqualityComparer.Default) &&
+        return (methodSymbol.ContainingType.Equals(typeSymbol, SymbolEqualityComparer.Default) ||
+            methodSymbol.ContainingType.OriginalDefinition.Equals(typeSymbol, SymbolEqualityComparer.Default)) &&
```
bachuv, line 122: Can you explain why we made this change?

test/Analyzers.Tests/Wrapper.cs
```diff
@@ -48,10 +48,16 @@ public static void Main()
...
+using Azure.Storage.Blobs;
+using Azure.Storage.Queues;
+using Azure.Data.Tables;
 using System;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Data.SqlClient;
```
bachuv, line 60: Do we need these using statements?

bachuv (commented):
Left some questions and comments, but overall looks good! Thanks!

test/Analyzers.Tests/Wrapper.cs
allantargino: Yes, we need them so the unit tests can run. The Wrapper methods add those usings to the wrapped classes/methods/etc. For the cases which the usings aren't needed (not all test classes/methods will use them all), the compiler will just discard them.

src/Analyzers/RoslynExtensions.cs
allantargino: To make both statements to work:
```cs
...
targetMethod.IsEqualTo(this.KnownTypeSymbols.Task, nameof(Task.Run)) ||
targetMethod.IsEqualTo(this.KnownTypeSymbols.TaskT, nameof(Task.Run)) ||
...
```

Not only I need to check if a given type is equal to the provided Symbol as an argument, but also if the type was constructed from a generic definition. Thus, the change.

src/Analyzers/KnownTypeSymbols.Net.cs
allantargino: done :)

bachuv (approved):
LGTM!

PR #310, opened 2024-05-10
Increase patch ver to 1.2.3

nytian:
As titled. Routine PR for v1.2.3 release.

DurableTask.Core v2.16.2 -> v2.17.0 and this has been updated in PR #290

### DIFF
--- a/eng/targets/Release.props
+++ b/eng/targets/Release.props
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.2.2</VersionPrefix>
+    <VersionPrefix>1.2.3</VersionPrefix>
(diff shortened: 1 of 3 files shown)


eng/targets/Release.props
```diff
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.2.2</VersionPrefix>
+    <VersionPrefix>1.2.3</VersionPrefix>
```
nytian, line 20: I wonder this should be v1.2.3 or v1.3.0. Since [PR#290](https://github.com/microsoft/durabletask-dotnet/pull/290/files) seems like a new feature.

davidmrdavid (approved):
#290 is more akin to a bug fix to me, so I think the patch-version update is correct

PR #314, opened 2024-05-21
 Add RetryPolicy.Handle Property to Allow for Exception Filtering on Retries

tomseida:
Add Handle delegate property to RetryPolicy and ensure it is passed to CoreRetryOptions so users of RetryPOlicy have ability to determine retry policy based on exception,

Testing shows it correctly constrains the number of times the task is called, but the check for the number of times the Handle method is called does not meet expectations. The Handle method is being invoked many more times than the task .

### DIFF
--- a/src/Abstractions/RetryPolicy.cs
+++ b/src/Abstractions/RetryPolicy.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Microsoft.DurableTask.Abstractions;
+
 namespace Microsoft.DurableTask;
 
 /// <summary>
@@ -20,7 +22,12 @@ public class RetryPolicy
     /// The maximum time to delay between attempts, regardless of<paramref name="backoffCoefficient"/>.
     /// </param>
     /// <param name="retryTimeout">The overall timeout for retries.</param>
-    /// <param name="handle">Delegate to call on exception to determine if retries should proceed.</param>
+    /// <param name="handle">
+    /// Optional delegate to invoke on exceptions to determine if retries should proceed. The delegate shall receive a
+    /// <see cref="TaskFailureDetails"/> instance and returns bool value where true means that a retry
+    /// is attempted and false means no retry is attempted. Time and attempt count constraints
+    /// take precedence over this delegate for determining if retry attempts are performed.
+    /// </param>
     /// <remarks>
     /// The value <see cref="Timeout.InfiniteTimeSpan"/> can be used to specify an unlimited timeout for
     /// <paramref name="maxRetryInterval"/> or <paramref name="retryTimeout"/>.
@@ -41,7 +48,7 @@ public class RetryPolicy
         double backoffCoefficient = 1.0,
         TimeSpan? maxRetryInterval = null,
         TimeSpan? retryTimeout = null,
-        Func<TaskFailedException, bool>? handle = null)
+        Func<TaskFailureDetails, bool>? handle = null)
     {
         if (maxNumberOfAttempts <= 0)
         {
@@ -88,7 +95,7 @@ public class RetryPolicy
         this.BackoffCoefficient = backoffCoefficient;
         this.MaxRetryInterval = maxRetryInterval ?? TimeSpan.FromHours(1);
         this.RetryTimeout = retryTimeout ?? Timeout.InfiniteTimeSpan;
-        this.SetHandler(handle);
+        this.Handle = CreateHandler(handle);
     }
 
     /// <summary>
@@ -134,42 +141,6 @@ public class RetryPolicy
     /// </value>
     public Func<Exception, bool> Handle { get; private set; }
 
-    /// <summary>
-    /// Set <see cref="Handle"/> delegate property.
-    /// </summary>
-    /// <param name="handle">
-    /// Deletegate that receives <see cref="TaskFailedException"/> and returns boolean that
-    /// determines if the task should be retried.
-    /// </param>
-    /// <exception cref="InvalidOperationException">
-    /// This represents a defect in this library in that it should always receive wither
-    /// <see cref="global::DurableTask.Core.Exceptions.TaskFailedException"/> or
-    /// <see cref="global::DurableTask.Core.Exceptions.SubOrchestrationFailedException"/>.
-    /// </exception>
-    void SetHandler(Func<TaskFailedException, bool>? handle)
-    {
-        this.Handle = handle is null
-            ? ((ex) => true)
-            : ((ex) =>
-            {
-                if (ex is global::DurableTask.Core.Exceptions.TaskFailedException globalTaskFailedException)
-                {
-                    var taskFailedException = new TaskFailedException(globalTaskFailedException.Name, globalTaskFailedException.ScheduleId, globalTaskFailedException);
-                    return handle.Invoke(taskFailedException);
-                }
-                else if (ex is global::DurableTask.Core.Exceptions.SubOrchestrationFailedException globalSubOrchestrationFailedException)
-                {
-                    var taskFailedException = new TaskFailedException(globalSubOrchestrationFailedException.Name, globalSubOrchestrationFailedException.ScheduleId, globalSubOrchestrationFailedException);
-                    return handle.Invoke(taskFailedException);
-                }
-                else
-                {
-                    throw new InvalidOperationException("TaskFailedException nor SubOrchestrationFailedException were not received.");
-                }
-            });
-    }
-
-
 #pragma warning disable SA1623 // Property summary documentation should match accessors
     /// <summary>
     /// This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler instead.
@@ -177,4 +148,41 @@ public class RetryPolicy
     [Obsolete("This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler instead.")]
     public Func<Exception, Task<bool>>? HandleAsync { get; set; }
 #pragma warning restore SA1623 // Property summary documentation should match accessors
+
+    /// <summary>
+    /// Create delegate for <see cref="Handle"/> property.
+    /// </summary>
+    /// <param name="handle">
+    /// Deletegate that receives <see cref="TaskFailedException"/> and returns boolean that
+    /// determines if the task should be retried.
+    /// </param>
+    /// <exception cref="InvalidOperationException">
+    /// This represents a defect in this library in that it should always receive either
+    /// <see cref="global::DurableTask.Core.Exceptions.TaskFailedException"/> or
+    /// <see cref="global::DurableTask.Core.Exceptions.SubOrchestrationFailedException"/>.
+    /// </exception>
+    static Func<Exception, bool> CreateHandler(Func<TaskFailureDetails, bool>? handle)
+    {
+        return handle is null
+            ? ((ex) => true)
+            : ((ex) =>
+            {
+                TaskFailureDetails? taskFailureDetails = null;
+                if (ex is global::DurableTask.Core.Exceptions.TaskFailedException taskFailedException)
+                {
+                    taskFailureDetails = taskFailedException.ToTaskFailureDetails();
+                }
+                else if (ex is global::DurableTask.Core.Exceptions.SubOrchestrationFailedException subOrchestrationFailedException)
+                {
+                    taskFailureDetails = subOrchestrationFailedException.ToTaskFailureDetails();
+                }
+
+                if (taskFailureDetails is null)
+                {
+                    throw new InvalidOperationException("Unable to create TaskFailureDetails since TaskFailedException nor SubOrchestrationFailedException was not received.");
+                }
+
+                return handle.Invoke(taskFailureDetails);
+            });
+    }
 }
--- a/src/Abstractions/RetryPolicy.cs
+++ b/src/Abstractions/RetryPolicy.cs
@@ -89,9 +89,7 @@ public class RetryPolicy
         this.BackoffCoefficient = backoffCoefficient;
         this.MaxRetryInterval = maxRetryInterval ?? TimeSpan.FromHours(1);
         this.RetryTimeout = retryTimeout ?? Timeout.InfiniteTimeSpan;
-#pragma warning disable CS0618 // Type or member is obsolete
         this.Handle = (ex) => true;
-#pragma warning restore CS0618 // Type or member is obsolete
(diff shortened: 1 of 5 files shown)


src/Abstractions/RetryPolicy.cs
```diff
@@ -39,7 +40,8 @@ public class RetryPolicy
         TimeSpan firstRetryInterval,
         double backoffCoefficient = 1.0,
         TimeSpan? maxRetryInterval = null,
-        TimeSpan? retryTimeout = null)
+        TimeSpan? retryTimeout = null,
+        Func<TaskFailedException, bool>? handle = null)
```
cgillum, line 44: Would it be simpler to use `TaskFailureDetails` instead of `TaskFailedException` for the parameter? It seems your test cases are only using the failure details anyways.

tomseida: I'm good with that change. Will do that tonight.

Any idea why the retry handler delegate would be getting called so many times? On a test case allowing ten retries, the handler delegate was called 108 times.

tomseida: Change to `TaskFailureDetails` is complete.

tomseida:
@cgillum, any timeline for reviewing the changes I made based on your comment?

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
```diff
@@ -222,6 +222,66 @@ public async Task RetryActivityFailuresCustomLogic(int expectedNumberOfAttempts,
...
+
+        string instanceId = await server.Client.ScheduleNewOrchestrationInstanceAsync(orchestratorName);
+        OrchestrationMetadata metadata = await server.Client.WaitForInstanceCompletionAsync(
+            instanceId, getInputsAndOutputs: true, this.TimeoutToken);
+
+        Assert.NotNull(metadata);
+        Assert.Equal(instanceId, metadata.InstanceId);
+        Assert.Equal(expRuntimeStatus, metadata.RuntimeStatus);
+        // More calls to retry handler than expected.
+        //Assert.Equal(expectedNumberOfAttempts, retryHandlerCalls);
```
cgillum, line 281: Should we be including the assert on `retryHandlerCalls`?

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
```diff
@@ -269,6 +329,76 @@ public async Task RetrySubOrchestrationFailures(int expectedNumberOfAttempts, Ty
...
```
cgillum, line 387: Same comment here.

src/Abstractions/RetryPolicy.cs
```diff
@@ -123,11 +133,56 @@ public class RetryPolicy
     /// </value>
     public TimeSpan RetryTimeout { get; }
 
+    /// <summary>
+    /// Gets a delegate to call on exception to determine if retries should proceed.
+    /// </summary>
+    /// <value>
+    /// Defaults delegate that always returns true (i.e., all exceptions are retried).
+    /// </value>
+    public Func<Exception, bool> Handle { get; private set; }
```
cgillum, line 142: Can the `private set` be removed?
```suggestion
    public Func<Exception, bool> Handle { get; }
```

cgillum (commented):
Just a few small things from me. @jviau, could you take a look as well?

src/Abstractions/RetryPolicy.cs
```diff
@@ -39,7 +47,8 @@ public class RetryPolicy
         TimeSpan firstRetryInterval,
         double backoffCoefficient = 1.0,
         TimeSpan? maxRetryInterval = null,
-        TimeSpan? retryTimeout = null)
+        TimeSpan? retryTimeout = null,
+        Func<TaskFailureDetails, bool>? handle = null)
```
jviau, line 51: This is a binary breaking change. We can't include it in this ctor. I recommend using `get; init;` on the property, and have the init callback set a private field with `CreateHandler`

src/Abstractions/RetryPolicy.cs
jviau, line 142: I'd rather we used the existing `HandleAsync`, but I understand the core options don't support async retry. But we _could_ support it - we just can't convert to core options anymore and need to do the retry entirely in this library. Which isn't the worst, it is just a fancy loop helper.

@cgillum what are your thoughts on sync and use core retry options, or async and implement it entirely here?

jviau (changes requested):

src/Abstractions/RetryPolicy.cs
cgillum: > what are your thoughts on sync and use core retry options, or async and implement it entirely here?

For this PR, I suggest we stick with sync and use core retry options to meet the current need. But I'd be open to reconsidering this as part of a larger refactor at some point to get rid of the DurableTask.Core dependency entirely.

jviau: Sounds good - my only concern then is the ctor breaking change.

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
tomseida: I think this should be part of the unit test, but it fails. I do not see how the changes in this PR would alter the flow of the code that calls into the RetryPolicy Handler. It appears to me that existing code is calling the handler more than once per activity function retry.

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
tomseida: Same reply as above.

src/Abstractions/RetryPolicy.cs
tomseida: I could not get the code to compile when removing the `private set`. Compiler said the property cannot be set since it was read only.  Keeping with the immutability of this class, I changed this to `private init`.

src/Abstractions/RetryPolicy.cs
tomseida: Returned the constructor to its original parameters and created property `HandleTaskFailureDetails` that has no getter and a public `init`. Remove the static method `CreateHandler` and moved that code into the `init` callback. Note that the ctor code sets `Handle` to a delegate that will retry on all exceptions.

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
cgillum: IIRC, the DurableTask.Core may have an old bug (which can't be fixed in a non-breaking way) that causes the retry handler to execute once more than its supposed to. That may be showing up here.

cgillum (approved):
I'm good with these changes.

src/Abstractions/RetryPolicy.cs
```diff
@@ -123,11 +126,56 @@ public class RetryPolicy
     /// </value>
     public TimeSpan RetryTimeout { get; }
 
+    /// <summary>
+    /// Gets a delegate to call on exception to determine if retries should proceed.
+    /// </summary>
+    /// <value>
+    /// Defaults delegate that always returns true (i.e., all exceptions are retried).
+    /// </value>
+    public Func<Exception, bool> Handle { get; private init; }
```
jviau, line 135: Not sure I like this being public - it is a bit misleading. Making it `internal` though will require moving other parts around. Can you add `[EditorBrowsable(EditorBrowsableState.Never)]` to this?

src/Abstractions/RetryPolicy.cs
```diff
@@ -123,11 +126,56 @@ public class RetryPolicy
...
+
 #pragma warning disable SA1623 // Property summary documentation should match accessors
     /// <summary>
     /// This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler instead.
     /// </summary>
     [Obsolete("This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler instead.")]
```
jviau, line 141:
```suggestion
    [Obsolete("This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler or HandleTaskFailureDetails instead.")]
```

src/Abstractions/RetryPolicy.cs
```diff
@@ -123,11 +126,56 @@ public class RetryPolicy
...
+    /// <see cref="TaskFailureDetails"/> instance and returns bool value where true means that a retry
+    /// is attempted and false means no retry is attempted. Time and attempt count constraints
+    /// take precedence over this delegate for determining if retry attempts are performed.
+    /// </summary>
+    /// <exception cref="InvalidOperationException">
+    /// This represents a defect in this library in that it should always receive either
+    /// <see cref="global::DurableTask.Core.Exceptions.TaskFailedException"/> or
+    /// <see cref="global::DurableTask.Core.Exceptions.SubOrchestrationFailedException"/>.
+    /// </exception>
+    public Func<TaskFailureDetails, bool> HandleTaskFailureDetails
```
jviau, line 156: Small naming nit:
```suggestion
    public Func<TaskFailureDetails, bool> HandleFailure
```

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
tomseida: The integrations tests do show that the activity functions are being executed the correct number of times and, thus, would not impact an orchestration's end result.

src/Abstractions/RetryPolicy.cs
tomseida: Left as public and added the attribute.

src/Abstractions/RetryPolicy.cs
tomseida: Changed.

src/Abstractions/RetryPolicy.cs
tomseida: Because I added the Obsolete attribute, the code generates the following warning twice:

`'RetryPolicy.Handle' is obsolete: 'This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler or HandleFailure instead.'	Abstractions	C:\DevOps\durabletask-dotnet\src\Abstractions\RetryPolicy.cs	163`

I have added the suppression of this warning around the two statements that set `Handle` so no warning appear.

src/Abstractions/RetryPolicy.cs
```diff
@@ -132,6 +135,8 @@ public class RetryPolicy
     /// <value>
     /// Defaults delegate that always returns true (i.e., all exceptions are retried).
     /// </value>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [Obsolete("This functionality is not implemented. Will be removed in the future. Use TaskOptions.FromRetryHandler or HandleFailure instead.")]
```
jviau, line 139: Oh, you didn't need to implement `Obsolete` on this method. I just wanted the `EditorBrowsable` attribute.

src/Abstractions/RetryPolicy.cs
jviau: May have been some miscommunication: can we add the ` or HandleFailure` to this existing `[Obsolete]` attribute? No need to add the attribute to the new `Handle` property. Because it has `private init;` it is well-enough protected.

src/Abstractions/RetryPolicy.cs
```diff
@@ -123,11 +129,60 @@ public class RetryPolicy
     /// </value>
     public TimeSpan RetryTimeout { get; }
 
+    /// <summary>
+    /// Gets a delegate to call on exception to determine if retries should proceed.
```
jviau, line 133: It may be useful to expand this comment, directing users to `HandleFailure`.

```suggestion
    /// Gets a delegate to call on exception to determine if retries should proceed.
   /// For internal usage. Use <see cref="HandleFailure" /> for setting this delegate.
```

src/Abstractions/RetryPolicy.cs
tomseida: Corrected.

jviau (approved):

Kobudzik:
Hello, is there any approximate date when will this be released?

cgillum:
@jviau or @davidmrdavid, when is the next planned release of the .NET Isolated SDK?

tomseida:
I have been monitoring nuget for updates. Once there is a prerelease, I will perform some additional testing. Also hoping the [Schedule orchestration feature](https://github.com/Azure/azure-functions-durable-extension/pull/2805) fix is also in the next release.

Fazer01:
Aah, we also would love the feature introduced with this PR. Awaiting new release. Any ETA on this? @jviau @tomseida?

tomseida:
> Aah, we also would love the feature introduced with this PR. Awaiting new release. Any ETA on this? @jviau @tomseida?

I have been checking twice a week to start using these changes. Not sure what's the hold-up. Would think at least a pre-release would be available.

tomseida:
@cgillum, any update on a release date? Odd that it has been over 3 months since the PR was completed and nothing has been made available.

cgillum:
I believe @jviau is still on leave until the end of this month. @davidmrdavid, @AnatoliB, or @lilyjma can we come up with a plan for updating the .NET Isolated release in the meantime?

davidmrdavid:
Should be out now: https://github.com/Azure/azure-functions-durable-extension/releases/tag/v1.1.6-worker-extension

tomseida:
I tested the new nuget packages and the HandleFailure method worked for me.  Thanks for finally get this released. Will save on unnecessary compute. 
image

Kobudzik:
Is there any reason why TaskFailureDetails are used as Func argument? I want to retry on two types of transient errors:

- when SqlServerTransientExceptionDetector.ShouldRetryOn(**Exception ex**) resolves true
- when exception is **ApiException** with status code 400/401/404

With current implementation it cannot be achieved as far as I know. Would it make sense to e.g. serialize the original Exception within TaskFailureDetails? Or if TaskFailureDetails shouldn't grow much, maybe there should be some other flow?

https://github.com/microsoft/durabletask-dotnet/blob/e0b54b8f8bc800301e9d82ab520b54151d14f6a6/src/Abstractions/RetryPolicy.cs#L145
This non-implemented function would fit well in my use case, is there any particular reason why wasn't it implemented?

cgillum:
Hi @Kobudzik.  I think your question(s) might be best to have in a separate GitHub issue since they don’t appear to be simple. Please create an issue and feel free to tag me there. That way we can go into more details without automatically notifying all the participants in this PR.

PR #316, opened 2024-05-22
Matching Input and Output Type Activity Analyzer

allantargino:
This new analyzer will search for `TaskOrchestrationContext.CallActivityAsync` invocations and check if the provided input and output types matches the related Activity definition. The activity can be either an Azure Function activity, a Durable Task Activity or a Durable Task Func/Action. If there is a mismatch between the types, the analyzer reports `DURABLE2001` (input mismatch) or `DURABLE2002` (output mismatch).

The code uses activity name to correlate the invocations and their definitions, which must be constant in compile time. For instance, they can be defined using `nameof` operator, string `const` fields or locals, and string literals. If the invocation name can't be determined (because it is not constant a compile time), the code ignores the checking to avoid a false positive report.

### DIFF
--- a/src/Analyzers/Resources.resx
+++ b/src/Analyzers/Resources.resx
@@ -183,4 +183,16 @@
   <data name="ThreadTaskOrchestrationAnalyzerTitle" xml:space="preserve">
     <value>Thread and Task calls must be deterministic inside an orchestrator function</value>
   </data>
+  <data name="InputArgumentTypeMismatchAnalyzerMessageFormat" xml:space="preserve">
+    <value>CallActivityAsync is passing the incorrect type '{0}' instead of '{1}' to the activity function '{2}'</value>
+  </data>
+  <data name="InputArgumentTypeMismatchAnalyzerTitle" xml:space="preserve">
+    <value>Activity function calls use the wrong argument type</value>
+  </data>
+  <data name="OutputArgumentTypeMismatchAnalyzerMessageFormat" xml:space="preserve">
+    <value>CallActivityAsync is expecting the return type '{0}' that does not match the return type '{1}' of the activity function '{2}'</value>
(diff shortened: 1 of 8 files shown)


src/Analyzers/Resources.resx
```diff
@@ -183,4 +183,16 @@
...
     <value>Thread and Task calls must be deterministic inside an orchestrator function</value>
   </data>
+  <data name="InputArgumentTypeMismatchAnalyzerMessageFormat" xml:space="preserve">
+    <value>CallActivityAsync is passing the incorrect type '{0}' instead of '{1}' to the activity function '{2}'</value>
+  </data>
+  <data name="InputArgumentTypeMismatchAnalyzerTitle" xml:space="preserve">
+    <value>Activity function calls use the wrong argument type</value>
+  </data>
+  <data name="OutputArgumentTypeMismatchAnalyzerMessageFormat" xml:space="preserve">
+    <value>CallActivityAsync is expecting the return type '{0}' that does not match the return type '{1}' of the activity function '{2}'</value>
```
bachuv, line 193:
```suggestion
    <value>CallActivityAsync is expecting the return type '{0}' and that does not match the return type '{1}' of the activity function '{2}'</value>
```

src/Analyzers/Resources.resx
```diff
@@ -183,4 +183,16 @@
...
+  </data>
+  <data name="OutputArgumentTypeMismatchAnalyzerTitle" xml:space="preserve">
+    <value>Activity function call return type doesn't match function definition return type</value>
```
bachuv, line 196:
```suggestion
    <value>Activity function call return type doesn't match the function definition return type</value>
```

bachuv (approved):
LGTM!

Issue #282, opened 2024-04-04
DurableTaskClient.PurgeAllInstancesAsync: NotSupportedException and RpcException for .net 8 isolated


th3ragex:
I am periodically purging instances and old history of my durable function runs. We noticed that this logic stopped working some time ago.

Runtime: 
.net 8 isolated azure function

Dependencies:
```xml
<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.21.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.1.2" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.2" />
```

My TimerTrigger Azure Function which is implemented next to my durable function looks likes this:

```csharp
    [Function(nameof(OrchestrationCleanupFunction))]
    public async Task Run(
        [TimerTrigger("0 */1 * * * *")] TimerInfo myTimer,
        [DurableClient] DurableTaskClient orchestrationClient)
    {
        var createdTimeFrom = DateTime.UtcNow.Subtract(TimeSpan.FromDays(365 * 5));
        var createdTimeTo = DateTime.UtcNow.Subtract(TimeSpan.FromDays(5));

        var runtimeStatus = new List<OrchestrationRuntimeStatus>
        {
            OrchestrationRuntimeStatus.Completed
        };

        await orchestrationClient.PurgeAllInstancesAsync(new PurgeInstancesFilter(createdTimeFrom, createdTimeTo, runtimeStatus));
    }
```
This code worked at some point and was initially developed as .Net 7 isolated function.

Now i am getting a System.NotSupportedException: `Microsoft.Azure.Functions.Worker.FunctionsDurableTaskClient does not support purging of orchestration instances.`

The implementation of `DurableTaskClient.PurgeAllInstancesAsync` is very confusing. 

- DurableTaskClient is abstract and has multiple virtual overloads of PurgeAllInstancesAsync with a default implementation that throws a System.NotSupportedException.
- Each derived type does not override all "virtual" variations of PurgeAllInstancesAsync. (GrpcDurableTaskClient & FunctionsDurableTaskClient)

Example:
FunctionsDurableTaskClient implements: `Task<PurgeResult> PurgeAllInstancesAsync(PurgeInstancesFilter filter, CancellationToken cancellation)` 
but not `public virtual Task<PurgeResult> PurgeAllInstancesAsync(
[...]

PR #317, opened 2024-05-22
Add optional cancellation tokens again in `DurableTaskClient`

davidmrdavid:
Between tags 1.1.1. and 1.2.0 (https://github.com/microsoft/durabletask-dotnet/compare/v1.1.1...v1.2.0#diff-c7a6ad429a79519f6c23e141c7d033c7066d66ef34af21302c877fbdd62171c7R373), we made several changes to `DurableTaskClient` that unfortunately changed regressed some behavior.

In particular, we made the following 3 changes:

| Before | After |
| -------| ------|
|     public virtual Task TerminateInstanceAsync(string instanceId, object? output = null, CancellationToken cancellation = default) |     public virtual Task TerminateInstanceAsync(string instanceId, object? output, CancellationToken cancellation = default)|
|     public virtual Task<PurgeResult> PurgeInstanceAsync(string instanceId, CancellationToken cancellation = default) |     public virtual Task<PurgeResult> PurgeInstanceAsync(string instanceId, CancellationToken cancellation) |
|     public virtual Task<PurgeResult> PurgeAllInstancesAsync(PurgeInstancesFilter filter, CancellationToken cancellation = default) |      public virtual Task<PurgeResult> PurgeAllInstancesAsync(PurgeInstancesFilter filter, CancellationToken cancellation) |

In other words, we made several previously "optional" parameters mandatory. This caused .NET to start dispatching **different** implementations for this method, which led to users seeing "method not supported" exceptions, as reported here: https://github.com/microsoft/durabletask-dotnet/issues/282#issuecomment-2114590648

This PR aims to bring back the original interfaces by making those parameters optional again, in hopes to restore the original method dispatch resolution.

_Still to be tested._

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -337,7 +337,7 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     public abstract AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(OrchestrationQuery? filter = null);
 
     /// <inheritdoc cref="PurgeInstanceAsync(string, PurgeInstanceOptions, CancellationToken)"/>
-    public virtual Task<PurgeResult> PurgeInstanceAsync(string instanceId, CancellationToken cancellation = default)
+    public virtual Task<PurgeResult> PurgeInstanceAsync(string instanceId, CancellationToken cancellation)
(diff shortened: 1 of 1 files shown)


davidmrdavid:
Converting this to draft due to ambiguity in the resolution still

davidmrdavid:
Abandoning, I think this can be resolved fully by implementing the interfaces in the worker extension.

PR #321, opened 2024-05-29
Benchmarking DateTimeOrchestrationAnalyzer

allantargino:
Adding a benchmark to compare DateTimeOrchestrationAnalyzer against a compilation with no analyzers and also against a compilation with an empty analyzer. The overall result (pasted below) shows an acceptable initial performance for the analyzer.


allantargino:
image

jviau (approved):

PR #323, opened 2024-06-05
DateTime, Guid and Task Delay roslyn code fixers

allantargino:
Following the same structure as #318, this PR adds roslyn code fixers for the diagnostics reported about non-deterministic usage of DateTime, Guid and Task delay. It suggests replacing them for the `TaskOrchestrationContext` alternatives, as suggested by [our docs](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-code-constraints?tabs=csharp).

### DIFF
--- a/test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
+++ b/test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
@@ -4,7 +4,7 @@
 using Microsoft.CodeAnalysis.Testing;
 using Microsoft.DurableTask.Analyzers.Orchestration;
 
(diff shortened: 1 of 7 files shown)


test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
```diff
@@ -45,7 +45,6 @@ public async Task NonOrchestrationHasNoDiag()
     [Theory]
     [InlineData("DateTime.Now")]
     [InlineData("DateTime.UtcNow")]
-    [InlineData("DateTime.Today")]
```
bachuv, line 48: Why did you remove this line?

allantargino: I extract that use case into a brand new `Fact` unit test. I did it because the expected fix for `DateTime.Today` is different from `Now` and `UtcNow`. So it was more a refactor than removing a test.

bachuv (approved):
LGTM!

PR #333, opened 2024-07-11
Add CodeQL checks

davidmrdavid:
As part of our 1ES migration, we need to add "CodeQL" (Code Query Language?) checks to our repos. This CodeQL service effectively checks against CVEs and other compliance requirements using static analysis of our source code. This needs to run in two places:

(1) Directly in the GitHub repo
(2) In our 1ES code-mirror

It needs to run in these two places because CodeQL doesn't realize our 1ES ADO repo is a clone of the GitHub repo. Additionally, CodeQL needs to run at least weekly.

To tackle these requirements, this PR does the following:
(1) Add GitHub action that runs CodeQL checks directly in GitHub. This runs weekly.
(2) It makes our 1ES Official pipeline, which automatically adds injects CodeQL checks, run weekly.

That's all.


YunchuWang:
no longer applicable

PR #335, opened 2024-07-24
Add codeQL support

davidmrdavid:
DurableTask-Dotnet version of this PR: https://github.com/Azure/durabletask/pull/1131

CodeQL (Code Query Language?) is a service that checks against CVEs and other compliance requirements using static analysis in our repos. It needs to be executed weekly.

Our 1ES pipeline already has this, but we need to make it run weekly, so our 1ES pipeline now has a weekly schedule.
We also need to add it directly on GitHub, which is done through the codeQL.yml GitHub workflow.


YunchuWang:
no longer applicable

PR #342, opened 2024-09-13
Add sourcelink reconstruction task in msbuild

davidmrdavid:
When building in 1ES, source-link information gets lost due to the code-mirror'ing. This creates issues during debugging, as Visual Studio would be unable to determine where to download the source code from. This PR aims to fix that by following conventions set by the Azure Functions Host.

This is the durabletask-dotnet version of this DTFx PR: https://github.com/Azure/durabletask/pull/1147

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -16,4 +16,38 @@
 
   <Import Project="$(EngRoot)targets/InternalsVisibleTo.targets" />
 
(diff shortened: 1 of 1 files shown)


nytian (approved):

cgillum (approved):
😵‍💫

PR #344, opened 2024-09-25
Add publish pipeline

davidmrdavid:
This PR adds a package publishing pipeline to automate releases to ADO and NuGet

### DIFF
--- a/eng/publish/publish.yml
+++ b/eng/publish/publish.yml
@@ -1,0 +1,281 @@
+# This is our package-publishing pipeline.
+# When executed, it automatically publishes the output of the 'official pipeline' (the nupkgs) to our internal ADO feed.
+# It may optionally also publish the packages to NuGet, but that is gated behind a manual approval.
+
+trigger: none # only trigger is manual
+pr: none # only trigger is manual
+
+resources:
+    repositories:
+        - repository: 1es
+          type: git
+          name: 1ESPipelineTemplates/1ESPipelineTemplates
+          ref: refs/tags/release
+        - repository: eng
+          type: git
+          name: engineering
+          ref: refs/tags/release
+
+    pipelines:
+    - pipeline: officialPipeline # Reference to the pipeline to be used as an artifact source
+      source: 'durabletask-dotnet.official'
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+        name: 1es-pool-azfunc
+        image: 1es-windows-2022
+        os: windows
+
+    stages:
+    - stage: release
+      jobs:
+
+      # ADO release
+      - job: AdoRelease
+        displayName: ADO Release
+        templateContext:
+          type: releaseJob
+          isProduction: true
+          inputs:
+          - input: pipelineArtifact
+            pipeline: officialPipeline  # Pipeline reference, as defined in the resources section
+            artifactName: drop
+            targetPath: $(System.DefaultWorkingDirectory)/drop
+        steps:
+        - task: 1ES.PublishNuget@1
+          displayName: 'Push to durabletask ADO feed'
+          inputs:
+            command: push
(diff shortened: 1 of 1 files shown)


nytian (approved):

eng/publish/publish.yml
```diff
@@ -0,0 +1,281 @@
...
+            artifactName: drop
+            targetPath: $(System.DefaultWorkingDirectory)/drop
+        steps:
+        - task: 1ES.PublishNuget@1
+          displayName: 'NuGet push (Microsoft.DurableTask.Abstractions)'
+          inputs:
+            command: push
+            nuGetFeedType: external
+            publishFeedCredentials: 'DurableTask org NuGet API Key'
+            packagesToPush: '$(System.DefaultWorkingDirectory)/drop/Microsoft.DurableTask.Abstractions.*.nupkg;!$(System.DefaultWorkingDirectory)/**/*.symbols.nupkg' # Despite this being a custom command, we need to keep this for 1ES validation
```
jviau, line 114: `!$(System.DefaultWorkingDirectory)/**/*.symbols.nupkg` needs to be included?` Weird - I don't believe we produce those. Instead we produce `.snupkg` (which I believe will automatically be included when a nupkg is pushed)

eng/publish/publish.yml
```diff
@@ -0,0 +1,281 @@
...
+          type: releaseJob
+          isProduction: true
+          inputs:
+          - input: pipelineArtifact
+            pipeline: officialPipeline  # Pipeline reference as defined in the resources section
+            artifactName: drop
+            targetPath: $(System.DefaultWorkingDirectory)/drop
+        steps:
+        - task: 1ES.PublishNuget@1
+          displayName: 'NuGet push (Microsoft.DurableTask.Client)'
```
jviau, line 132: Splitting these all into separate jobs looks quite cumbersome, especially now we will need to edit this every time a new project is added.

In legacy release, I worked around this by using the dotnet CLI for pushing to nuget.org. Did you try using `useDotNetTask: true` and `allowPackageConflicts: true`?

Hopefully with those 2 inputs this can be consolidated down to a single job.

jviau (commented):
Curious why you used the task approach instead of output?

eng/publish/publish.yml
```diff
@@ -0,0 +1,281 @@
...
+          name: engineering
+          ref: refs/tags/release
+
+    pipelines:
+    - pipeline: officialPipeline # Reference to the pipeline to be used as an artifact source
+      source: 'durabletask-dotnet.official'
+
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
+  parameters:
```
jviau, line 25: Consider adding:

``` suggestion
  parameters:
    featureFlags:
      sdlNugetPathOptimization: true
```

davidmrdavid: Can you tell me a bit more about what this does? I'd like to include some context as an inline comment

eng/publish/publish.yml
davidmrdavid: Oh, I don't recall seeing a `useDotNetTask` option, perhaps that's what I need. I can give it a go.
The key issue is that the 1ES  nuget release task has a limitation of not allowing package conflicts, so if this setting allows me to go back to the original nuget release task, then this would be it

eng/publish/publish.yml
davidmrdavid: The `/**/*.symbols.nupkg` bit appears here because the default value for `packagesToPush` includes it, and I was just modifying that default template. You're right, we should be able to remove it; let me know into that

eng/publish/publish.yml
jviau: I am not sure what it all does _exactly_. SDL is something 1ES team controls and their nuget release docs recommended adding it. Its probably fine to omit though

PR #358, opened 2024-12-17
Add parent trace context information when scheduling an orchestration

bachuv:
These changes send trace context information (`Activity.Current.Id`) through the protobuf CreateInstanceRequest.ParentTraceContext field when scheduling a new orchestration instance to help correlate client and orchestrator functions for distributed tracing.


src/Abstractions/TaskOptions.cs
```diff
@@ -100,4 +100,5 @@ public SubOrchestrationOptions(TaskOptions options, string? instanceId = null)
 /// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
 /// is specified, the orchestration instance will be scheduled immediately.
 /// </param>
-public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null);
+/// <param name="ParentTraceId">Parent Trace Id.</param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null, string? ParentTraceId = null);
```
jviau, line 104: We should not include trace information in the public API. Capturing of trace details should be done implicitly in the implementation by checking `Activity.Current`.

jviau (changes requested):

jviau (approved):

cgillum (approved):

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -83,6 +84,7 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
             Version = orchestratorName.Version,
             InstanceId = options?.InstanceId ?? Guid.NewGuid().ToString("N"),
             Input = this.DataConverter.Serialize(input),
+            ParentTraceContext = new P.TraceContext() { TraceParent = Activity.Current?.Id },
```
philliphoff, line 87: The `TraceState` property [must also be propagated](https://www.w3.org/TR/trace-context/#design-overview), I believe from the `Activity.Current?.TraceStateString` property.

jviau:
> Also, while this will push the trace context to the backend, and the backend may pass back it back, for example, as part of an activity request, unless the SDK creates an `ActivitySource` and `Activity` using that context, worker activity won't be associated with the trace.

The DurableTask.Core SDK has all of the appropriate `ActivitySource`'s implemented. This SDK will need it added though for customers using the `AzureManaged` backend.

philliphoff (approved):

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -84,9 +84,18 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
             Version = orchestratorName.Version,
             InstanceId = options?.InstanceId ?? Guid.NewGuid().ToString("N"),
             Input = this.DataConverter.Serialize(input),
-            ParentTraceContext = new P.TraceContext() { TraceParent = Activity.Current?.Id, TraceState = Activity.Current?.TraceStateString },
         };
 
+        if (Activity.Current?.Id != null)
+        {
+            request.ParentTraceContext.TraceParent = Activity.Current?.Id;
```
philliphoff, line 91: Is `request` guaranteed to have a valid `ParentTraceContext`?

bachuv: Can you clarify what you mean by "valid `ParentTraceContext`"?

philliphoff: @bachuv The question was whether `request.ParentTraceContext` was guaranteed to be non-`null`.

jviau: @bachuv - I believe @philliphoff is asking if `request.ParentTraceContext` may be `null`. In which case trying to access to `.TraceParent` will throw.

bachuv: Got it, I'll update the code to add a null check for `request.ParentTraceContext`.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -86,14 +86,22 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
             Input = this.DataConverter.Serialize(input),
         };
 
-        if (Activity.Current?.Id != null)
+        if (Activity.Current?.Id != null || Activity.Current?.TraceStateString != null)
         {
-            request.ParentTraceContext.TraceParent = Activity.Current?.Id;
-        }
+            if (request.ParentTraceContext == null)
```
jviau, line 91: Might be able to consolidate this to:

``` CSharp
 {
    request.ParentTraceContext = new()
    {
        TraceParent = activity.Id,
        TraceState = activity.TraceStateString,
    };
}
```

Reasons being:

1. We don't care if `ParentTraceContext` is already set or not, we can overwrite it.
2. if `Activity` is not null, we don't care if Id or TraceStateString are null or not - just set them.

jviau (approved):

PR #359, opened 2024-12-17
Update to latest protobuf definitions

cgillum:
Summary of changes:

- Updates to latest proto definitions in microsoft/durabletask-proto
- Implements new work item completion token APIs
- Updates System.Text.Json dependency to 6.0.10 to address CVEs
- Adds work item concurrency configuration
- Bump the version from 1.4.0 to 1.5.0 since there is new public surface area (concurrency configuration)

Note that changes to the worker packages generally aren't applicable to Durable Functions but rather are intended to support alternative "sidecar" endpoints, such as Dapr or DTS. The protobuf-related changes in particular are required for DTS.

Note that this change should be merged only after https://github.com/microsoft/durabletask-protobuf/pull/28 is merged. We also need to update the submodule pointer to point to the `main` branch rather than the PR branch.

### DIFF
--- a/src/Shared/Grpc/ProtoUtils.cs
+++ b/src/Shared/Grpc/ProtoUtils.cs
@@ -223,18 +223,21 @@ static class ProtoUtils
     /// <param name="instanceId">The orchestrator instance ID.</param>
     /// <param name="customStatus">The orchestrator customer status or <c>null</c> if no custom status.</param>
     /// <param name="actions">The orchestrator actions.</param>
+    /// <param name="completionToken">The completion token from the work item.</param>
     /// <returns>The orchestrator response.</returns>
     /// <exception cref="NotSupportedException">When an orchestrator action is unknown.</exception>
     internal static P.OrchestratorResponse ConstructOrchestratorResponse(
         string instanceId,
         string? customStatus,
-        IEnumerable<OrchestratorAction> actions)
+        IEnumerable<OrchestratorAction> actions,
+        string completionToken)
     {
         Check.NotNull(actions);
         var response = new P.OrchestratorResponse
         {
             InstanceId = instanceId,
             CustomStatus = customStatus,
+            CompletionToken = completionToken,
         };
 
         foreach (OrchestratorAction action in actions)
@@ -605,9 +608,12 @@ static class ProtoUtils
     /// Converts a <see cref="EntityBatchResult" /> to <see cref="P.EntityBatchResult" />.
     /// </summary>
     /// <param name="entityBatchResult">The operation result to convert.</param>
+    /// <param name="completionToken">The completion token from the work item.</param>
     /// <returns>The converted operation result.</returns>
     [return: NotNullIfNotNull(nameof(entityBatchResult))]
-    internal static P.EntityBatchResult? ToEntityBatchResult(this EntityBatchResult? entityBatchResult)
+    internal static P.EntityBatchResult? ToEntityBatchResult(
+        this EntityBatchResult? entityBatchResult,
+        string completionToken)
     {
         if (entityBatchResult == null)
         {
@@ -618,6 +624,7 @@ static class ProtoUtils
         {
             EntityState = entityBatchResult.EntityState,
             FailureDetails = entityBatchResult.FailureDetails.ToProtobuf(),
+            CompletionToken = completionToken,
         };
 
         foreach (OperationAction action in entityBatchResult.Actions!)
--- a/src/Shared/Grpc/ProtoUtils.cs
+++ b/src/Shared/Grpc/ProtoUtils.cs
@@ -624,7 +624,6 @@ static class ProtoUtils
         {
             EntityState = entityBatchResult.EntityState,
             FailureDetails = entityBatchResult.FailureDetails.ToProtobuf(),
-            CompletionToken = completionToken,
         };
 
         foreach (OperationAction action in entityBatchResult.Actions!)
--- a/src/Shared/Grpc/ProtoUtils.cs
+++ b/src/Shared/Grpc/ProtoUtils.cs
@@ -223,7 +223,10 @@ static class ProtoUtils
     /// <param name="instanceId">The orchestrator instance ID.</param>
     /// <param name="customStatus">The orchestrator customer status or <c>null</c> if no custom status.</param>
     /// <param name="actions">The orchestrator actions.</param>
-    /// <param name="completionToken">The completion token from the work item.</param>
+    /// <param name="completionToken">
+    /// The completion token for the work item. It must be the exact same <see cref="P.WorkItem.CompletionToken" />
+    /// value that was provided by the corresponding <see cref="P.WorkItem"/> that triggered the orchestrator execution.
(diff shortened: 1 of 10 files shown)


src/Worker/Core/DurableTaskWorkerOptions.cs
```diff
@@ -83,6 +83,21 @@ public DataConverter DataConverter
     /// </remarks>
     public TimeSpan MaximumTimerInterval { get; set; } = TimeSpan.FromDays(3);
 
+    /// <summary>
+    /// Gets or sets the maximum number of concurrent activity work items that can be processed by the worker.
+    /// </summary>
+    public int MaximumConcurrentActivityWorkItems { get; set; } = 100 * Environment.ProcessorCount;
```
jviau, line 89: What do you think of grouping these into a sub-class?

``` Csharp
public ConcurrencyOptions Concurrency { get; } = new();

// can be a nested class
public class ConcurrencyOptions
{
    public int MaximumActivityWorkItems { get; set; }

    // and the other fields.
}
```

jviau (commented):
I am curious what this `completionToken` concept is? There isn't any details in this PR about what it does.

cgillum:
> I am curious what this `completionToken` concept is? There isn't any details in this PR about what it does.

Yes, sorry for not including that. The basic idea is that each work item received from the backend contains a completion token, which is a backend-specific opaque string value. When a worker wants to complete or abandon the work item, it includes the completion token in the call to complete or abandon that work item. It's how the server can correlate completion or abandon requests with a particular work item. The idea is that every work item has a unique completion token. Most implementations are expected to use a GUID in this value.

Previously, we used things like orchestration instance IDs and task IDs for the purpose of work item correlation. However, it would cause problems in certain edge cases because these values aren't strictly unique across work items (like when doing ContinueAsNew). There are also some concerns about whether it's safe to use user-generated instance IDs for certain types of internal book-keeping. For DTS, we decided that we needed a more robust way to keep track of work items, and completion tokens was the solution. Legacy sidecars (like Dapr) can continue to rely on the old mechanism, but newer sidecars/servers like DTS should use completion tokens to keep track of which work items are being completed/abandoned.

src/Worker/Core/DurableTaskWorkerOptions.cs
cgillum: I'm fine either way, but a flat list seemed simple enough.

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -605,9 +608,12 @@ internal static OrchestrationStatus ToCore(this P.OrchestrationStatus status)
     /// Converts a <see cref="EntityBatchResult" /> to <see cref="P.EntityBatchResult" />.
     /// </summary>
     /// <param name="entityBatchResult">The operation result to convert.</param>
+    /// <param name="completionToken">The completion token from the work item.</param>
     /// <returns>The converted operation result.</returns>
     [return: NotNullIfNotNull(nameof(entityBatchResult))]
-    internal static P.EntityBatchResult? ToEntityBatchResult(this EntityBatchResult? entityBatchResult)
+    internal static P.EntityBatchResult? ToEntityBatchResult(
+        this EntityBatchResult? entityBatchResult,
+        string completionToken)
```
bachuv, line 616: Where is `completionToken` used in `ToEntityBatchResult()`?

cgillum: It's not being used. I accidentally left this in when working on another PR to add Durable Entity support to DTS. I can remove this.

jviau:
> > I am curious what this `completionToken` concept is? There isn't any details in this PR about what it does.
> 
> Yes, sorry for not including that. The basic idea is that each work item received from the backend contains a completion token, which is a backend-specific opaque string value. When a worker wants to complete or abandon the work item, it includes the completion token in the call to complete or abandon that work item. It's how the server can correlate completion or abandon requests with a particular work item. The idea is that every work item has a unique completion token. Most implementations are expected to use a GUID in this value.
> 
> Previously, we used things like orchestration instance IDs and task IDs for the purpose of work item correlation. However, it would cause problems in certain edge cases because these values aren't strictly unique across work items (like when doing ContinueAsNew). There are also some concerns about whether it's safe to use user-generated instance IDs for certain types of internal book-keeping. For DTS, we decided that we needed a more robust way to keep track of work items, and completion tokens was the solution. Legacy sidecars (like Dapr) can continue to rely on the old mechanism, but newer sidecars/servers like DTS should use completion tokens to keep track of which work items are being completed/abandoned.

Is it safe to consider this a replacement for (and maybe even improvement upon) executionId?

cgillum:
> Thanks for the explanation! Is it safe to consider this a replacement for (and maybe even improvement upon) executionId?

Almost. ExecutionId only applies to orchestrations, and the same execution ID will continue to exist for the lifetime of that orchestration instance. However, there's no equivalent concept for activities. This approach gives us a general-purpose identifier that we can use for work items of any type (including entities).

src/Shared/Grpc/ProtoUtils.cs
cgillum: I've removed this. We can add it back when we finish adding entity support for DTS.

src/Worker/Core/DurableTaskWorkerOptions.cs
jviau: I am a fan of nesting relevant options together. Mostly for organization for both ourselves as maintainers, and for users for discovering all concurrency related config together in one spot (and not having unrelated config mixed in).

But I am not so adamant about it to block a PR on that. Just something for you to consider.

jviau (approved):

bachuv (approved):

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -223,18 +223,21 @@ internal static Timestamp ToTimestamp(this DateTime dateTime)
     /// <param name="instanceId">The orchestrator instance ID.</param>
     /// <param name="customStatus">The orchestrator customer status or <c>null</c> if no custom status.</param>
     /// <param name="actions">The orchestrator actions.</param>
+    /// <param name="completionToken">The completion token from the work item.</param>
```
nytian, line 226: nit: Could we add more details here? Like the completion token generated from the DTS backend, as we don’t have this information for other backends.

nytian (approved):

src/Worker/Core/DurableTaskWorkerOptions.cs
cgillum: I went ahead and added it. I can see this being useful over time as we add more options.

src/Shared/Grpc/ProtoUtils.cs
cgillum: I updated the comment here to provide a bit more context. Note that it's not really a "backend" concept per-se, but more of a remote durable task engine (which we sometimes call the "sidecar") concept. For example, the implementation could be different for Dapr or for DTS, and it doesn't really matter what kind of backend store is used.

PR #362, opened 2025-01-09
durable task scheduler auth extension support

YunchuWang:
This pull request introduces several significant changes, primarily focusing on the setup of new projects and the integration of Azure Managed extensions for the Durable Task Framework client. The most important changes include the addition of new projects to the solution, updates to the build configuration, and the implementation of new client extensions and options for Azure Managed services.

### New Projects and Build Configuration:
* Added new projects `Worker.AzureManaged`, `Client.AzureManaged`, `Shared.AzureManaged.Tests`, `Client.AzureManaged.Tests`, and `Worker.AzureManaged.Tests` to the solution file `Microsoft.DurableTask.sln`.
* Updated the build configuration in `Microsoft.DurableTask.sln` to include the newly added projects. [[1]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eR198-R225) [[2]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eR261-R265)
* Set up .NET 6.0 in the GitHub Actions workflow file `.github/workflows/validate-build.yml`.

### Azure Managed Client Implementation:
* Created the `Client.AzureManaged.csproj` project file with necessary package references and project dependencies.
* Implemented `DurableTaskSchedulerClientExtensions` to provide extension methods for configuring Durable Task clients to use the Azure Durable Task Scheduler service.
* Defined `DurableTaskSchedulerClientOptions` to configure the Durable Task Scheduler, including methods for creating gRPC channels and handling connection strings.
* Added `AccessTokenCache` to manage and refresh Azure access tokens.
* Implemented `DurableTaskSchedulerConnectionString` to parse and handle connection strings for the Durable Task Scheduler service.


src/Extensions/Azure/DurableTaskSchedulerConnectionString.cs
```diff
@@ -0,0 +1,82 @@
+﻿// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.All rights reserved.
+// ------------------------------------------------------------
+
+using System.Data.Common;
+
+namespace DurableTask.Extensions.Azure;
```
cgillum, line 7: The namespace for all these new files must start with `Microsoft.`

src/Extensions/Azure/DurableTaskSchedulerOptions.cs
```diff
@@ -0,0 +1,147 @@
...
+    public string? ResourceId { get; set; }
+
+    /// <summary>
+    /// The worker ID used to identify the worker instance.
+    /// The default value is a string containing the machine name and the process ID.
+    /// </summary>
+    public string? WorkerId { get; set; }
+
+
+    public static DurableTaskSchedulerOptions FromConnectionString(string connectionString)
```
cgillum, line 50: Please add XML doc comments for all public methods.

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+            endpoint = $"https://{endpoint}";
+        }
+
+        string resourceId = options.ResourceId ?? "https://durabletask.io";
+#if NET6_0
+        int processId = Environment.ProcessId;
+#else
+        int processId = Process.GetCurrentProcess().Id;
+#endif
+        string workerId = options.WorkerId ?? $"{Environment.MachineName},{processId},{Guid.NewGuid()}";
```
cgillum, line 89:
```suggestion
        string workerId = options.WorkerId ?? $"{Environment.MachineName},{processId},{Guid.NewGuid():N}";
```

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,156 @@
+﻿using Azure.Core;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Worker;
+using System.Diagnostics;
+
+namespace DurableTask.Extensions.Azure;
+
+// NOTE: These extension methods will eventually be provided by the Durable Task SDK itself.
```
cgillum, line 8: You can remove this comment now.

cgillum (commented):
Looking great! I added some comments.

cgillum:
@jviau would you mind taking a look at this as well?

The main feedback I'm interested in from you is the package name/organization. The idea behind this PR is to introduce a mechanism for configuring the gRPC channels to work with the Durable Task Scheduler (DTS). This includes setting up the appropriate auth and some other service-specific headers (like the task hub name). I figured this might be the right time to introduce an "Extensions" concept and make "Azure" the extension consumers should use for talking to DTS. There could be others in the future. WDYT?

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,182 @@
...
+    public static void UseDurableTaskScheduler(
+        this IDurableTaskWorkerBuilder builder,
+        string endpointAddress,
+        string taskHubName,
+        TokenCredential credential,
+        Action<DurableTaskSchedulerOptions>? configure = null)
+    {
+        DurableTaskSchedulerOptions options = new(endpointAddress, taskHubName, credential);
+
+        configure?.Invoke(options);
```
jviau, line 32: This class should leverage options configuration instead of directly calling `configure?.Invoke`.

IE:

``` csharp
builder.Services.Configure(builder.Name, configure);
```

You will then need to make some adjustments to `UseGrpc` call:

``` CSharp
// method body
if (configure is not null)
{
    builder.Services.Configure(builder.Name, configure); // from above, to configure this options object
}

builder.Services.TryAddEnumerable(
    ServiceDescription.Singleton<IConfigureOptions<GrpcDurableTaskWorkerOptions>, ConfigureGrpcChannel>());
builder.UseGrpc(_ => { }); // empty callback as the line above does the work. This call is to get the build target registered.
// end method body

// helper internal class
class ConfigureGrpcChannel(IOptionsMonitor<DurableTaskSchedulerOptions> schedulerOptions) : IConfigureNamedOptions<GrpcDurableTaskWorkerOptions>
{
    public void Configure(GrpcDurableTaskWorkerOptions options) => this.Configure(Options.DefaultName, options);

    public void Configure(string name, GrpcDurableTaskWorkerOptions options)
    {
        DurableTaskSchedulerOptions source = schedulerOptions.Get(name);
        options.Channel = source.CreateChannel(); // method added to options object.
    }
}
```

The benefit here is now users can call chain _and_ they can use options configuration directly.

src/Extensions/Azure/DurableTaskSchedulerOptions.cs
```diff
@@ -0,0 +1,162 @@
...
+
+/// <summary>
+/// Options for configuring the Durable Task Scheduler.
+/// </summary>
+public class DurableTaskSchedulerOptions
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DurableTaskSchedulerOptions"/> class.
+    /// </summary>
+    internal DurableTaskSchedulerOptions(string endpointAddress, string taskHubName, TokenCredential? credential = null)
```
jviau, line 18: To follow options pattern, this will need a parameterless constructor. You can leverage options validation to ensure all required fields are set.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-9.0#options-validation

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
jviau: Want to also mention that this is an important change. It will allow for more flexible configuration of this, such as consuming configuration from more varied sources (such as Azure AppConfiguration). With this, an advanced user could write the following code:

```
builder.Services.AddOptions<DurableTaskSchedulerOptions>(builder.Name)
    .Configure<IConfiguration>((options, config) => options.Endpoint = config["SomeConfigKeyThatHasMyEndpoint"]);
```

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,182 @@
...
+                UnsafeUseInsecureChannelCallCredentials = true,
+            });
+    }
+
+    static Exception RequiredOptionMissing(string optionName)
+    {
+        return new ArgumentException(message: $"Required option '{optionName}' was not provided.");
+    }
+
+    sealed class TokenCache(TokenCredential credential, TokenRequestContext context, TimeSpan margin)
```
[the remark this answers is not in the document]
jviau: I found this: maybe you can copy and adapt it for usage here: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Pipeline/BearerTokenAuthenticationPolicy.cs#L219

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            throw RequiredOptionMissing(nameof(options.TaskHubName));
+        }
+
+        string taskHubName = options.TaskHubName;
+        string endpoint = options.EndpointAddress;
+
+        if (!endpoint.Contains("://"))
+        {
+            endpoint = $"https://{endpoint}";
```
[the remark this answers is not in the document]
cgillum: Not sure what the best practice is, but we did this because a ton of people were making mistakes in terms of whether or not to include the `https://` prefix, so we decided to just support both.

src/Extensions/Azure/DurableTaskSchedulerOptions.cs
```diff
@@ -0,0 +1,162 @@
...
+    /// <summary>
+    /// Gets the credential used to authenticate with the Durable Task Scheduler task hub resource.
+    /// </summary>
+    public TokenCredential? Credential { get; }
+
+    /// <summary>
+    /// Gets or sets the resource ID of the Durable Task Scheduler resource.
+    /// The default value is https://durabletask.io.
+    /// </summary>
+    public string? ResourceId { get; set; }
```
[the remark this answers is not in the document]
cgillum: This is not part of the endpoint address. We could include it as part of the connection string, but I'm not sure yet whether/when we'll need to make this generally configurable. I suggest we not worry about this until we're actually ready to test other clouds.

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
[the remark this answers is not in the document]
jviau: Understood. I wonder if we can move the manipulation to the options object, so it will prepend the `https://` in the setter of `Endpoint`.

src/Extensions/Azure/DurableTaskSchedulerOptions.cs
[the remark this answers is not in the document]
jviau: Oh it isn't part of the address? So endpoints aren't of the form `{my-resource}.{region}.durabletask.io`?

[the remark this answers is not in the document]
cgillum: No - this is an OAuth 2.0 resource ID. It represents the entire DTS service owned by us. It's not a customer-owned scheduler resource.

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        string resourceId = options.ResourceId ?? "https://durabletask.io";
+        int processId = Environment.ProcessId;
+        string workerId = options.WorkerId ?? $"{Environment.MachineName},{processId},{Guid.NewGuid():N}";
```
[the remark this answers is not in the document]
cgillum: No formality yet. We expect to iterate on this over time.

jviau:
> I figured this might be the right time to introduce an "Extensions" concept and make "Azure" the extension consumers should use for talking to DTS. There could be others in the future. WDYT?

I wonder if we want a name other that _Extensions_ for backends? I am worried this may get confused with a backend vs helpers for Azure / a specific Azure service. Say we wanted to port an AzureStorage backend hook to durable - where would that go? Or if we wanted to introduce helpers (not backends) for interacting with Azure services (Compute, Storage, CosmosDB, etc). Where would those go?

There is the `Worker` and `Client` naming for this already. Worker's are backends, Clients are for interacting with a backend. Would separate `Worker.AzureManaged` and `Client.AzureManaged` (or just "Azure") be better?

This also brings up the point of keeping a separate Worker and Client package. While it is more setup work on our end, the benefit is customers can bring in just the clients for their API projects and not need the entire worker.

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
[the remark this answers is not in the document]
YunchuWang: https://github.com/Azure/azure-sdk-for-net/blob/8992dc64edcedebcbde48e33f376d8489e4a8d19/sdk/identity/Azure.Identity/samples/TokenCache.md
https://azuresdkdocs.blob.core.windows.net/%24web/dotnet/Azure.Identity/1.13.1/index.html?utm_source=chatgpt.com

{34383AAB-75E3-480E-8EE7-28B3F7CC9DD5}

 yes i checked azure.identity 1.31.1 doc mentions
 Thread safety
We guarantee that all credential instance methods are thread-safe and independent of each other ([guideline](https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-service-methods-thread-safety)). This ensures that the recommendation of reusing credential instances is always safe, even across threads.

also default it use in memory cache. our current implementation dont need reimplement these

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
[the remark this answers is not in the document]
YunchuWang: added todo

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
YunchuWang: usedtscheduler functions are for client and worker both. if using builder.Name which defaults to options.defaultvalue, wont they override each other? 
{DF322177-4FD3-478E-806C-9AD6884AC088}
how about    separating it out to another func to configure options once,  public static void ConfigureDurableTaskSchedulerOptions(
        this IServiceCollection services,
        Action<DurableTaskSchedulerOptions> configure)
    {

        services.Configure("somename", configure);
    }

src/Extensions/Azure/DurableTaskSchedulerOptions.cs
YunchuWang: updated

src/Extensions/Azure/DurableTaskSchedulerExtensions.cs
YunchuWang: updated

src/Extensions/Azure/Azure.csproj
```diff
@@ -0,0 +1,27 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <PackageDescription>Azure extensions for the Durable Task Framework.</PackageDescription>
+    <EnableStyleCop>true</EnableStyleCop>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.13.1" />
+    <PackageReference Include="Grpc.Net.Client" Version="2.67.0" />
```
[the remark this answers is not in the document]
YunchuWang: https://learn.microsoft.com/en-us/aspnet/core/grpc/clientfactory?view=aspnetcore-9.0 i checked this is a good improvement but require changes out of scope for this pr purpose. 
, we are creating and passing grpcchannel to GrpcDurableTaskClientOptions in this package which is then used to configure grpcdurabletaskclient in 
{4DB5163E-A2DD-428D-9E0D-D52FD1DAA9A2}
but to use the clientfactory we need to change usegrpc to configure the grpcclient to use the httpclientfactory features ex 
{DC89C573-E16E-4BE9-A965-4AD3E2597BDD}

i will send another pr to try integrate this

YunchuWang:
> > I figured this might be the right time to introduce an "Extensions" concept and make "Azure" the extension consumers should use for talking to DTS. There could be others in the future. WDYT?
> 
> I wonder if we want a name other that _Extensions_ for backends? I am worried this may get confused with a backend vs helpers for Azure / a specific Azure service. Say we wanted to port an AzureStorage backend hook to durable - where would that go? Or if we wanted to introduce helpers (not backends) for interacting with Azure services (Compute, Storage, CosmosDB, etc). Where would those go?
> 
> There is the `Worker` and `Client` naming for this already. Worker's are backends, Clients are for interacting with a backend. Would separate `Worker.AzureManaged` and `Client.AzureManaged` (or just "Azure") be better?
> 
> This also brings up the point of keeping a separate Worker and Client package. While it is more setup work on our end, the benefit is customers can bring in just the clients for their API projects and not need the entire worker.

any advice on jacob's comment @cgillum ?

YunchuWang:
@jviau thanks for all the comments! i updated the pr and its ready for another review, added more tests and tested it manually with sample and working

cgillum:
> There is the `Worker` and `Client` naming for this already. Worker's are backends, Clients are for interacting with a backend. Would separate `Worker.AzureManaged` and `Client.AzureManaged` (or just "Azure") be better?
>
> This also brings up the point of keeping a separate Worker and Client package. While it is more setup work on our end, the benefit is customers can bring in just the clients for their API projects and not need the entire worker.

@jviau this makes sense to me. How about this - there will be two new nuget packages:

* **Microsoft.DurableTask.Client.AzureManaged**
* **Microsoft.DurableTask.Worker.AzureManaged**

In terms of project structure, how about this?

* `/src/Client/AzureManaged/Client.AzureManaged.csproj`
* `/src/Worker/AzureManaged/Worker.AzureManaged.csproj`
* `/src/Shared/AzureManaged/*.cs`

The `/src/Shared/AzureManaged` directory would just be shared .cs files that get imported into both projects.

YunchuWang:
split into separate packages and split tests, also add @cgillum 's sample to here using split packages and tested working
{9C0F3874-3AD9-459E-AF9F-4A643215CF8A}, can i get approval?

src/Shared/AzureManaged/Shared.AzureManaged.csproj
```diff
@@ -0,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
cgillum, line 1: There should not be a shared project. Rather, we should important the code directly into the both the client and worker projects. The other directories under `/src/Shared` as examples.

The reason for this is that we don't want to have to publish and maintain a "Shared" nuget package.

jviau: Correct - move these files to `src/Shared/AzureManaged`, then include them via:

``` xml
    <SharedSection Include="AzureManaged" />
```

src/Shared/AzureManaged/DurableTaskSchedulerOptions.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Globalization;
+using Azure.Core;
+using Azure.Identity;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Options for configuring the Durable Task Scheduler.
+/// </summary>
+public class DurableTaskSchedulerOptions
```
jviau, line 13: Typically we don't want public types in shared files, as then users get conflicting copies of this type when they import the two different packages.

We may need a `Microsoft.DurableTask.Abstractions.AzureManaged` package to unify these public types.

src/Shared/AzureManaged/DurableTaskSchedulerOptions.cs
```diff
@@ -0,0 +1,164 @@
...
+            Credential = GetCredentialFromConnectionString(connectionString),
+        };
+
+    /// <summary>
+    /// Creates a gRPC channel for communicating with the Durable Task Scheduler service.
+    /// </summary>
+    /// <returns>A configured <see cref="GrpcChannel"/> instance that can be used to make gRPC calls.</returns>
+    public GrpcChannel CreateChannel()
+    {
+        Check.NotNullOrEmpty(this.EndpointAddress, nameof(this.EndpointAddress));
```
jviau, line 80: `Check` is used for argument validation. Since these are not explicitly supplied arguments, use `Verify.NotNull` instead (you can also add a `Verify.NotNullOrEmpty` method to `Verify.cs` if you want.) Verify is used for verifying your expected state is correct.

samples/AspNetWebApp/DockerFile
jviau, line 1: Is there a reason we are including a dockerfile? Wondering why a docker sample is necessary at all.

And if you really want dockerization, use .net SDK's built in container publishing support: https://learn.microsoft.com/en-us/dotnet/core/containers/sdk-publish (no need for a dockerfile! just some msbuild props and then `dotnet publish -t:PublishContainer`)

samples/AspNetWebApp/ScenariosController.cs
```diff
@@ -0,0 +1,50 @@
...
+            return this.BadRequest(new { error = "A 'count' query string parameter is required and it must contain a positive number." });
+        }
+
+        // Generate a semi-unique prefix for the instance IDs to simplify tracking
+        prefix ??= $"hellocities-{count}-";
+        prefix += DateTime.UtcNow.ToString("yyyyMMdd-hhmmss");
+
+        this.logger.LogInformation("Scheduling {count} orchestrations with a prefix of '{prefix}'...", count, prefix);
+
+        Stopwatch sw = Stopwatch.StartNew();
```
jviau, line 31: Recommend removing the stopwatch log. This is redundant with distributed tracing, which will already calculate the duration for this API call.

If you really want to have a sample showing duration to enqueue orchestrations, I recommend looking into custom `ActivitySource` and using an internal activity (distributed tracing again)

samples/AspNetWebApp/Utils.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace AspNetWebApp;
+
+static class Utils
+{
+    public static async Task ParallelForEachAsync<T>(this IEnumerable<T> items, int maxConcurrency, Func<T, Task> action)
```
jviau, line 5: C# has a `Parallel` helper already. I recommend using that:

https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.parallel.foreach?view=net-9.0

src/Shared/AzureManaged/Shared.AzureManaged.csproj
```diff
@@ -0,0 +1,25 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
jviau, line 4: `TargetFrameworks` (with an 's') is for multi-targeting.

```suggestion
    <TargetFramework>net6.0</TargetFramework>
```

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
```diff
@@ -0,0 +1,182 @@
...
+using Azure.Core;
+using Azure.Identity;
+using FluentAssertions;
+using Grpc.Net.Client;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Client.Grpc;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Moq;
+using System.ComponentModel.DataAnnotations;
```
jviau, line 13: nit: System should come first.

src/Shared/AzureManaged/Shared.AzureManaged.csproj
YunchuWang: i like number 2, 3 packages seem a bit cumbersome, basically we do a rename and go back to original approach before/ @cgillum for advice

samples/AspNetWebApp/ScenariosController.cs
cgillum: I don't think this is a good sample for us to include in the public repo. The original code was not designed to be a sample, but rather to be a performance testing benchmark tool. I suggest we do something simpler for this PR.

samples/AspNetWebApp/DockerFile
YunchuWang: this is for using for deploying it to container app, isnt keeping it here easier for customer to try it out?

src/Shared/AzureManaged/Shared.AzureManaged.csproj
YunchuWang: as discussed with Chris offline, he suggested to keep two packages and use differently named option types to separate shared ones

src/Shared/AzureManaged/DurableTaskSchedulerOptions.cs
YunchuWang: as discussed, will split this in client/worker

samples/AspNetWebApp/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+using Azure.Identity;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Worker;
+using Microsoft.DurableTask.Worker.AzureManaged;
+using Microsoft.DurableTask.Client.AzureManaged;
+
+WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
+
+string endpointAddress = builder.Configuration["DURABLE_TASK_SCHEDULER_ENDPOINT_ADDRESS"]
```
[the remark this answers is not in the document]
YunchuWang: deleted

src/Worker/AzureManaged/DurableTaskSchedulerWorkerExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+{
+    /// <summary>
+    /// Configures Durable Task worker to use the Azure Durable Task Scheduler service.
+    /// </summary>
+    /// <param name="builder">The Durable Task worker builder to configure.</param>
+    /// <param name="endpointAddress">The endpoint address of the Durable Task Scheduler resource. Expected to be in the format "https://{scheduler-name}.{region}.durabletask.io".</param>
+    /// <param name="taskHubName">The name of the task hub resource associated with the Durable Task Scheduler resource.</param>
+    /// <param name="credential">The credential used to authenticate with the Durable Task Scheduler task hub resource.</param>
+    /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerOptions.</param>
+    public static void UseDurableTaskScheduler(
```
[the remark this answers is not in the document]
YunchuWang: deleted

samples/AspNetWebApp/Utils.cs
YunchuWang: deleted

samples/AspNetWebApp/appsettings.Production.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "DURABLE_TASK_SCHEDULER_ENDPOINT_ADDRESS": "https://{your-durable-task-endpoint}.durabletask.io",
+  "DURABLE_TASK_SCHEDULER_TASK_HUB_NAME": "{your-task-hub-name}",
+  "CONTAINER_APP_UMI_CLIENT_ID": "{your-user-managed-identity-client-id}"
```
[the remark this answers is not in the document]
YunchuWang: deleted

samples/AspNetWebApp/DockerFile
YunchuWang: deleted

samples/AspNetWebApp/ScenariosController.cs
YunchuWang: deleted

src/Worker/AzureManaged/DurableTaskSchedulerWorkerExtensions.cs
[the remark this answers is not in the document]
YunchuWang: added

src/Worker/AzureManaged/DurableTaskSchedulerWorkerExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    /// <param name="builder">The Durable Task worker builder to configure.</param>
+    /// <param name="connectionString">The connection string used to connect to the Durable Task Scheduler service.</param>
+    /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerOptions.</param>
+    public static void UseDurableTaskScheduler(
+        this IDurableTaskWorkerBuilder builder,
+        string connectionString,
+        Action<DurableTaskSchedulerOptions>? configure = null)
+    {
+        var connectionOptions = DurableTaskSchedulerOptions.FromConnectionString(connectionString);
```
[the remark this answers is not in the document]
YunchuWang: refactored

YunchuWang:
thanks, can i get another review？

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
```diff
@@ -0,0 +1,182 @@
...
+    public void UseDurableTaskScheduler_WithNullOrEmptyConnectionString_ShouldThrowArgumentException(string connectionString)
+    {
+        // Arrange
+        var services = new ServiceCollection();
+        var mockBuilder = new Mock<IDurableTaskClientBuilder>();
+        mockBuilder.Setup(b => b.Services).Returns(services);
+
+        // Act & Assert
+        var action = () => mockBuilder.Object.UseDurableTaskScheduler(connectionString);
+        action.Should().Throw<ArgumentException>();
```
cgillum, line 130: The correct exception to throw in this case is `ArgumentNullException`.

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientOptionsTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        // Arrange
+        var options = new DurableTaskSchedulerClientOptions
+        {
+            EndpointAddress = $"https://{ValidEndpoint}",
+            TaskHubName = ValidTaskHub,
+            Credential = new DefaultAzureCredential()
+        };
+
+        // Act
+        var channel = options.CreateChannel();
```
cgillum, line 143: Should we dispose this channel to avoid leaking resources?

cgillum (commented):
Just a few minor things.

.github/workflows/validate-build.yml
```diff
@@ -25,7 +25,12 @@ jobs:
       with:
         submodules: true
 
-    - name: Setup .NET
+    - name: Setup .NET 6.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '6.0.x'
```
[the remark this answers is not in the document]
YunchuWang: it was failing and asking to install dotnet 6. i cant find the run history now..

.editorconfig
```diff
@@ -232,3 +232,9 @@ dotnet_diagnostic.SA1400.severity = none  # Access modifier must be declared
 dotnet_diagnostic.SA1402.severity = none  # File may only contain a single type
 dotnet_diagnostic.SA1633.severity = warning  # File must have header -- TODO: replace with IDE0073 eventually
 dotnet_diagnostic.SA1649.severity = none  # File name must match type name
+
+# Enforce explicit types instead of `var`
+dotnet_style_prefer_var_for_built_in_types = false:error
```
jviau, line 237: Can we keep editorconfig changes to a separate PR?

src/Client/AzureManaged/Client.AzureManaged.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <EnableStyleCop>true</EnableStyleCop>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../Client/Grpc/Client.Grpc.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.13.1" />
+    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="8.0.0" />
```
jviau, line 15: We are targeting net6 but using a v8.x package. While this is technically 'valid', it may be confusing to users. Are we able to use a 6.x package here?

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -0,0 +1,158 @@
...
+    public static DurableTaskSchedulerClientOptions FromConnectionString(string connectionString)
+    {
+        return FromConnectionString(new DurableTaskSchedulerConnectionString(connectionString));
+    }
+
+    /// <summary>
+    /// Creates a gRPC channel for communicating with the Durable Task Scheduler service.
+    /// </summary>
+    /// <returns>A configured <see cref="GrpcChannel"/> instance that can be used to make gRPC calls.</returns>
+    public GrpcChannel CreateChannel()
```
jviau, line 60: This shouldn't be public.

```suggestion
    internal GrpcChannel CreateChannel()
```

If you need to access it for testing you can add InternalsVisibleTo like so:

https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/Abstractions.csproj#L22

src/Worker/AzureManaged/DurableTaskSchedulerWorkerOptions.cs
```diff
@@ -0,0 +1,165 @@
...
+using Azure.Identity;
+using Grpc.Core;
+using Grpc.Net.Client;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Options for configuring the Durable Task Scheduler.
+/// </summary>
+public class DurableTaskSchedulerWorkerOptions
```
jviau, line 14: To avoid code duplication, but also keep these types separate, I wonder if you can create an internal `DurableTaskSchedulerOptions` in the shared project, have all of the properties public and then inherit it here?

I am not 100% sure that will work, but worth a shot if you want. If not then it is fine to keep it duplicated.

src/Worker/AzureManaged/DurableTaskSchedulerWorkerOptions.cs
```diff
@@ -0,0 +1,165 @@
...
+    public static DurableTaskSchedulerWorkerOptions FromConnectionString(string connectionString)
+    {
+        return FromConnectionString(new DurableTaskSchedulerConnectionString(connectionString));
+    }
+
+    /// <summary>
+    /// Creates a gRPC channel for communicating with the Durable Task Scheduler service.
+    /// </summary>
+    /// <returns>A configured <see cref="GrpcChannel"/> instance that can be used to make gRPC calls.</returns>
+    public GrpcChannel CreateChannel()
```
jviau, line 66: If my previous comment about code deduplication does not work, then you could take a separate approach: define an `internal` interface with all the properties you need in the AzureManaged shared section, and then an internal static class to add these helpers as extension methods. 

But again, if you prefer not to mess around with ways to deduplicate the code that is fine with me also.

jviau, line 66:
```suggestion
    internal GrpcChannel CreateChannel()
```

src/Worker/AzureManaged/Worker.AzureManaged.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <EnableStyleCop>true</EnableStyleCop>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../../Worker/Grpc/Worker.Grpc.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.13.1" />
+    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="8.0.0" />
```
jviau, line 15: Same here - can we use a `6.x` package?

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        ServiceCollection services = new ServiceCollection();
+        Mock<IDurableTaskClientBuilder> mockBuilder = new Mock<IDurableTaskClientBuilder>();
+        mockBuilder.Setup(b => b.Services).Returns(services);
+        DefaultAzureCredential credential = new DefaultAzureCredential();
+
+        // Act
+        Action action = () => mockBuilder.Object.UseDurableTaskScheduler(endpoint, taskHub, credential);
+
+        // Assert
+        action.Should().NotThrow(); // The validation happens when building the service provider
```
jviau, line 83: The comment and action here do not align: the service provider is not built here and no options validation will have ran yet.

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        if (endpoint == null || taskHub == null)
+        {
+            ServiceProvider provider = services.BuildServiceProvider();
+            OptionsValidationException ex = Assert.Throws<OptionsValidationException>(() =>
+            {
+                DurableTaskSchedulerClientOptions options = provider.GetRequiredService<IOptions<DurableTaskSchedulerClientOptions>>().Value;
+            });
+            Assert.Contains(endpoint == null ? "EndpointAddress" : "TaskHubName", ex.Message);
```
jviau, line 92: Can you use fluent validation on the exception?

cgillum (approved):
All my main concerns are resolved. Thanks @YunchuWang! I agree with Jacob's recent round of feedback too, so let's address those (I'll leave it up to you to decide on whether to take on some of the code de-duplication, I don't have a strong opinion on this) and then I think we're good to merge this.

src/Client/AzureManaged/Client.AzureManaged.csproj
YunchuWang: {791DA31B-A16A-4053-8E00-379D46984313} v6 of this package does not target dotnet 6, and v7 does not have 
{16E38BDC-D45C-4567-ADB7-9851B06452D0} validateonstart. v8 is min version targeting dotnet 6 and has validateonstart api

src/Worker/AzureManaged/DurableTaskSchedulerWorkerOptions.cs
YunchuWang: discussed with Chris, plan to have this as a future todo

src/Worker/AzureManaged/Worker.AzureManaged.csproj
YunchuWang: ditto, this pkg major version are not aligned with dotnet target version.

src/Worker/AzureManaged/DurableTaskSchedulerWorkerOptions.cs
YunchuWang: i will leave it for now thanks Jacob

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
YunchuWang: do you refer to FluentValidation package?

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
YunchuWang: this is unneccesary will remove

test/Client/AzureManaged.Tests/DurableTaskSchedulerClientExtensionsTests.cs
YunchuWang: updated to validate exact exception message

src/Client/AzureManaged/Client.AzureManaged.csproj
jviau: Do you need ValidateOnStart? This will effectively be validated on start, as validation will occur on first import of it, which will be done at start due to this being eventually imported by a `IHostedService` (the DurableTaskWorker)

YunchuWang: good point, updated ver and removed validateonstart

src/Client/AzureManaged/Client.AzureManaged.csproj
```diff
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.13.1" />
-    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="7.0.0" />
```
cgillum, line 15: This needs to be `6.x`
```suggestion
    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="6.0.1" />
```

src/Worker/AzureManaged/Worker.AzureManaged.csproj
```diff
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.13.1" />
-    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="7.0.0" />
```
cgillum, line 15:
```suggestion
    <PackageReference Include="Microsoft.Extensions.Options.DataAnnotations" Version="6.0.1" />
```

cgillum (approved):

PR #366, opened 2025-01-24
Add new create builder overloads without callbacks

jviau:
This PR introduces new create builder overloads for the client and worker with a slight difference. The existing overloads all take in a callback to configure the builder. The new overloads return the builder. This allows for a slightly cleaner `HostApplicationBuilder` experience. Some samples have also been updated to use `HostApplicationBuilder` and the new overloads.

Current style:

``` CSharp
services.AddDurableTaskWorker(b =>
{ 
    b.Configure(...);
    b.AddTasks(...);
    b.UseGrpc();
});
```

New style:

``` CSharp
services.AddDurableTaskWorker()
    .Configure(...)
    .AddTasks(...)
    .UseGrpc();
```


cgillum (approved):
LGTM. Glad to have this simplified setup experience. Thanks!

PR #367, opened 2025-01-25
Fix method ordering, xmldoc, and remove obsolete XUnit verifiers

jviau:
Fixing some warnings to get back to 0 warnings

- Re-orders a couple methods
- Fixes xmldoc
- Removes obsolete XUnit analysis packages, following guidance on what to replace with.
  - this exposed an anti-pattern in one of the analyzers, add a TODO comment regarding it.


src/Analyzers/Analyzers.csproj
```diff
@@ -9,7 +9,7 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
     <IncludeShared>false</IncludeShared>
-    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
cgillum, line 12: Why do we want to generate the documentation file for the analyzer? I wouldn't expect that we'd want to since we wouldn't normally expect developers to program against the analyzer's APIs, would we?

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
```diff
@@ -367,7 +367,11 @@ static DateTime MyRunAsync(TaskOrchestrationContext context)
 
         DiagnosticResult expected = BuildDiagnostic().WithLocation(0).WithArguments("MyRunAsync", "System.DateTime.Now", "MyRun");
 
-        await VerifyCS.VerifyDurableTaskCodeFixAsync(code, expected, fix);
+        await VerifyCS.VerifyDurableTaskCodeFixAsync(code, expected, fix, test =>
+        {
+            // TODO: CodeFixBehaviors.SkipLocalDiagnosticCheck is failing this test. This is apparently a bad analyzer practice.
+            test.CodeFixTestBehaviors |= CodeFixTestBehaviors.SkipLocalDiagnosticCheck;
```
cgillum, line 373: I'm not sure I understand what this comment is saying. It seems to be saying that `SkipLocalDiagnosticCheck` is a bad practice and is causing failures, but the line of code directly under the comment seems to be *enabling* it, which seems contradictory.

src/Analyzers/Analyzers.csproj
jviau: StyleCop was producing a warning without it. Figured it was easier to just generate it then dig into that warning.

test/Analyzers.Tests/Orchestration/DateTimeOrchestrationAnalyzerTests.cs
jviau: I'll clarify the comment. By default this will throw for any non-local diagnostic checks (what that means I am unsure). So we tell it to skip that check until we can address it. My understanding this fails by default to point out the bad practice to analyzer developers.

src/Analyzers/Analyzers.csproj
```diff
@@ -2,30 +2,22 @@
...
-  <PropertyGroup>
-    <PackageDescription>.NET Analyzers for the Durable Task Framework.</PackageDescription>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <NoWarn>$(NoWarn);SA0001;</NoWarn>
   </PropertyGroup>
 
   <PropertyGroup>
     <VersionPrefix>0.0.1</VersionPrefix>
     <VersionSuffix>preview.1</VersionSuffix>
+    <PackageDescription>.NET Analyzers for the Durable Task Framework.</PackageDescription>
```
cgillum, line 16:
```suggestion
    <PackageDescription>.NET Analyzers for the Durable Task SDK.</PackageDescription>
```

src/Analyzers/Analyzers.csproj
```diff
@@ -2,30 +2,22 @@
...
     <NeutralLanguage>en</NeutralLanguage>
+    <IncludeBuildOutput>false</IncludeBuildOutput> <!-- Do not include the analyzer as a lib dependency -->
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+    <IsPackable>false</IsPackable> <!-- Do not publish it until it is ready to release -->
```
cgillum, line 20: I see that you're just reordering this, but why do we set this to `false`? I assume we only generate the package anyways when we do `dotnet pack` and I'm wondering why we'd want to block someone from using `dotnet pack`. Let me know if I'm misunderstanding.

src/Analyzers/Analyzers.csproj
```diff
@@ -2,30 +2,22 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeShared>false</IncludeShared>
```
cgillum, line 5: What does `<IncludeShared>` do?

cgillum (approved):
PR looks good. A couple questions and one minor suggestion on the package description.

src/Analyzers/Analyzers.csproj
jviau: Our CI pipeline runs `dotnet pack` on the solution file. To avoid this from being included in that step and being unintentionally released, we disabled packing. If someone wants to pack locally, they can override this behavior with `-p:IsPackable=true`

src/Analyzers/Analyzers.csproj
jviau: Suppresses the logic we have to including files from `Shared.csproj`

PR #368, opened 2025-01-26
feat: Schedule Support

YunchuWang:
Schedule Support for Durable Task Framework
Overview
This PR introduces comprehensive scheduling capabilities to the Durable Task Framework, enabling users to create and manage recurring task executions with fine-grained control over timing and execution parameters.
Key Features
1. Schedule Management
Create, pause, resume, update, and delete scheduled tasks
Flexible schedule configuration with support for:
Custom intervals (minimum 1 second)
Start and end times
Immediate or delayed first execution
Custom orchestration instance IDs
Orchestration input parameters
2. Schedule Configuration
ScheduleCreationOptions for initial setup with required parameters:
scheduleId: Unique identifier for the schedule
orchestrationName: Target orchestration to execute
interval: Time between executions
Optional parameters:
startAt: Custom start time
endAt: Schedule termination time
orchestrationInput: Input data for the orchestration
orchestrationInstanceId: Custom instance ID
startImmediatelyIfLate: Controls immediate execution behavior
3. Schedule State Management
Robust state transitions with validation
Schedule status tracking (Active, Paused, etc.)
Execution history tracking:
Last run time
Next scheduled run time
Execution token for operation validation
4. Client Interface
Clean and intuitive client API through ScheduledTaskClient
Async operations for all schedule management functions
Comprehensive error handling and logging
Support for cancellation tokens
5. Worker Integration
Simple worker configuration through extension methods
Automatic registration of required components:
Schedule entity
Schedule operation orchestrator


src/ScheduledTasks/Models/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,190 @@
+// Copyright (c) Microsoft Corporation.
```
YunchuWang, line 1: TODO: scheduleconfiguration, schedulecreationoptions, scheduleupdateoptions are pojos based on same schema, need to refactor to remove dup codes

src/ScheduledTasks/Client/ScheduleHandle.cs
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Client.Entities;
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.ScheduledTasks;
```
YunchuWang, line 9: TODO: need to provide customer operation polling functions or wrap function to poll for resource(schedule) reaching target status functions or both. either way need to give customer failure details if it failed. still testing to find out what returned from get entityinstance and how it captures the operation failures to decide

src/ScheduledTasks/Client/IScheduleHandle.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <returns>A task that completes when the schedule is updated.</returns>
+    Task UpdateAsync(ScheduleUpdateOptions updateOptions);
+
+    /// <summary>
+    /// Gets the details of the schedule's underlying orchestration instance.
+    /// </summary>
+    /// <param name="getInputsAndOutputs">If true, includes the serialized inputs and outputs in the returned metadata.</param>
+    /// <param name="cancellation">Optional cancellation token.</param>
+    /// <returns>The orchestration metadata for the schedule instance, or null if not found.</returns>
+    Task<OrchestrationMetadata?> GetScheduleInstanceDetailsAsync(
```
YunchuWang, line 55: TODO: Can ignore this function, for testing only, will refactor or remove this later

Microsoft.DurableTask.sln
```diff
@@ -83,8 +83,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shared", "Shared", "{CECADD
 EndProject
```
YunchuWang, line 83: TODO: adding tests

YunchuWang:
TODO: Figure out why logging in entity operation logs multiple times, i debugged it only got called once
TODO: Waiter interface should need some refactoring
TODO: All logs are on client side, need to determine how to best present and collect them
TODO: Manually tested happy and a few failure cases look good, but need to test more. add test cases

YunchuWang:
TODO: enhance sample or add other samples

YunchuWang:
TODO: code comments need to be cleaned up or revisited

YunchuWang:
Scheduling with Startat and interval are supported in this pr, other options are placeholders for next phase. I can remove/comment them out before releasing phase 0.

src/ScheduledTasks/Exception/ScheduleOperationFailedException.cs
```diff
@@ -0,0 +1,53 @@
...
+
+/// <summary>
+/// Exception thrown when a schedule operation fails.
+/// </summary>
+public class ScheduleOperationFailedException : Exception
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ScheduleOperationFailedException"/> class.
+    /// </summary>
+    /// <param name="schedule">The schedule that failed.</param>
```
YunchuWang, line 14: can anyone advise whats event id ranges available to use for these loggings?

src/ScheduledTasks/Logging/Entity/Logs.cs
```diff
@@ -0,0 +1,63 @@
...
+
+/// <summary>
+/// Log messages.
+/// </summary>
+/// <remarks>
+/// NOTE: Trying to make logs consistent with https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Logging/LogEvents.cs.
+/// </remarks>
+static partial class Logs
+{
+    [LoggerMessage(EventId = 1, Level = LogLevel.Information, Message = "Schedule '{scheduleId}' is being created")]
```
YunchuWang, line 16: can anyone advise whats event id ranges available to use for these loggings?

YunchuWang: i use eventids outside durable eventids

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+            ?? throw new InvalidOperationException("Missing required environment variable 'DURABLE_TASK_SCHEDULER_CONNECTION_STRING'");
+
+        // Configure the worker
+        _ = services.AddDurableTaskWorker(builder =>
+        {
+            // Add the Schedule entity and demo orchestration
+            builder.AddTasks(r =>
+            {
+                // Add a demo orchestration that will be triggered by the schedule
+                r.AddOrchestratorFunc<string, string>("DemoOrchestration", async (context, symbol) =>
```
cgillum, line 26: I suggest you create this orchestration as a class rather than a lambda. That will make the sample much easier to read.

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+        await handle.DeleteAsync();
+        Console.WriteLine($"Deleted schedule {schedule.ScheduleId}");
+    }
+
+    // Create schedule options that runs every 30 seconds
+    ScheduleCreationOptions scheduleOptions = new ScheduleCreationOptions
+    {
+        OrchestrationName = "DemoOrchestration",
+        ScheduleId = "demo-schedule101",
+        Interval = TimeSpan.FromSeconds(4),
```
cgillum, line 99: The comment above says 30 seoncds but the interval here says 4 seconds.

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+        StartAt = DateTimeOffset.UtcNow,
+        OrchestrationInput = "MSFT"
+    };
+
+    // Get schedule handle
+    IScheduleHandle scheduleHandle = scheduledTaskClient.GetScheduleHandle(scheduleOptions.ScheduleId);
```
cgillum, line 105: Why do we get the handle before creating?

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+
+    // Create the schedule
+    Console.WriteLine("Creating schedule...");
+    IScheduleWaiter waiter = await scheduleHandle.CreateAsync(scheduleOptions);
+    ScheduleDescription scheduleDescription = await waiter.WaitUntilActiveAsync();
```
cgillum, line 110: What does active mean in the context of a schedule, and why do we need to wait for it to be active?

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+    Console.WriteLine(scheduleDescription.ToJsonString(true));
+
+    Console.WriteLine("");
+    Console.WriteLine("");
+    Console.WriteLine("");
+
+    // Pause the schedule
+    Console.WriteLine("\nPausing schedule...");
+    IScheduleWaiter pauseWaiter = await scheduleHandle.PauseAsync();
+    scheduleDescription = await pauseWaiter.WaitUntilPausedAsync();
```
cgillum, line 121: Same question: why do we need to wait for the state change? If it's truly necessary, would it make sense to have `PauseAsync()` do the waiting?

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+    Console.WriteLine("");
+    Console.WriteLine("");
+    Console.WriteLine("");
+
+
+    // Resume the schedule
+    Console.WriteLine("\nResuming schedule...");
+    IScheduleWaiter resumeWaiter = await scheduleHandle.ResumeAsync();
+
+    scheduleDescription = await resumeWaiter.WaitUntilActiveAsync();
```
cgillum, line 132: Same question about waiting for state changes.

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+    Console.WriteLine(scheduleDescription.ToJsonString(true));
+
+    Console.WriteLine("");
+    Console.WriteLine("");
+    Console.WriteLine("");
+
+    await Task.Delay(2000000);
```
cgillum, line 139: nit: Please use `TimeSpan.FromSeconds(...)` or `TimeSpan.FromMinutes(...)`

samples/ScheduleDemo/Program.cs
```diff
@@ -0,0 +1,153 @@
...
+    {
+        var handle = scheduledTaskClient.GetScheduleHandle(schedule.ScheduleId);
+        await handle.DeleteAsync();
+        Console.WriteLine($"Deleted schedule {schedule.ScheduleId}");
+    }
+
+    // Create schedule options that runs every 30 seconds
+    ScheduleCreationOptions scheduleOptions = new ScheduleCreationOptions
+    {
+        OrchestrationName = "DemoOrchestration",
```
cgillum, line 97: Make the orchestration name a required argument instead of a property that users can set since it is always needed. That will make the API more discoverable.

cgillum (commented):
First round of comments. I've only reviewed the sample code so that I can focus on the customer experience. I'll help review the implementation after we agree on the sample code.

samples/ScheduleDemo/Program.cs
YunchuWang: it was a limitation using waiter to poll. originally createscheduleasync was in scheduletasksclient which should return a schedulehandle from createschedule, but it need to poll until schedule is created using waiter while waiter needs the schedulehandle to do polling. 

will update based on using polling free approach

samples/ScheduleDemo/Program.cs
YunchuWang: i define couple of resource status for schedule : active, paused, unintialized. and poll for schedule reaching target status expected for different schedule operations.  with poll free approach, this wont be necceesary

samples/ScheduleDemo/Program.cs
YunchuWang: i was doing that, @torosent mentioned to use paramless ctor and obj init to be more flexible. i think it make sense to highlight the required param.

samples/ScheduleDemo/Program.cs
YunchuWang: updated

samples/ScheduleDemo/Program.cs
YunchuWang: updated

samples/ScheduleDemo/Program.cs
YunchuWang: updated

samples/ScheduleDemo/Program.cs
YunchuWang: updated

samples/ScheduleDemo/Program.cs
YunchuWang: updated

YunchuWang:
https://medium.com/p/cee63463216f added a post to summarize my finding of LRO using polling for resource status vs polling for operation completion status approaches. In schedule case, with poll free callentity support, we can easily retrieve each entity operation execution status without additional complexity. Decoupling op status from resource status is more clear and extensible approach in this case

src/ScheduledTasks/ScheduledTasks.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <PackageDescription>Durable Task Scheduled Tasks Client</PackageDescription>
+    <EnableStyleCop>true</EnableStyleCop>
+    <VersionSuffix>preview.1</VersionSuffix>
```
cgillum, line 7: Do we need any pipeline updates to build the signed nugets for this new project?

YunchuWang: will take a look

src/ScheduledTasks/Client/ScheduledTaskClient.cs
```diff
@@ -0,0 +1,266 @@
...
+            // Check if schedule already exists
+            bool scheduleExists = await this.CheckScheduleExists(scheduleId, cancellation);
+            if (scheduleExists)
+            {
+                throw new ScheduleAlreadyExistException(scheduleId);
+            }
+
+            // Call the orchestrator to create the schedule
+            ScheduleOperationRequest request = new ScheduleOperationRequest(entityId, nameof(Schedule.CreateSchedule), creationOptions);
+            string instanceId = await this.durableTaskClient.ScheduleNewOrchestrationInstanceAsync(
```
cgillum, line 50: I was expecting schedules to be implemented using simple entities, but I see you're creating an orchestration. Why is the orchestration needed?

cgillum (commented):
Some initial feedback

jviau (changes requested):
Going to need some time to look at the core schedule logic.

src/ScheduledTasks/Client/ScheduledTaskClient.cs
YunchuWang: It serves as an indirect method for achieving a poll-free signal operation call, as recommended by @sebastianburckhardt. Within the orchestration entity context, it can utilize the CallEntity API, which not only eliminates polling for the consumer but also manages tracking, retrieval, and decoupling of feedback. This ensures that the operation status, including the result or failure details, is captured at the entity operation level, rather than being embedded within the entity state, as was done in my old polling-based approach.

src/ScheduledTasks/ScheduledTasks.csproj
YunchuWang: checked they are included in sln, thanks

src/Grpc/versions.txt
```diff
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch main at 2025-02-11 19:40:58 UTC
+# The following files were downloaded from branch main at 2025-02-19 06:25:02 UTC
```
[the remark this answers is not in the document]
YunchuWang: https://github.com/microsoft/durabletask-protobuf/commit/589cb5ecd9dd4b1fe463750defa3e2c84276b079. the durable entity commit, i think the timestamp is auto updated to latest downloaded timestamp when i run refresh-proto.ps1 even commit already downloaded and tracked in this file before. 
{F9648555-5F3E-49F9-91A5-C1C85670BBE7}

src/ScheduledTasks/Extension/DurableTaskClientBuilderExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+public static class DurableTaskClientBuilderExtensions
+{
+    /// <summary>
+    /// Enables scheduled task support for the client builder.
+    /// </summary>
+    /// <param name="builder">The client builder to add scheduled task support to.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskClientBuilder EnableScheduledTasksSupport(this IDurableTaskClientBuilder builder)
+    {
+        builder.Services.AddTransient<IScheduledTaskClient>(sp =>
```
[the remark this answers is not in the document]
YunchuWang: had checked with Chris before, extending durabletaskclient to get scheduledtaskclient does not have access to di container to instantiate scheduledtasksclient, thus i have to pass ctor arguments like logger when calling extension function which customers can feel strange? for example 


namespace Microsoft.DurableTask.ScheduledTasks
{
    public static class DurableTaskClientExtensions
    {
        public static ScheduledTaskClient ScheduledTasks(this DurableTaskClient client, ILogger logger)
        {
            return new ScheduledTaskClient(client, logger);
        }
    }
}

src/Client/Core/DurableTaskClient.cs
```diff
@@ -408,4 +409,35 @@ public virtual Task<PurgeResult> PurgeAllInstancesAsync(
     /// </summary>
     /// <returns>A <see cref="ValueTask"/> that completes when the disposal completes.</returns>
     public abstract ValueTask DisposeAsync();
+
+    public virtual Task<ScheduleState> GetScheduleAsync(string scheduleId)
```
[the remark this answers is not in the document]
YunchuWang: discussed with Chris on this before, extending durabletaskclient to get scheduledtaskclient does not have access to di container to instantiate scheduledtasksclient, thus i have to pass ctor arguments like logger when calling extension function which customers can feel strange? 
image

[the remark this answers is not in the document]
YunchuWang: that was added when pr not ready for review, outdated

samples/ScheduleDemo/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    public string? CronExpression { get; set; }
+
+    public int MaxOccurrence { get; set; }
+
+    public bool? StartImmediatelyIfLate { get; set; }
+
+    internal int Version { get; set; } // Tracking schedule config version
```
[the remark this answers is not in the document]
YunchuWang: that was added when pr not ready for review, outdated

src/ScheduledTasks/Models/ScheduleDescription.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// Gets the next scheduled run time.
+    /// </summary>
+    public DateTimeOffset? NextRunAt { get; init; }
+
+    /// <summary>
+    /// Returns a JSON string representation of the schedule description.
+    /// </summary>
+    /// <param name="pretty">If true, formats the JSON with indentation for readability.</param>
+    /// <returns>A JSON string containing the schedule details.</returns>
+    public string ToJsonString(bool pretty = false)
```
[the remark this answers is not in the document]
YunchuWang: updated

src/ScheduledTasks/ScheduledTasks.csproj
```diff
@@ -0,0 +1,20 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="../Abstractions/Abstractions.csproj" />
+    <ProjectReference Include="../Client/Core/Client.csproj" />
+    <ProjectReference Include="..\Worker\Core\Worker.csproj" />
```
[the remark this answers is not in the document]
YunchuWang: updated

samples/ScheduleWebApp/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+// Add all the generated orchestrations and activities automatically
+builder.Services.AddDurableTaskWorker(builder =>
+{
+    builder.AddTasks(r =>
+    {
+        // Add your orchestrators and activities here
+        r.AddOrchestrator<CacheClearingOrchestrator>();
+    });
+    builder.UseDurableTaskScheduler(connectionString);
+    builder.EnableScheduledTasksSupport();
```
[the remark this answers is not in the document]
YunchuWang: updated to builder.UseScheduledTasks();

jviau:
I'm not sure I agree with those codeql alerts. It will impact logging performance. Also, this is not the equivalent of an HTTP call at all - hard to even call those areas "user" input.

samples/ScheduleDemo/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+class ScheduleConfiguration
+{
+    public ScheduleConfiguration(string orchestrationName, string scheduleId)
+    {
+        this.OrchestrationName = orchestrationName;
+        this.ScheduleId = scheduleId ?? Guid.NewGuid().ToString("N");
```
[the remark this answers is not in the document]
YunchuWang: this file was in old pr not ready for review yet, outdated

samples/ScheduleDemo/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    public string OrchestrationName { get; set; }
+
+    public string ScheduleId { get; set; }
+
+    public string? OrchestrationInput { get; set; }
+
+    public string? OrchestrationInstanceId { get; set; } = Guid.NewGuid().ToString("N");
+
+    public DateTimeOffset? StartAt { get; set; }
```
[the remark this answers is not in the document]
YunchuWang: this file was in old pr not ready for review yet, outdated

samples/ScheduleDemo/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (value.Value.TotalSeconds < 1)
+            {
+                throw new ArgumentException("Interval must be at least 1 second", nameof(value));
+            }
+
+            this.interval = value;
+        }
+    }
+
+    public string? CronExpression { get; set; }
```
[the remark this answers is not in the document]
YunchuWang: this file was in old pr not ready for review yet, outdated

samples/ScheduleDemo/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    public int MaxOccurrence { get; set; }
+
+    public bool? StartImmediatelyIfLate { get; set; }
```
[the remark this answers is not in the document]
YunchuWang: this file was in old pr not ready for review yet, outdated

src/ScheduledTasks/Extension/DurableTaskClientBuilderExtensions.cs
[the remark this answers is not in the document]
jviau: We will need some way to associate a schedule client with a given `DurableTaskClient` - the single registration will not work.

We could add an `IFeatureCollection` AspNetCore style to `DurableTaskClient`, which allows for open ended attaching of new objects ('features') to a client. But probably doesn't need to be as robust as what AspNetCore is. Some general property bag will work

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,312 @@
...
+        {
+            StartOrchestrationOptions? startOrchestrationOptions = string.IsNullOrEmpty(this.State.ScheduleConfiguration?.OrchestrationInstanceId)
+                ? null
+                : new StartOrchestrationOptions(this.State.ScheduleConfiguration.OrchestrationInstanceId);
+            context.ScheduleNewOrchestration(
+                new TaskName(this.State.ScheduleConfiguration!.OrchestrationName),
+                this.State.ScheduleConfiguration.OrchestrationInput,
+                startOrchestrationOptions);
+        }
+        catch (Exception ex)
```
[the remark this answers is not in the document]
YunchuWang: rethrowing will disrupt the runschedule operation to call recursively, and stopping the schedule and it is not directly customer facing but internal operation. i think it is better to log it and customers can check orchestration run history/logs to troubleshoot. can be an option for customer to specify whether they want to stop schedule on orch failure later

src/ScheduledTasks/Client/ScheduleHandle.cs
```diff
@@ -0,0 +1,235 @@
...
+            };
+        }
+        catch (OperationCanceledException ex)
+        {
+            this.logger.ClientError(
+                nameof(this.DescribeAsync),
+                this.ScheduleId,
+                ex);
+
+            throw new OperationCanceledException(
```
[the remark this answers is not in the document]
YunchuWang: i was thinking this exception is client facing, the exception thrown from getentity would be get entity timed out which can confuse customers?  we can expose get schedule time out for more clarity to customers?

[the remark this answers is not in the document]
YunchuWang: you are right, should set innerexception, updated.

src/ScheduledTasks/Client/ScheduledTaskClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        EntityQuery query = new EntityQuery
+        {
+            InstanceIdStartsWith = filter?.ScheduleIdPrefix ?? nameof(Schedule),
+            IncludeState = false, // We don't need the state since we only want IDs
+            PageSize = filter?.PageSize ?? ScheduleQuery.DefaultPageSize,
+            ContinuationToken = filter?.ContinuationToken,
+        };
+
+        // Create an async pageable using the Pageable.Create helper
+        return Task.FromResult(Pageable.Create(async (continuationToken, pageSize, cancellation) =>
```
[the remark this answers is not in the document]
YunchuWang: removed this one as Chris suggested not really needed

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <param name="cancellation">Optional cancellation token.</param>
+    /// <returns>The schedule description if found, null otherwise.</returns>
+    Task<ScheduleDescription?> GetScheduleAsync(string scheduleId, CancellationToken cancellation = default);
+
+    /// <summary>
+    /// Gets a pageable list of schedules matching the specified filter criteria.
+    /// </summary>
+    /// <param name="filter">Optional filter criteria for the schedules. If null, returns all schedules.</param>
+    /// <returns>A pageable list of schedule descriptions.</returns>
+    Task<AsyncPageable<ScheduleDescription>> ListSchedulesAsync(ScheduleQuery? filter = null);
```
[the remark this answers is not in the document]
YunchuWang: i think without cancellation is consistent with other durable interfaces like durabletaskclient and as Jacob mentioned customer can still cancel it.

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Gets a pageable list of schedule IDs matching the specified filter criteria.
+    /// This is a more efficient version of ListSchedulesAsync when only the IDs are needed.
+    /// </summary>
+    /// <param name="filter">Optional filter criteria for the schedules. If null, returns all schedule IDs.</param>
+    /// <returns>A pageable list of schedule IDs.</returns>
+    Task<AsyncPageable<string>> ListScheduleIdsAsync(ScheduleQuery? filter = null);
```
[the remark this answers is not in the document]
YunchuWang: makse sense, will remove

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Interface for managing scheduled tasks in a Durable Task application.
+/// </summary>
+public interface IScheduledTaskClient
```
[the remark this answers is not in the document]
YunchuWang: any suggestions for abstract class name and subclass for scheduletaskclient？i am thinking scheduletaskbase and scheduletaskclient.

samples/ScheduleConsoleApp/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    // Create schedule options that runs every 4 seconds
+    ScheduleCreationOptions scheduleOptions = new ScheduleCreationOptions("demo-schedule101", nameof(StockPriceOrchestrator), TimeSpan.FromSeconds(4))
+    {
+        StartAt = DateTimeOffset.UtcNow,
+        OrchestrationInput = "MSFT"
+    };
+
+    // Create the schedule and get a handle to it
+    ScheduleHandle scheduleHandle = await scheduledTaskClient.CreateScheduleAsync(scheduleOptions);
```
sebastianburckhardt: I tried to run this sample locally but it throws an exception on this line:

**System.InvalidOperationException**: 'Failed to create schedule 'demo-schedule101': No orchestrator task named 'ExecuteScheduleOperationOrchestrator' was found.'

samples/ScheduleConsoleApp/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+    await scheduleHandle.ResumeAsync();
+    scheduleDescription = await scheduleHandle.DescribeAsync();
+    Console.WriteLine(scheduleDescription.ToJsonString(true));
+
+    Console.WriteLine("");
+    Console.WriteLine("");
+    Console.WriteLine("");
+
+    await Task.Delay(TimeSpan.FromMinutes(30));
+    //Console.WriteLine("\nPress any key to delete the schedule and exit...");
```
sebastianburckhardt: I would remove the commented-out code.

src/ScheduledTasks/Models/ScheduleTransitions.cs
```diff
@@ -0,0 +1,62 @@
...
+        };
+
+    /// <summary>
+    /// Attempts to get the valid target states for a given schedule state and operation.
+    /// </summary>
+    /// <param name="operationName">The name of the operation being performed.</param>
+    /// <param name="from">The current schedule state.</param>
+    /// <param name="validTargetStates">When this method returns, contains the valid target states if found; otherwise, an empty set.</param>
+    /// <returns>True if valid transitions exist for the given state and operation; otherwise, false.</returns>
+    public static bool TryGetValidTransitions(string operationName, ScheduleStatus from, out HashSet<ScheduleStatus> validTargetStates)
```
sebastianburckhardt, line 47: Creating a set, and then testing membership in that set, seems a bit more complicated than needed (also incurs an unnecessary allocation).

A more direct way to write this test would be to include the target state that is to be checked:

```csharp
public static bool IsValidTransition(string operationName, ScheduleStatus from, ScheduleStatus targetState)
```

and use a nested switch statement where the innermost clauses `return true` or `return false`

sebastianburckhardt (commented):
I did not get very far with my review yet but I figured I submit the comments I have so far. Will resume tomorrow.

src/ScheduledTasks/Client/ScheduledTaskClient.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            try
+            {
+                List<ScheduleDescription> schedules = new List<ScheduleDescription>();
+
+                await foreach (EntityMetadata<ScheduleState> metadata in this.durableTaskClient.Entities.GetAllEntitiesAsync<ScheduleState>(query))
```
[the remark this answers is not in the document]
YunchuWang: updated

samples/ScheduleConsoleApp/Program.cs
YunchuWang: thx, already removed

src/Abstractions/Entities/Schedule/Schedule.cs
```diff
@@ -0,0 +1,388 @@
...
+
+    public string? OrchestrationInstanceId { get; set; } = Guid.NewGuid().ToString("N");
+
+    public DateTimeOffset? StartAt { get; set; }
+
+    public DateTimeOffset? EndAt { get; set; }
+
+    TimeSpan? interval;
+
+    public TimeSpan? Interval
```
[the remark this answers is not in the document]
YunchuWang: it is relative to start time

src/ScheduledTasks/Client/ScheduledTaskClient.cs
```diff
@@ -0,0 +1,266 @@
...
+    {
+        Check.NotNull(creationOptions, nameof(creationOptions));
+        this.logger.ClientCreatingSchedule(creationOptions);
+
+        try
+        {
+            string scheduleId = creationOptions.ScheduleId;
+            EntityInstanceId entityId = new EntityInstanceId(nameof(Schedule), scheduleId);
+
+            // Check if schedule already exists
```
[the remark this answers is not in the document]
YunchuWang: it is not a concurrent safe check, will remove. it wont check duplicate schedules as we validate again on server side. i agree createorupdate provides a better dev experience. will update.

src/Abstractions/Entities/Schedule/Schedule.cs
```diff
@@ -0,0 +1,388 @@
...
+        {
+            // no need to interrupt and update current schedule run as there is no change in the schedule config
+            this.logger.LogInformation("Schedule configuration is up to date.");
+            return;
+        }
+
+        // after schedule config is updated, perform post-config-update logic separately
+        foreach (string updatedScheduleConfigField in updatedScheduleConfigFields)
+        {
+            switch (updatedScheduleConfigField)
```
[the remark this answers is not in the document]
YunchuWang: yes, this commit is outdated. it is in later commit.

src/ScheduledTasks/Models/ScheduleCreationOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    public DateTimeOffset? EndAt { get; init; }
+
+    /// <summary>
+    /// Gets the interval of the schedule.
+    /// </summary>
+    public TimeSpan Interval { get; }
+
+    /// <summary>
+    /// Gets a value indicating whether to start the schedule immediately if it is late. Default is false.
```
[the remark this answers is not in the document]
YunchuWang: yes, thats it refer to. updated the comment

src/ScheduledTasks/Models/ScheduleCreationOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+
+/// <summary>
+/// Configuration for a scheduled task.
+/// </summary>
+public record ScheduleCreationOptions
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ScheduleCreationOptions"/> class.
+    /// </summary>
+    /// <param name="scheduleId">The ID of the schedule, or null to generate one.</param>
```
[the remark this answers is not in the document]
YunchuWang: updated

src/ScheduledTasks/Models/ScheduleCreationOptions.cs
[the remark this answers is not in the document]
YunchuWang: updated

src/ScheduledTasks/Models/ScheduleCreationOptions.cs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// <summary>
+    /// Gets the input to the orchestration function.
+    /// </summary>
+    public string? OrchestrationInput { get; init; }
+
+    /// <summary>
+    /// Gets the instance ID of the orchestration function.
+    /// </summary>
+    public string? OrchestrationInstanceId { get; init; }
```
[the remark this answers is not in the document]
YunchuWang: customer can specify orch instance id to schedule and this make the orchestration scheduling idempotent and dedup overlapped runs. customer can leave this empty which is default and we will auto gen new instance id per run for them.

samples/ScheduleConsoleApp/Program.cs
YunchuWang: tested on latest pr, it is working. can you give another try?

src/Abstractions/Entities/Schedule/Schedule.cs
```diff
@@ -0,0 +1,388 @@
...
+    TimeSpan? interval;
+
+    public TimeSpan? Interval
+    {
+        get => this.interval;
+        set
+        {
+            if (!value.HasValue)
+            {
+                return;
```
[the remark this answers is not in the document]
YunchuWang: this is outdated, for phase 0 we support interval only and it is a required field. for phase 1, when adding cronexpression support, we can add a overloaded constructor taking cronexpression only and assign a default value to interval to avoid breaking change. as discussed with Tomer, cronexpression will take precedence over interval if both specified.

src/Abstractions/Entities/Schedule/Schedule.cs
```diff
@@ -0,0 +1,388 @@
...
+    // To stop potential runSchedule operation scheduled after the schedule update/pause, invalidate the execution token and let it exit gracefully
+    // This could incur little overhead as ideally the runSchedule with old token should be killed immediately
+    // but there is no support to cancel pending entity operations currently, can be a todo item
+    public void RefreshScheduleRunExecutionToken()
+    {
+        this.ExecutionToken = Guid.NewGuid().ToString("N");
+    }
+}
+
+public class ScheduleConfiguration
```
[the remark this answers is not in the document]
YunchuWang: this is a great point, for phase 0, due to time constraint, we would not consider this at this moment. checked temporal has this support as calendar based scheduling as well in later of their sdks.

samples/ScheduleDemo/ScheduleConfiguration.cs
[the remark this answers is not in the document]
YunchuWang: cron cant dictate start/end time i believe, they both need startat/endat. grouping into separate interfaces not bring much benefit here currently.

src/ScheduledTasks/Exception/ScheduleAlreadyExistException.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Exception thrown when attempting to create a schedule with an ID that already exists.
+/// </summary>
+public class ScheduleAlreadyExistException : InvalidOperationException
```
sebastianburckhardt, line 9: nit: I would call this `ScheduleAlreadyExistsException`.

src/ScheduledTasks/Client/IScheduleHandle.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Represents a handle to a schedule, allowing operations on a specific schedule instance.
+/// </summary>
+public interface IScheduleHandle
```
sebastianburckhardt, line 9: It may be better to use an abstract class than an interface, so that we can add new methods or overloads in the future as a non-breaking change.

Also, with regards to the name: to me the term 'Handle' suggests a local resource, but here, the resource is inside a remote service. In the Azure Storage SDK, such "handles for remote resources" are usually called clients (e.g. `BlobClient`, `BlobContainerClient`). So this could be called `ScheduleClient`, following that logic. I don't have a strong opinion on that though.

src/ScheduledTasks/Client/IScheduledTaskClient.cs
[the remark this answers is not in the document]
sebastianburckhardt: I think we usually omit the 'Base' suffix from the base class since this is the class our users work with, so we want their code to look as nice as possible (and have the more messy names be internal implementations).

See for example how this is done for `DurableTaskClient`.

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <returns>A pageable list of schedule descriptions.</returns>
+    AsyncPageable<ScheduleDescription> ListSchedulesAsync(ScheduleQuery? filter = null);
+
+    /// <summary>
+    /// Creates a new schedule with the specified configuration.
+    /// </summary>
+    /// <param name="creationOptions">The options for creating the schedule.</param>
+    /// <param name="cancellation">Optional cancellation token.</param>
+    /// <returns>A handle to the created schedule.</returns>
+    Task<IScheduleHandle> CreateScheduleAsync(ScheduleCreationOptions creationOptions, CancellationToken cancellation = default);
```
sebastianburckhardt, line 39: for consistency, I think we should consider moving this into `IScheduleHandle`, i.e. users would instead call `GetScheduleHandle().CreateAsync()`, or rather `GetScheduleClient().CreateAsync()` if we rename `IScheduleHandle` as suggested.

src/ScheduledTasks/Client/IScheduledTaskClient.cs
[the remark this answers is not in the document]
YunchuWang: make sense, for durabletaskclient there are grpcdurabletaskclient and shim. i was thinking scheduletaskclient as abstract class but uncertain which subclass names to use, right now the scheduledtaskclient is general to use both grpc or shim. maybe call subclass DefaultScheduledTaskClient?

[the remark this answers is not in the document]
YunchuWang: same for scheduleHandle, schedulehandle(abstract class) vs defaultschedulehandle(subclass)?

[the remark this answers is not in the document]
sebastianburckhardt: I would probably use `ScheduledTaskClientImpl` and `ScheduleClientImpl`. These should be internal so our users don't see them and we can change the name whenever we want to.

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <returns>A pageable list of schedule descriptions.</returns>
+    AsyncPageable<ScheduleDescription> ListSchedulesAsync(ScheduleQuery? filter = null);
+
+    /// <summary>
+    /// Creates a new schedule with the specified configuration.
+    /// </summary>
+    /// <param name="creationOptions">The options for creating the schedule.</param>
+    /// <param name="cancellation">Optional cancellation token.</param>
+    /// <returns>A handle to the created schedule.</returns>
+    Task<ScheduleClient> CreateScheduleAsync(ScheduleCreationOptions creationOptions, CancellationToken cancellation = default);
```
sebastianburckhardt, line 39: To be consistent it may be better to put this method into the `ScheduleClient` as well. So then the user calls `GetScheduleClient().CreateAsync(...)`.

src/ScheduledTasks/Models/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,176 @@
...
+    /// </summary>
+    /// <param name="scheduleId">The ID of the schedule.</param>
+    /// <param name="orchestrationName">The name of the orchestration to schedule.</param>
+    /// <param name="interval">The interval between schedule executions. Must be positive and at least 1 second.</param>
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
+    public ScheduleConfiguration(string scheduleId, string orchestrationName, TimeSpan interval)
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring as nullable.
+    {
+        this.ScheduleId = Check.NotNullOrEmpty(scheduleId, nameof(scheduleId));
+        this.OrchestrationName = Check.NotNullOrEmpty(orchestrationName, nameof(orchestrationName));
```
sebastianburckhardt, line 25: if you change this to `this.orchestrationName = ...`  then you can remove the `#pragma warning disable`

src/ScheduledTasks/Client/IScheduledTaskClient.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <returns>A handle to manage the schedule.</returns>
+    IScheduleHandle GetScheduleHandle(string scheduleId);
+
+    /// <summary>
+    /// Gets a schedule description by its ID.
+    /// </summary>
+    /// <param name="scheduleId">The ID of the schedule to retrieve.</param>
+    /// <param name="cancellation">Optional cancellation token.</param>
+    /// <returns>The schedule description if found, null otherwise.</returns>
+    Task<ScheduleDescription?> GetScheduleAsync(string scheduleId, CancellationToken cancellation = default);
```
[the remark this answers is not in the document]
YunchuWang: they differ in describe throw notfound exception and this one return nullable schedule. to give customers different options of getting schedule.

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,182 @@
...
+                EndAt = config?.EndAt,
+                Interval = config?.Interval,
+                StartImmediatelyIfLate = config?.StartImmediatelyIfLate,
+                Status = state.Status,
+                ExecutionToken = state.ExecutionToken,
+                LastRunAt = state.LastRunAt,
+                NextRunAt = state.NextRunAt,
+            };
+        }
+        catch (OperationCanceledException ex)
```
sebastianburckhardt, line 98: This exception handler seems not quite right:
(a) when operations are cancelled normally we need not log it as an error
(b) we should be logging all other exceptions

Also, I don't think it is necessary to wrap exceptions.

Here is sort of what I would expect to see:
```csharp
catch (OperationCanceledException) when (cancellation.IsCancellationRequested)
{
    // the operation was cancelled as requested. No need to log this.
    throw;
}
catch (Exception ex)
{
    this.logger.ClientError(
        nameof(this.GetScheduleAsync),
        scheduleId,
        ex);

    throw;
}
```

src/ScheduledTasks/Client/IScheduledTaskClient.cs
YunchuWang: i was originally having create in schedulehandle. Chris suggested to move it out as schedulehandle shall exist after it is created. as schedule creation is responsibiilty of scheduledtasksclient. i am open to both

src/ScheduledTasks/Client/IScheduledTaskClient.cs
YunchuWang: ditto. i think it make sense to have create schedule logic in scheduledtaskclient and have all the operations on an existing schedule inside scheduleclient

src/ScheduledTasks/Exception/ScheduleAlreadyExistException.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Exception thrown when attempting to create a schedule with an ID that already exists.
+/// </summary>
+public class ScheduleAlreadyExistsException : InvalidOperationException
```
sebastianburckhardt, line 9: Where is this used? I couldn't find any references.

YunchuWang: since we do createorupdate or createifnotexists, this is not needed anymore, will remove

src/ScheduledTasks/Models/ScheduleConfiguration.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// Updates this configuration with the provided update options.
+    /// </summary>
+    /// <param name="updateOptions">The options to update the configuration with.</param>
+    /// <returns>A set of field names that were updated.</returns>
+    public HashSet<string> Update(ScheduleUpdateOptions updateOptions)
+    {
+        Check.NotNull(updateOptions, nameof(updateOptions));
+        HashSet<string> updatedFields = new HashSet<string>();
+
+        if (!string.IsNullOrEmpty(updateOptions.OrchestrationName))
```
sebastianburckhardt, line 130: should this be also checking if the value is different? Like
```csharp
if (!string.IsNullOrEmpty(updateOptions.OrchestrationName) 
    && updateOptions.OrchestrationName != this.OrchestrationName)
```

so that we can tell if the schedule is actually being updated.

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,301 @@
...
+                throw new ScheduleInvalidTransitionException(scheduleCreationOptions?.ScheduleId ?? string.Empty, this.State.Status, ScheduleStatus.Active, nameof(this.CreateSchedule));
+            }
+
+            if (scheduleCreationOptions == null)
+            {
+                throw new ScheduleClientValidationException(string.Empty, "Schedule creation options cannot be null");
+            }
+
+            this.State.ScheduleConfiguration = ScheduleConfiguration.FromCreateOptions(scheduleCreationOptions);
+            this.TryStatusTransition(nameof(this.CreateSchedule), ScheduleStatus.Active);
```
sebastianburckhardt, line 44: At this point `this.CanTransitionTo` has already been checked, so there is no need to check it again inside `TryStatusTransition`. 

So you could simplify this line to just `this.State.Status = ScheduleStatus.Active`

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,301 @@
...
+                ex);
+        }
+    }
+
+    bool CanTransitionTo(string operationName, ScheduleStatus targetStatus)
+    {
+        return ScheduleTransitions.IsValidTransition(operationName, this.State.Status, targetStatus);
+    }
+
+    void TryStatusTransition(string operationName, ScheduleStatus to)
```
sebastianburckhardt, line 263: Can be removed, if following suggestions above.

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,301 @@
...
+                        break;
+
+                    // TODO: add other fields's callback logic after config update if any
+                    default:
+                        break;
+                }
+            }
+
+            this.State.RefreshScheduleRunExecutionToken();
+
```
sebastianburckhardt, line 111: Don't we need a `context.SignalEntity(...` here also? 

The previous signal is no longer valid after we change the execution token.

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,301 @@
...
+            if (updatedScheduleConfigFields.Count == 0)
+            {
+                // no need to interrupt and update current schedule run as there is no change in the schedule config
+                this.logger.ScheduleOperationDebug(this.State.ScheduleConfiguration.ScheduleId, nameof(this.UpdateSchedule), "Schedule configuration is up to date.");
+                return;
+            }
+
+            this.State.ScheduleLastModifiedAt = DateTimeOffset.UtcNow;
+
+            // after schedule config is updated, perform post-config-update logic separately
```
sebastianburckhardt, line 94: why do we need this section?

src/ScheduledTasks/Entity/Schedule.cs
YunchuWang: should signal if status is active

src/ScheduledTasks/Logging/Logs.Entity.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Log messages.
+/// </summary>
+/// <remarks>
+/// NOTE: Trying to make logs consistent with https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Logging/LogEvents.cs.
```
sebastianburckhardt, line 12: consistent in what way?

YunchuWang: need some advice on this part, consistent with durabletask.core logging eventids. i checked it range from 0 to 80, and put my new loggings after it

YunchuWang: are these eventid ranges look right or there are specific ranges i should put into

samples/ScheduleWebApp/ScheduleController.cs
```diff
@@ -0,0 +1,261 @@
...
+    {
+        try
+        {
+            AsyncPageable<ScheduleDescription> schedules = this.scheduledTaskClient.ListSchedulesAsync();
+
+            // add schedule result list 
+            List<ScheduleDescription> scheduleList = new List<ScheduleDescription>();
+            // Initialize the continuation token
+            string? continuationToken = null;
+            await foreach (Page<ScheduleDescription> page in schedules.AsPages(continuationToken))
```
sebastianburckhardt, line 117: loop can be simplified:

```csharp
  await foreach (var schedule in schedules)
  {
      scheduleList.Add(schedule);
  }
```

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,186 @@
...
+                        })
+                        .ToList();
+
+                    return new Page<ScheduleDescription>(schedules, entityPage.ContinuationToken);
+                }
+
+                // Return empty page if no results
+                return new Page<ScheduleDescription>(new List<ScheduleDescription>(), null);
+            }
+            catch (OperationCanceledException e)
```
sebastianburckhardt, line 179: See earlier comment about handling cancellation exceptions (I think you may have missed this one when refactoring all the other ones).

src/ScheduledTasks/Client/ScheduleClientImpl.cs
```diff
@@ -0,0 +1,275 @@
...
+        this.EntityId = new EntityInstanceId(nameof(Schedule), this.ScheduleId);
+    }
+
+    /// <summary>
+    /// Gets the entity ID of the schedule.
+    /// </summary>
+    EntityInstanceId EntityId { get; }
+
+    /// <inheritdoc/>
+    public override async Task CreateAsync(ScheduleCreationOptions creationOptions, CancellationToken cancellation = default)
```
sebastianburckhardt, line 40: We should not have duplicate implementations of `CreateAsync` or `DescribeAsync`. It is bad for code maintenance as subtle differences can sneak in. They are already not identical (one of them creates a log, the other one doesn't) which is almost certainly going to confuse us at some point. 

I would just have one of these call the other ones, instead of duplicating the code.

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,195 @@
...
+/// </summary>
+#pragma warning disable CA1711 // Identifiers should not have incorrect suffix
+public class ScheduledTaskClientImpl(DurableTaskClient durableTaskClient, ILogger logger) : ScheduledTaskClient
+#pragma warning restore CA1711 // Identifiers should not have incorrect suffix
+{
+    readonly DurableTaskClient durableTaskClient = Check.NotNull(durableTaskClient, nameof(durableTaskClient));
+    readonly ILogger logger = Check.NotNull(logger, nameof(logger));
+
+    /// <inheritdoc/>
+    public override async Task<ScheduleClient> CreateScheduleAsync(ScheduleCreationOptions creationOptions, CancellationToken cancellation = default)
```
sebastianburckhardt, line 22: should not duplicate implementation code between ScheduleClientImpl.cs and ScheduledTaskClientImpl.cs

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,195 @@
...
+                nameof(this.CreateScheduleAsync),
+                creationOptions.ScheduleId,
+                ex);
+
+            throw;
+        }
+    }
+
+    /// <inheritdoc/>
+    public override async Task<ScheduleDescription?> GetScheduleAsync(string scheduleId, CancellationToken cancellation = default)
```
sebastianburckhardt, line 67: should not duplicate implementation code between ScheduleClientImpl.cs and ScheduledTaskClientImpl.cs

src/ScheduledTasks/Logging/Logs.Entity.cs
sebastianburckhardt: I am not sure if there is a requirement on ranges being distinct, and what should be done in this context. Maybe @cgillum has some answers.

cgillum: At one point I had an idea that using consistent event IDs could be beneficial in cases where we wanted to write Kusto queries that were consistent across different SDKs, but over time I've realized that it's quite hard to maintain this, and the benefit is very unclear. I'd suggest not worrying about maintaining consistency.

jviau: Yeah, I think we are making it harder for ourselves. I think we can even let the source generator determine these event IDs?

src/ScheduledTasks/Models/ScheduleState.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    public void RefreshScheduleRunExecutionToken()
+    {
+        this.ExecutionToken = Guid.NewGuid().ToString("N");
+    }
+
+    /// <summary>
+    /// Gets or sets the time when this schedule was last paused.
+    /// </summary>
+    public DateTimeOffset? LastPausedAt { get; set; }
```
sebastianburckhardt: Is this still needed? I couldn't find any uses.

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,317 @@
...
+            foreach (string updatedScheduleConfigField in updatedScheduleConfigFields)
+            {
+                switch (updatedScheduleConfigField)
+                {
+                    case nameof(this.State.ScheduleConfiguration.StartAt):
+                    case nameof(this.State.ScheduleConfiguration.Interval):
```
sebastianburckhardt, line 111: do you need to include StartImmediatelyIfLate in this list also? I think it can alter the NextRunAt calculation also.

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,180 @@
...
+    {
+        Check.NotNullOrEmpty(scheduleId, nameof(scheduleId));
+        return new ScheduleClientImpl(this.durableTaskClient, scheduleId, this.logger);
+    }
+
+    /// <inheritdoc/>
+    public override AsyncPageable<ScheduleDescription> ListSchedulesAsync(ScheduleQuery? filter = null)
+    {
+        // Create an async pageable using the Pageable.Create helper
+        return Pageable.Create(async (continuationToken, pageSize, cancellation) =>
```
sebastianburckhardt, line 97: This may returned underfilled or empty pages, since you are applying a filter to the pages that are returned by the entity query. We can probably live with that, given that I don't think there is a simple solution.

src/ScheduledTasks/Client/ScheduledTaskClientImpl.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            try
+            {
+                // TODO: map to entity query last modified from/to filters
+                EntityQuery query = new EntityQuery
+                {
+                    InstanceIdStartsWith = filter?.ScheduleIdPrefix ?? nameof(Schedule),
```
sebastianburckhardt, line 104: We need to always specify a filter, otherwise this returns ALL of the user-defined entities as well. There could be millions of those and which would completely break this query.

This should work:
```csharp
InstanceIdStartsWith = $"@{nameof(Scheduler)}@{filter?.ScheduleIdPrefix ?? string.Empty}"
```

YunchuWang:
fyi branch for e2e tests https://github.com/microsoft/durabletask-dotnet/tree/wangbill/schedule-e2e-tests. i added 32 testcases all passing 
image

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,348 @@
...
+            return;
+        }
+
+        this.State.NextRunAt = this.DetermineNextRunTime(scheduleConfig);
+
+        DateTimeOffset currentTime = DateTimeOffset.UtcNow;
+
+        if (this.State.NextRunAt!.Value <= currentTime)
+        {
+            this.StartOrchestration(context, currentTime);
```
sebastianburckhardt, line 254: I was expecting that you would use `this.State.NextRunAt!.Value` instead of `now` on this line and the next line. So that the next runtime is based on the ideal time, not the actual time. Otherwise imprecisions add up over time. For example, if the user specifies start time 00:00:00 and interval 1h, we want the start time to be precisely on the hour for all subsequent iterations (not slowly diverging).

src/ScheduledTasks/Client/ScheduleClientImpl.cs
```diff
@@ -0,0 +1,275 @@
...
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Represents a handle to a scheduled task, providing operations for managing the schedule.
+/// </summary>
+// TODO: Isolate system entity from user entities
+class ScheduleClientImpl : ScheduleClient
```
cgillum, line 15: nit: .NET convention is not to use "Impl" suffix, but to instead use "Default" prefix. So in this case, the class name would be `DefaultScheduleClient`.

src/ScheduledTasks/Client/ScheduleClientImpl.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            // the operation was cancelled as requested. No need to log this.
+            throw;
+        }
+        catch (Exception ex)
+        {
+            this.logger.ClientError(
+                nameof(this.CreateAsync),
+                this.ScheduleId,
+                ex);
```
cgillum, line 71: nit: make these log calls a single line. No need to wrap such short lines, creating more vertical scroll.

samples/ScheduleConsoleApp/Program.cs
```diff
@@ -0,0 +1,58 @@
...
+        options.TimestampFormat = "yyyy-MM-ddTHH:mm:ss.fffZ ";
+    });
+});
+
+IHost host = builder.Build();
+await host.StartAsync();
+
+// Run the schedule operations
+ScheduledTaskClient scheduledTaskClient = host.Services.GetRequiredService<ScheduledTaskClient>();
+ScheduleOperations scheduleOperations = new ScheduleOperations(scheduledTaskClient);
```
cgillum, line 55: Why is this `ScheduleOperations` class needed? It looks like something that would belong in an SDK and I think it will confuse users who evaluate this sample app. If the purpose is just to remove test code from Program.cs, then use a static method for this. Something like `await ScheduleOperationTests.RunAsync(scheduledTaskClient)`.

sebastianburckhardt (approved):
LGTM. 

(I did not review the tests)

src/ScheduledTasks/Client/DefaultScheduleClient.cs
```diff
@@ -0,0 +1,257 @@
...
+using Microsoft.DurableTask.Client.Entities;
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Represents a handle to a scheduled task, providing operations for managing the schedule.
+/// </summary>
+// TODO: Isolate system entity from user entities
```
cgillum, line 14: What's the plan for isolating system entities from user entities? Do we have a work item tracking this?

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,348 @@
...
+    /// </summary>
+    /// <param name="context">The task entity context.</param>
+    /// <param name="executionToken">The execution token for the schedule.</param>
+    /// <exception cref="InvalidOperationException">Thrown when the schedule is not active or interval is not specified.</exception>
+    public void RunSchedule(TaskEntityContext context, string executionToken)
+    {
+        if (this.State.Status == ScheduleStatus.Uninitialized)
+        {
+            // this signal is no longer useful since the schedule has been deleted.
+            this.State = null!; // delete again, otherwise an uninitialized schedule will stick around
```
cgillum, line 211: Looking at this use of `null!`, shouldn't we make `this.State` to be nullable? I'm worried about unintended `NullReferenceExceptions`. For example, if you call `RunSchedule` twice in a row, won't hat result in a NRE?

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,348 @@
...
+            else
+            {
+                // Use configured instance ID which will prevent concurrent runs
+                startOrchestrationOptions = new StartOrchestrationOptions(instanceId);
+            }
+
+            this.logger.ScheduleOperationInfo(
+                this.State.ScheduleConfiguration!.ScheduleId,
+                nameof(this.StartOrchestration),
+                $"Starting new orchestration with instance ID: {instanceId}");
```
cgillum, line 291: Should we log the name of the orchestration we're starting?

src/ScheduledTasks/Entity/Schedule.cs
```diff
@@ -0,0 +1,348 @@
...
+                startOrchestrationOptions = new StartOrchestrationOptions(instanceId);
+            }
+            else
+            {
+                // Use configured instance ID which will prevent concurrent runs
+                startOrchestrationOptions = new StartOrchestrationOptions(instanceId);
+            }
+
+            this.logger.ScheduleOperationInfo(
+                this.State.ScheduleConfiguration!.ScheduleId,
```
cgillum, line 289: Isn't there a risk of a NRE here? Line 273 suggests that a null `ScheduleConfiguration` is possible, and if it's null there, it will be null here too.

src/ScheduledTasks/Extension/DurableTaskClientBuilderExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+public static class DurableTaskClientBuilderExtensions
+{
+    /// <summary>
+    /// Enables scheduled tasks support for the client builder.
+    /// </summary>
+    /// <param name="builder">The client builder to add scheduled task support to.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskClientBuilder UseScheduledTasks(this IDurableTaskClientBuilder builder)
+    {
+        builder.Services.AddTransient<ScheduledTaskClient>(sp =>
```
cgillum, line 22: Why transient and not singleton? Also, why are we manually constructing the `DefaultScheduledTaskClient`? Couldn't we just do this?
```csharp
builder.Services.AddSingleton<ScheduledTaskClient, DefaultScheduledTaskClient>();
```

src/ScheduledTasks/Extension/DurableTaskSchedulerWorkerExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License.
+using Microsoft.DurableTask.Worker;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.DurableTask.ScheduledTasks;
+
+/// <summary>
+/// Extension methods for configuring Durable Task workers to use the Azure Durable Task Scheduler service.
+/// </summary>
+public static class DurableTaskSchedulerWorkerExtensions
```
cgillum, line 11: Wrong class name.
```suggestion
public static class DurableTaskWorkerExtensions
```

src/ScheduledTasks/Logging/Logs.Client.cs
```diff
@@ -0,0 +1,36 @@
...
+
+/// <summary>
+/// Log messages.
+/// </summary>
+static partial class Logs
+{
+    [LoggerMessage(EventId = 80, Level = LogLevel.Information, Message = "Creating schedule with options: {scheduleConfigCreateOptions}")]
+    public static partial void ClientCreatingSchedule(this ILogger logger, ScheduleCreationOptions scheduleConfigCreateOptions);
+
+    [LoggerMessage(EventId = 82, Level = LogLevel.Information, Message = "Pausing schedule '{scheduleId}'")]
```
cgillum, line 16: Non-blocking, but it seems we skipped `Eventid = 81`. :)

cgillum (commented):
Added some comments. I'm particularly concerned about the possibility of null-ref exceptions in the Schedule.cs code. We're overriding the nullable reference types checks in ways that seem to open the door for lots of null-ref exceptions.

PR #369, opened 2025-01-27
Provide entity support for direct grpc connections to DTS

sebastianburckhardt:
implements the necessary protocol conversions and configurations for providing entity support when connecting a DTS backend directly ("portable SDK").

### DIFF
--- a/src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
+++ b/src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
@@ -85,6 +85,11 @@ public static class DurableTaskSchedulerClientExtensions
             .Configure(additionalConfig ?? (_ => { }))
             .ValidateDataAnnotations();
 
+        builder.Configure(options =>
+        {
+            options.EnableEntitySupport = true;
+        });
+
         builder.Services.TryAddEnumerable(
             ServiceDescriptor.Singleton<IConfigureOptions<GrpcDurableTaskClientOptions>, ConfigureGrpcChannel>());
         builder.UseGrpc(_ => { });
--- a/src/Shared/Grpc/EntityConversions.cs
+++ b/src/Shared/Grpc/EntityConversions.cs
@@ -1,0 +1,443 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+using System.Threading.Tasks;
+using DurableTask.Core;
+using DurableTask.Core.Entities;
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using Microsoft.DurableTask;
+using Newtonsoft.Json;
+using DTCore = DurableTask.Core;
+using P = Microsoft.DurableTask.Protobuf;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Utilities for converting between representations of entity history events. The older backends represent entity
+/// messages as external events, using a JSON encoding defined in DT Core.
+/// Starting with the DTS backend, we use explicit, separate
+/// protobuf encodings for all entity-related history events.
+/// </summary>
+static class EntityConversions
+{
+    // we copied the relevant data members from DT.Core to allow us to convert between the data structures
+    // used in the backend, and the legacy encoding of entities within orchestration histories.
+
+    // The point of this class is to reverse a Newtonsoft serialization that happened in prior DT code.
+    // To do this reliably we use the same Newtonsoft.
+    // This is not introducing a new dependency, and should be eliminated once the original dependency is eliminated.
+    static readonly JsonSerializerSettings ConversionSettings = new JsonSerializerSettings()
+    {
+        TypeNameHandling = TypeNameHandling.None, // the type names do not match, so we have to disable them
+    };
+
+    /// <summary>
+    /// Encodes an operation signal.
+    /// </summary>
+    /// <param name="protoEvent">The proto event.</param>
+    /// <returns>The core event.</returns>
+    public static DTCore.History.HistoryEvent EncodeOperationSignaled(P.HistoryEvent protoEvent)
+    {
+        P.EntityOperationSignaledEvent signaledEvent = protoEvent.EntityOperationSignaled;
+        DateTime? scheduledTime = signaledEvent.ScheduledTime?.ToDateTime();
+        string name = EncodeEventName(scheduledTime);
+        string input = JsonConvert.SerializeObject(
+            new RequestMessage()
+            {
+                Operation = signaledEvent.Operation,
+                IsSignal = true,
+                Input = signaledEvent.Input,
+                Id = signaledEvent.RequestId,
+                ScheduledTime = scheduledTime,
+            },
+            ConversionSettings);
+
+        string? target = signaledEvent.TargetInstanceId
+            ?? throw new InvalidOperationException("missing target instance id");
+
+        return CreateEventRaisedOrSentEvent(protoEvent.EventId, name, input, target);
+    }
+
+    /// <summary>
+    /// Encodes an operation call.
+    /// </summary>
+    /// <param name="protoEvent">The proto event.</param>
+    /// <param name="instance">The orchestration instance.</param>
+    /// <returns>The core event.</returns>
+    public static DTCore.History.HistoryEvent EncodeOperationCalled(
+        P.HistoryEvent protoEvent,
+        OrchestrationInstance? instance)
+    {
+        P.EntityOperationCalledEvent calledEvent = protoEvent.EntityOperationCalled;
+        DateTime? scheduledTime = calledEvent.ScheduledTime?.ToDateTime();
+        string name = EncodeEventName(scheduledTime);
+        string input = JsonConvert.SerializeObject(
+            new RequestMessage()
+            {
+                Operation = calledEvent.Operation,
+                IsSignal = false,
+                Input = calledEvent.Input,
+                Id = calledEvent.RequestId,
+                ScheduledTime = scheduledTime,
+                ParentInstanceId = instance?.InstanceId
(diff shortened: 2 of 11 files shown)


sebastianburckhardt:
I am marking this as ready for review now that I have some tests passing. 

Given that iterations are slow, I want to run a lot more tests before merging though.

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
```diff
@@ -85,6 +85,12 @@ static void ConfigureSchedulerOptions(
             .Configure(additionalConfig ?? (_ => { }))
             .ValidateDataAnnotations();
 
+        builder.Services.AddOptions<DurableTaskClientOptions>(builder.Name)
+            .Configure(options =>
```
YunchuWang, line 89: shall we chain this configure in         builder.Services.AddOptions<DurableTaskSchedulerClientOptions>(builder.Name)
            .Configure(initialConfig)
            .Configure(additionalConfig ?? (_ => { }))
            .ValidateDataAnnotations();

YunchuWang (approved):

YunchuWang:
LGTM

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
```diff
@@ -85,6 +85,12 @@ static void ConfigureSchedulerOptions(
             .Configure(additionalConfig ?? (_ => { }))
             .ValidateDataAnnotations();
 
+        builder.Services.AddOptions<DurableTaskClientOptions>(builder.Name)
```
jviau, line 88: I think there is a `builder.Configure(optionsCallback)` overload you can use

src/Worker/Grpc/GrpcDurableTaskWorkerOptions.cs
```diff
@@ -22,4 +22,17 @@ public sealed class GrpcDurableTaskWorkerOptions : DurableTaskWorkerOptions
     /// Gets or sets the gRPC call invoker to use. Will supersede <see cref="Address" /> when provided.
     /// </summary>
     public CallInvoker? CallInvoker { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether entity-related events appearing in orchestration histories should be
+    /// automatically converted back and forth between the old DT Core representation (JSON-encoded external events)
+    /// and the new protobuf representation (explicit history events), which is used by the DTS scheduler backend.
+    /// </summary>
+    public bool ConvertOrchestrationEntityEvents { get; set; }
```
jviau, line 31: This value and `InsertEntityUnlocksOnCompletion` look like they are more implementation details and not a public API. I would prefer we figure out a way to make these non-public if possible. If this must be public due to different projects, then we should find a way to hide this in an `.Internal` namespace and add comments declaring this not meant for public consumption.

Here is an example: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/Internal/IOrchestrationSubmitter.cs

One idea would be to make these properties `internal` (maybe even in a sub-class to keep them together). Then add an extension method `ConfigureForAzureManaged(this GrpcDurableTaskWorkerOptions options)` under `Microsoft.DurableTask.Worker.Grpc.Internal` namespace with the comments. Then AzureManaged can call that.

jviau: Yes, you can do that for now. I did it in the Abstraction package:

https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/Abstractions.csproj#L22-L23

Please just add a comment saying these are used by AzureManaged to treat them as public API breaking change rules

But I have since changed my stance on how much I like that (I don't recommend it anymore). The reason is this can still cause issues down the road. If customers use `Worker.AzureManager/1.n` which uses these cross-assembly `internal` properties, but _also_ use `Worker.Grpc/1.n+1` which has since removed these properties in a refactor, the app will throw method missing exceptions.

BUT given we already use this pattern in the code base, I think the comment will be good enough.

sebastianburckhardt: I tried the "internals visible" approach and it worked but it is also causing some warnings (because there are conflicts between doubly defined classes, those come from files that are shared between the projects). 

So I will instead do what you suggested before (create a public class that is hidden in a namespace and has disclaimers on it).

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
sebastianburckhardt: yes, much better.

src/Worker/Grpc/InternalOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <summary>
+/// Provides access to configuring internal options for the gRPC worker.
+/// </summary>
+public static class InternalOptionsExtensions
+{
+    /// <summary>
+    /// Configure the worker to use the default settings for connecting to the Azure Managed Durable Task service.
+    /// </summary>
+    /// <param name="options">The gRPC worker options.</param>
+    public static void ConfigureForAzureManaged(this GrpcDurableTaskWorkerOptions options)
```
cgillum, line 19: Can this be moved to the `AzureManaged` project? Code in the `Grpc` project is supposed to be backend-agnostic.

src/Shared/Grpc/EntityConversions.cs
```diff
@@ -0,0 +1,443 @@
...
+/// protobuf encodings for all entity-related history events.
+/// </summary>
+static class EntityConversions
+{
+    // we copied the relevant data members from DT.Core to allow us to convert between the data structures
+    // used in the backend, and the legacy encoding of entities within orchestration histories.
+
+    // The point of this class is to reverse a Newtonsoft serialization that happened in prior DT code.
+    // To do this reliably we use the same Newtonsoft.
+    // This is not introducing a new dependency, and should be eliminated once the original dependency is eliminated.
```
cgillum, line 35: Just to be clear, if we remove the DT.Core dependency from this SDK, will that also allow us to remove this dependency on Newtonsoft.Json?

cgillum:
@sebastianburckhardt you'll need to rebase this PR on the latest from main. We no longer use submodules for updating the protobuf definitions. See this change: https://github.com/microsoft/durabletask-dotnet/pull/370.

src/Worker/Grpc/InternalOptionsExtensions.cs
jviau: This is based on a suggestion of mine. The properties being added in `InternalOptions` are specialized for AzureManaged scenario, but they need to be in the project. To avoid having these options that customer's are not expected to ever configure, we make the values themselves `internal` and only expose a single `public` point to configure them.

With that said, we can name this method more backend-agnostic, but it does need to remain in this project so it can access the internal properties. Additionally, we should include a disclaimer in the XML doc explaining this is considered an internal API surface and not subject to breaking change guarantees.

eng/proto
jviau, line 1: I thought this was removed

src/Worker/Grpc/InternalOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.DurableTask.Worker.Grpc.Internal;
+
+/// <summary>
+/// Provides access to configuring internal options for the gRPC worker.
+/// </summary>
+public static class InternalOptionsExtensions
```
jviau, line 13: Please include disclaimers as seen on this type: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/Internal/IOrchestrationSubmitter.cs

src/Shared/Grpc/EntityConversions.cs
sebastianburckhardt: Yes. The point of this class is to "undo Newtonsoft serialization" that was unfortunately already done inside DT.Core. So the Newtonsoft dependency goes away when either (a) Newtonsoft dependency is removed from DT.Core, or (b) DT.Core dependency is removed.

src/Worker/Grpc/InternalOptions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Worker.Grpc;
+
+/// <summary>
+/// Options for controlling backend-specific features. These are not exposed directly, but can be accessed via
+/// the extension methods in <see cref="Internal.InternalOptionsExtensions"/>.
+/// </summary>
+class InternalOptions
```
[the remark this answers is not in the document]
sebastianburckhardt: I am making this a nested class.

eng/proto
sebastianburckhardt: Will merge second-to-latest commit in main which introduced the vendor pattern.

sebastianburckhardt: I rebased on latest main now so this is no longer there.

src/Worker/Grpc/InternalOptionsExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.DurableTask.Worker.Grpc.Internal;
```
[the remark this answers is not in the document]
sebastianburckhardt: Moved.

src/Worker/Grpc/InternalOptionsExtensions.cs
sebastianburckhardt: I kept the name (we will confuse only ourselves if we invent more names) but there is now an XML doc that states that this is an internal API.

src/Worker/Grpc/InternalOptionsExtensions.cs
sebastianburckhardt: I have added an XML comment on the extension method

eng/proto
sebastianburckhardt, line 1: For some reason this is still showing up. Will try to get rid of it.

PR #370, opened 2025-01-28
Add support for history streaming

cgillum:
This PR adds support for history event streaming, which is a mechanism intended to unblock scenarios where the size of the orchestration instance history exceeds the gRPC message size limit. It assumes that the backend service that implements the gRPC protocol supports a newly added `supportsHistoryStreaming` feature. The corresponding protobuf change is here: https://github.com/microsoft/durabletask-protobuf/pull/30.

This PR also removes our use of submodules and replaces it with a script that downloads the latest protobuf files locally. This is to make the management of the protobuf files easier and more maintainable.


src/Grpc/Grpc.csproj
```diff
@@ -18,7 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Protobuf Include="$(EngRoot)proto/protos/orchestrator_service.proto" GrpcServices="both" Link="proto/orchestrator_service.proto" />
+    <Protobuf Include="*.proto" GrpcServices="Both" />
```
jviau, line 21: Consider a recursive glob just in-case we structure this folder down the road.

```suggestion
    <Protobuf Include="**/*.proto" GrpcServices="Both" />
```

src/Grpc/refresh-protos.ps1
jviau, line 1: Curious why you are taking this approach over submodules?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -87,26 +86,17 @@ public async Task ExecuteAsync(CancellationToken cancellation)
             }
         }
 
-        static OrchestrationRuntimeState BuildRuntimeState(P.OrchestratorRequest request)
+        static IEnumerable<T> Append<T>(IEnumerable<T> source, IEnumerable<T> items)
```
jviau, line 89: Is this method needed anymore, or can it be replaced with `source.Concat(items)`?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -128,21 +118,71 @@ static string GetActionsListForLogging(IReadOnlyList<P.OrchestratorAction> actio
...
+                    ExecutionId = orchestratorRequest.ExecutionId,
+                    ForWorkItemProcessing = true,
+                };
+
+                using AsyncServerStreamingCall<P.HistoryChunk> streamResponse =
+                    this.client.StreamInstanceHistory(streamRequest, cancellationToken: cancellation);
+
+                await foreach (P.HistoryChunk chunk in streamResponse.ResponseStream.ReadAllAsync(cancellation))
+                {
+                    pastEvents = Append(pastEvents, chunk.Events.Select(ProtoUtils.ConvertHistoryEvent));
```
jviau, line 141: Can use LINQ here:

```suggestion
                    pastEvents = pastEvents.Concat(chunk.Events.Select(ProtoUtils.ConvertHistoryEvent));
```

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -128,21 +118,71 @@ static string GetActionsListForLogging(IReadOnlyList<P.OrchestratorAction> actio
             }
         }
 
+        async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
+            P.OrchestratorRequest orchestratorRequest,
+            CancellationToken cancellation)
+        {
+            IEnumerable<HistoryEvent> pastEvents = [];
+            if (orchestratorRequest.RequiresHistoryStreaming)
```
jviau, line 126: consider moving this block into its own method `GetAllHistoryAsync(..)` or similarly named.

src/Grpc/refresh-protos.ps1
cgillum: This approach is known more broadly as _vendoring_. The Dapr engineers strongly encouraged this approach for managing protobufs rather than git submodules because the DevEx for working with git submodules is quite unintuitive (which I agree with). I went with this approach for the DTS SDKs for Functions and am quite happy with the simplicity it has created. Moving forward, I'd like all SDKs to use this approach instead of submodules to help improve the overall DevEx.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
cgillum: Oh, I got confused and thought the LINQ `Append` method (which is only in .NET 6+) was `Concat`. Yeah, `Concat` is what I want here. I'll remove this Append method.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
cgillum: It's not that much code and isn't required elsewhere, so I'd prefer to keep it inlined to minimize the code's overall conceptual overhead.

src/Grpc/Grpc.csproj
cgillum: `*/*.proto` didn't work, but I got it working with `**/*.proto`.

YunchuWang (approved):

PR #390, opened 2025-03-05
Add timeout to gRPC workitem streaming

halspang:
This commit adds a timeout to the gRPC stream used to communicate with the backend. This was done because the backend could restart and drop the connection and the worker would not know. This causes the worker to hang and not receive any new work items. The fix is to reset the connection if a long enough period of time has passed between receiving anything on the stream.

### DIFF
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -252,7 +252,7 @@ sealed partial class GrpcDurableTaskWorker
                     if (!cancellation.IsCancellationRequested)
                     {
                         // Since the cancellation came from the timeout, log a warning.
-                        this.Logger.ConnectionClosed();
+                        this.Logger.ConnectionTimeout();
                     }
 
                     return;
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -185,6 +185,8 @@ sealed partial class GrpcDurableTaskWorker
         async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, CancellationToken cancellation)
         {
             // Create a new token source for timing out and a final token source that keys off of them both.
+            // The timeout token is used to detect when we are no longer getting any messages, including health checks.
+            // If this is the case, it signifies the connection has been dropped silently and we need to reconnect.
             using var timeoutSource = new CancellationTokenSource();
             timeoutSource.CancelAfter(TimeSpan.FromSeconds(60));
             using var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, timeoutSource.Token);
@@ -247,6 +249,12 @@ sealed partial class GrpcDurableTaskWorker
                     // The token has cancelled, this means either:
                     // 1. The broader 'cancellation' was triggered, return here to start a graceful shutdown.
                     // 2. The timeoutSource was triggered, return here to trigger a reconnect to the backend.
+                    if (!cancellation.IsCancellationRequested)
+                    {
+                        // Since the cancellation came from the timeout, log a warning.
+                        this.Logger.ConnectionClosed();
+                    }
+
                     return;
                 }
             }
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -185,9 +185,9 @@ sealed partial class GrpcDurableTaskWorker
         async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, CancellationToken cancellation)
         {
             // Create a new token source for timing out and a final token source that keys off of them both.
-            var timeoutSource = new CancellationTokenSource();
+            using var timeoutSource = new CancellationTokenSource();
             timeoutSource.CancelAfter(TimeSpan.FromSeconds(60));
-            var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, timeoutSource.Token);
+            using var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, timeoutSource.Token);
 
             while (!cancellation.IsCancellationRequested)
             {
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -1,13 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using DurableTask.Core;
 using DurableTask.Core.Entities;
 using DurableTask.Core.Entities.OperationFormat;
 using DurableTask.Core.History;
-using Grpc.Core;
 using Microsoft.DurableTask.Entities;
 using Microsoft.DurableTask.Worker.Shims;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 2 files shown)


src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -186,10 +184,16 @@ async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
 
         async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, CancellationToken cancellation)
         {
+            // Create a new token source for timing out and a final token source that keys off of them both.
+            var timeoutSource = new CancellationTokenSource();
```
jviau, line 188:
```suggestion
            using CancellationTokenSource timeoutSource = new();
```

CTS with a timer will cause memory leaks if not disposed.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -186,10 +184,16 @@ async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
...
+            timeoutSource.CancelAfter(TimeSpan.FromSeconds(60));
+            var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, timeoutSource.Token);
```
jviau, line 190:
```suggestion
            using var tokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellation, timeoutSource.Token);
```

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -186,10 +184,16 @@ async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
...
+
             while (!cancellation.IsCancellationRequested)
             {
-                await foreach (P.WorkItem workItem in stream.ResponseStream.ReadAllAsync(cancellation))
+                await foreach (P.WorkItem workItem in stream.ResponseStream.ReadAllAsync(tokenSource.Token))
```
jviau, line 194: Would this not throw if the connection is closed?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -237,6 +241,14 @@ async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, Ca
                         this.Logger.UnexpectedWorkItemType(workItem.RequestCase.ToString());
                     }
                 }
+
+                if (tokenSource.IsCancellationRequested || tokenSource.Token.IsCancellationRequested)
```
jviau, line 245: Maybe I am missing something, but it seems unlikely this line would ever be `true`. If `IsCancellationRequested`, then more likely than not `stream.ResponseStream.ReadAllAsync` throw `OpeationCancelledException`.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: We thought it would too, but if you go into the `IAsyncStreamReader`, a cancellation is actually just treated as the end of the stream and it returns normally.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: See above. We thought this behavior was an odd choice for the stream reader as well, but it's documented that it doesn't throw.

jviau:
Can you explain the observed code flow when the scheduler shuts down? Is an exception throw? If so, `ProcessWorkItemsAsync` is called with many catch statements - are one of them hit? Do we misconstrue this as a worker shutdown?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: Oops, fixed.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: Fixed.

halspang:
> Can you explain the observed code flow when the scheduler shuts down? Is an exception throw? If so, `ProcessWorkItemsAsync` is called with many catch statements - are one of them hit? Do we misconstrue this as a worker shutdown?

It doesn't throw an exception, it returns as if the stream had just ended normally. So, once the cancellation is triggered, the foreach loop exits and we check the token. If the token is cancelled, we return. If the overall cancellation was cancelled, it will exit at that level as well. If not, it creates a new connection to the scheduler.

https://grpc.github.io/grpc/csharp/api/Grpc.Core.IAsyncStreamReader-1.html

jviau (approved):
Approving to unblock, but I think this is treating a symptom and not the underlying problem. If a scheduler restart isn't closing the connection to the worker, thus ending the stream, _something_ else is keeping that alive (such as a reverse proxy). This needs to be looked at and addressed, as it is violating some fundamental expectations of gRPC streams.

I would feel slightly better if we make this privately configurable by the `AzureManaged` package somehow. We already have some psuedo-internal options they use [here](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Worker/Grpc/Internal/InternalOptionsExtensions.cs), could add to that and have this behavior only enabled for DTS.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -186,10 +184,16 @@ async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
 
         async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, CancellationToken cancellation)
         {
+            // Create a new token source for timing out and a final token source that keys off of them both.
```
cgillum, line 187: This comment explains what the code is doing, but can we add comments explaining why we're doing this? It seems a bit unintuitive that we'd want to cancel our long-lived connection after 60 seconds. Only after speaking with Phil and Bernd for a bit did I realize that this is a mechanism for detecting when we've been silently disconnected from the server (because we're not receiving health ping work items).

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -237,6 +241,14 @@ async Task ProcessWorkItemsAsync(AsyncServerStreamingCall<P.WorkItem> stream, Ca
                         this.Logger.UnexpectedWorkItemType(workItem.RequestCase.ToString());
                     }
                 }
+
+                if (tokenSource.IsCancellationRequested || tokenSource.Token.IsCancellationRequested)
+                {
+                    // The token has cancelled, this means either:
+                    // 1. The broader 'cancellation' was triggered, return here to start a graceful shutdown.
+                    // 2. The timeoutSource was triggered, return here to trigger a reconnect to the backend.
```
cgillum, line 249: Can we add a warning log when we detect that we hit the timeout condition? I feel like that would be a good distinction for us to have some telemetry on.

src/Worker/Grpc/Logs.cs
```diff
@@ -48,5 +48,8 @@ static partial class Logs
 
         [LoggerMessage(EventId = 55, Level = LogLevel.Information, Message = "{instanceId}: Evaluating custom retry handler for failed '{name}' task. Attempt = {attempt}.")]
         public static partial void RetryingTask(this ILogger logger, string instanceId, string name, int attempt);
+
+        [LoggerMessage(EventId = 56, Level = LogLevel.Warning, Message = "Channel to backend has stopped receiving traffic, will attempt to reconnect.")]
+        public static partial void ConnectionClosed(this ILogger logger);
```
cgillum, line 53: nit: can we call this `ConnectionTimeout`?

cgillum (approved):

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: Added comments.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: Added a log.

src/Worker/Grpc/Logs.cs
halspang: Updated.

cgillum:
> I think this is treating a symptom and not the underlying problem

Agreed, but I think this is a safety mechanism we want anyways, regardless of what gRPC server implementation we're targeting, so I'm happy to go with this for now. Warning logs have been added so that we can observe this behavior and be reminded that it needs to be further root caused.

PR #393, opened 2025-03-11
Introduce versioning to the DurableTaskClient

halspang:
This change adds a DefaultVersion to the DurableTaskClient builder options. When no other version is set, which is currently always the case, this value is used as the version of any new orchestration started. The value is then passed down to workers via the TaskOrchestrationContext.

This allows users to specify a version during their app setup and then key off of the version during orchestration. Using conditional logic, changes can then be made to the orchestration without harming in-progress orchestrations.

### DIFF
--- a/src/Abstractions/TaskOrchestrationContext.cs
+++ b/src/Abstractions/TaskOrchestrationContext.cs
@@ -60,6 +60,11 @@ public abstract class TaskOrchestrationContext
     public abstract bool IsReplaying { get; }
 
     /// <summary>
+    /// Gets the version of the current orchestration instance, which was set when the instance was created.
+    /// </summary>
+    public abstract string Version { get; }
+
+    /// <summary>
     /// Gets the entity feature, for interacting with entities.
     /// </summary>
     public virtual TaskOrchestrationEntityFeature Entities =>
--- a/src/Client/Core/DependencyInjection/DurableTaskClientBuilderExtensions.cs
+++ b/src/Client/Core/DependencyInjection/DurableTaskClientBuilderExtensions.cs
@@ -89,4 +89,16 @@ public static class DurableTaskBuilderExtensions
             });
         return builder;
     }
+
+    /// <summary>
+    /// Sets the default version for this builder. This version will be applied by default to all orchestrations if set.
+    /// </summary>
+    /// <param name="builder">The builder to set the version for.</param>
+    /// <param name="version">The version that will be used as the default version.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskClientBuilder UseDefaultVersion(this IDurableTaskClientBuilder builder, string version)
+    {
+        builder.Configure(options => options.DefaultVersion = version);
+        return builder;
+    }
 }
--- a/src/Client/Core/DurableTaskClientOptions.cs
+++ b/src/Client/Core/DurableTaskClientOptions.cs
@@ -14,6 +14,14 @@ public class DurableTaskClientOptions
     bool enableEntitySupport;
 
     /// <summary>
+    /// Gets or sets the version of orchestrations that will be created.
+    /// </summary>
+    /// <remarks>
+    /// Currently, this is sourced from the AzureManaged client options.
+    /// </remarks>
+    public string DefaultVersion { get; set; } = string.Empty;
+
+    /// <summary>
     /// Gets or sets the data converter. Default value is <see cref="JsonDataConverter.Default" />.
     /// </summary>
     /// <remarks>
@@ -94,6 +102,11 @@ public class DurableTaskClientOptions
             if (!other.EntitySupportExplicitlySet)
             {
                 other.EnableEntitySupport = this.EnableEntitySupport;
+            }
+
+            if (!string.IsNullOrWhiteSpace(this.DefaultVersion))
+            {
+                other.DefaultVersion = this.DefaultVersion;
             }
         }
     }
--- a/src/Client/Grpc/GrpcDurableTaskClient.cs
+++ b/src/Client/Grpc/GrpcDurableTaskClient.cs
@@ -78,10 +78,20 @@ public sealed class GrpcDurableTaskClient : DurableTaskClient
     {
         Check.NotEntity(this.options.EnableEntitySupport, options?.InstanceId);
 
+        string version = string.Empty;
+        if (!string.IsNullOrEmpty(orchestratorName.Version))
+        {
+            version = orchestratorName.Version;
+        }
+        else if (!string.IsNullOrEmpty(this.options.DefaultVersion))
+        {
+            version = this.options.DefaultVersion;
+        }
+
         var request = new P.CreateInstanceRequest
         {
             Name = orchestratorName.Name,
-            Version = orchestratorName.Version,
+            Version = version,
             InstanceId = options?.InstanceId ?? Guid.NewGuid().ToString("N"),
             Input = this.DataConverter.Serialize(input),
         };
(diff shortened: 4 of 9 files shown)


src/Client/Core/DurableTaskClientOptions.cs
```diff
@@ -13,6 +13,14 @@ public class DurableTaskClientOptions
     DataConverter dataConverter = JsonDataConverter.Default;
     bool enableEntitySupport;
 
+    /// <summary>
+    /// Gets or sets the version of orchestrations that will be created.
+    /// <remarks>
+    /// Currently, this is sourced from the AzureManaged client options.
+    /// </remarks>
```
cgillum, line 20: The `<remarks>` tag need to be a peer of the `<summary>` tags. See the other property XML doc comments below as an example.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -78,10 +78,20 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
     {
         Check.NotEntity(this.options.EnableEntitySupport, options?.InstanceId);
 
+        var version = string.Empty;
```
cgillum, line 81: code style nit: we try to avoid using `var` in this repo.
```suggestion
        string version = string.Empty;
```

src/Client/Core/DependencyInjection/DurableTaskClientBuilderExtensions.cs
```diff
@@ -89,4 +89,17 @@ public static IDurableTaskClientBuilder UseBuildTarget<TTarget, TOptions>(this I
...
+    /// <summary>
+    /// Sets the default version for this builder. This version will be applied by default to all orchestrations if set.
+    /// </summary>
+    /// <param name="builder">The builder to set the version for.</param>
+    /// <param name="version">The version that will be used as the default version.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskClientBuilder UseDefaultVersion(this IDurableTaskClientBuilder builder, string version)
+    {
+        builder.Services.AddOptions<DurableTaskClientOptions>(builder.Name)
+            .Configure(options => options.DefaultVersion = version);
```
cgillum, line 102: Should we instead do `builder.Configure(options => options.DefaultVersion = version)`?

It doesn't seem quite right for us to be calling `AddOptions` here, but I'm not an expert on the .NET options pattern.

test/Client/Grpc.Tests/DependencyInjection/DurableTaskClientBuilderExtensionsTests.cs
```diff
@@ -67,6 +67,20 @@ public void UseGrpc_Callback_Sets()
...
+    [Fact]
+    public void UseDefaultVersion_DefaultVersion_Sets()
+    {
+        ServiceCollection services = new();
+        DefaultDurableTaskClientBuilder builder = new(null, services);
+        builder.UseDefaultVersion("0.1")
+            .UseGrpc();
+
+        var provider = services.BuildServiceProvider();
+        var options = provider.GetOptions<GrpcDurableTaskClientOptions>();
```
cgillum, line 79: nit: Same comment about the use of `var`. It would be best if we could keep the code style consistent when possible (see method above where we use explicit type names).

test/Grpc.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -43,9 +43,9 @@ void IDisposable.Dispose()
         GC.SuppressFinalize(this);
     }
 
-    protected async Task<HostTestLifetime> StartWorkerAsync(Action<IDurableTaskWorkerBuilder> configure)
+    protected async Task<HostTestLifetime> StartWorkerAsync(Action<IDurableTaskWorkerBuilder> configure, Action<IDurableTaskClientBuilder>? clientConfigure = null)
```
cgillum, line 46: Let's also rename `configure` to `workerConfigure` to make it less ambiguous.

test/Grpc.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -75,6 +75,10 @@ protected IHostBuilder CreateHostBuilder(Action<IDurableTaskWorkerBuilder> confi
                 {
                     b.UseGrpc(this.sidecarFixture.Channel);
                     b.RegisterDirectly();
+                    if (clientConfigure != null)
+                    {
+                        clientConfigure!(b);
```
cgillum, line 80: The `!` here is not needed since you already checked for null above.
```suggestion
                        clientConfigure(b);
```
Alternatively, you could remove the if-conditional above and just do this: `clientConfigure?.Invoke(b)` - which is a bit more common in this repo since it's less verbose.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -561,8 +561,37 @@ public async Task SpecialSerialization()
...
+                {
+                    return $"Orchestration version: {input}";
+                }));
+        }, c =>
+        {
+            c.UseDefaultVersion(version);
+        });
+
+        var instanceId = await server.Client.ScheduleNewOrchestrationInstanceAsync("Versioned_Orchestration", input: string.Empty);
+        var result = await server.Client.WaitForInstanceCompletionAsync(instanceId, getInputsAndOutputs: true, this.TimeoutToken);
```
cgillum, line 586: I'm realizing just now that we don't have a `Version` property in `OrchestrationMetadata`. We should add this, but it doesn't have to be part of this PR since it would also likely require backend changes.

src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -59,6 +59,11 @@ public abstract class TaskOrchestrationContext
     /// </value>
     public abstract bool IsReplaying { get; }
 
+    /// <summary>
+    /// Gets the version.
```
cgillum, line 63:
```suggestion
    /// Gets the version of the current orchestration instance, which was set when the instance was created.
```

cgillum (commented):
Some minor comments.

src/Client/Core/DurableTaskClientOptions.cs
halspang: Oops, those must've gotten intertwined by accident. Fixed.

src/Client/Grpc/GrpcDurableTaskClient.cs
halspang: Done, I'll keep this in mind for future PRs :)

src/Client/Core/DependencyInjection/DurableTaskClientBuilderExtensions.cs
halspang: I'm also not intimately aware of the options pattern, so this was just going off what I had seen in this file/what copilot advised. 

That being said, I changed it to this and it works, so let's go with it as doing less configuring/adding seems safer imo.

test/Client/Grpc.Tests/DependencyInjection/DurableTaskClientBuilderExtensionsTests.cs
halspang: Done.

test/Grpc.IntegrationTests/IntegrationTestBase.cs
halspang: Done.

test/Grpc.IntegrationTests/IntegrationTestBase.cs
halspang: Always a fan of being less verbose, updated to `?.`.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: Yeah, I can follow up with this, what I'd like to add next will involve backend changes eventually so I'll be there soon, hopefully.

src/Abstractions/TaskOrchestrationContext.cs
halspang: Updated.

cgillum (approved):

PR #394, opened 2025-03-13
Add utilities for easier versioning usage

halspang:
Introduce some utilities on the `TaskOrchestrationContext` that allow users to handle versioning scenarios easier. Right now, this includes checking for standard equalities and a check to make sure the version is the correct format when passed in.

### DIFF
--- a/src/Abstractions/TaskOrchestrationContext.cs
+++ b/src/Abstractions/TaskOrchestrationContext.cs
@@ -403,6 +403,52 @@ public abstract class TaskOrchestrationContext
     public virtual ILogger CreateReplaySafeLogger<T>()
         => new ReplaySafeLogger(this, this.LoggerFactory.CreateLogger<T>());
 
+    /// <summary>
+    /// Checks if the current orchestration version is greater than the specified version.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// If both versions are empty, this is considered false as neither can be greater.
+    /// </para>
+    /// <para>
+    /// An empty context version is less than a defined version in the parameter.
+    /// </para>
+    /// <para>
+    /// An empty parameter version is less than a defined version in the context.
+    /// </para>
+    /// </remarks>
+    /// <param name="version">The version to check against.</param>
+    /// <returns>True if the orchestration's version is greater than the provided version, false otherwise.</returns>
+    public virtual int CompareVersionTo(string version)
+    {
+        // Both versions are empty, treat as equal.
+        if (string.IsNullOrWhiteSpace(this.Version) && string.IsNullOrWhiteSpace(version))
+        {
+            return 0;
+        }
+
+        // An empty version in the context is always less than a defined version in the parameter.
+        if (string.IsNullOrWhiteSpace(this.Version))
+        {
+            return -1;
+        }
+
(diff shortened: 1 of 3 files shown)


src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -403,6 +403,115 @@ public virtual ILogger CreateReplaySafeLogger(Type type)
...
+        Version contextVersion = new Version(this.Version);
+        try
+        {
+            Version parameterVersion = new Version(version);
+            return contextVersion > parameterVersion;
+        }
+        catch (Exception) // This can be several different exception types.
+        {
+            // The version string is not in a valid format.
+            return false;
```
cgillum, line 451: I wonder if we should have some fallback options here. Ideally, we could be compatible with a few different types of versions, like timestamps (e.g., `"2025-03-14"`). Maybe we even just fall back to a simple string sort, like this:
```suggestion
            return this.Version.CompareTo(version) > 0;
```

The fallback option could work in a variety of cases, even "alpha", "beta". This way, we can be a bit less opinionated about how you do versioning, which I think makes sense to do since we currently allow it to be an opaque string.

cgillum: In fact, you can probably consolidate the 3 APIs into one by making a single `int CompareToVersion(string version)` method that behaves similarly to the existing [IComparable.CompareTo](https://learn.microsoft.com/en-us/dotnet/api/system.icomparable.compareto?view=net-9.0) method, which most experienced .NET devs will be familiar with.

halspang: I feel like version strings are one of the better places to be opinionated than not. The more abstract we make the version, the more difficult it is for us to maintain a consistent ordering. 

We also can't really use a string comparison for versions because lexigraphical comparisons do things like state that `1.2 > 1.10`. Lexigraphically, that's true, but in a versioned sense, it is false. 

The `Version` class in .NET only accepts 2-4 part versions, all of which must be numbers. If we commit to that we can go down to just a `CompareTo` internally, but we do also need to keep the ability to process unversioned/mixed flows.

cgillum: The problem is that the version string is already unopinionated since it's in a string format and we don't put any restrictions on the input. This is why I think it makes sense to use string comparison in cases where we don't recognize the format. The problem of `1.2 > 1.10` is solved by the use of [`Version.TryParse`](https://learn.microsoft.com/en-us/dotnet/api/system.version.tryparse?view=net-9.0), which happens prior to the generic string comparison.

halspang: Oh, I think I misunderstood you. Are you saying to just add this as the path if we weren't able to parse it as a standard version? If that's the case, yes I can go ahead and add this as a fallback and use the `Version` library as the main parsing method.

halspang: Moved to the `CompareTo` style and removed the style enforcement on the builder.

cgillum (approved):
LGTM! Can you update the CHANGELOG.md file to reference this change?

PR #397, opened 2025-03-19
Add tags to CreateInstanceRequest

torosent:


src/Abstractions/TaskOptions.cs
```diff
@@ -1,103 +1,106 @@
...
+/// The unique ID of the orchestration instance to schedule. If not specified, a new GUID value is used.
+/// </param>
+/// <param name="StartAt">
+/// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
+/// is specified, the orchestration instance will be scheduled immediately.
+/// </param>
+/// <param name="Tags">
+/// A dictionary of tags to associate with the orchestration instance.
+/// </param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null, IDictionary<string, string>? Tags = null);
```
cgillum, line 106: I think this might be a binary breaking change. I suggest instead adding this as an init-only property:

```suggestion
public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null)
{
    public IDictionary<string, string> Tags { get; init; } = []
}
```

src/Grpc/orchestrator_service.proto
```diff
@@ -75,6 +75,7 @@ message ExecutionStartedEvent {
     google.protobuf.Timestamp scheduledStartTimestamp = 6;
     TraceContext parentTraceContext = 7;
     google.protobuf.StringValue orchestrationSpanID = 8;
+    map<string, string> tags = 9;
```
cgillum, line 78: Protobuf changes need to be made first in `microsoft/durabletask-protobuf` and then synchronized to this repo using the PowerShell script in this directory.

src/Client/Core/OrchestrationMetadata.cs
```diff
@@ -1,233 +1,238 @@
...
+    /// <summary>
+    /// Gets the orchestration instance's serialized custom status, if any, as a string value.
+    /// </summary>
+    /// <value>The serialized custom status or <c>null</c>.</value>
+    public string? SerializedCustomStatus { get; init; }
+
+    /// <summary>
+    /// Gets the tags associated with the orchestration instance.
+    /// </summary>
+    public IDictionary<string, string>? Tags { get; init; }
```
cgillum, line 85: To make this a bit more ergonomic, consider making this a non-null read-only dictionary initialized to an empty value by default.
```suggestion
    public IReadOnlyDictionary<string, string> Tags { get; init; } = [];
```

src/Grpc/orchestrator_service.proto
torosent: fixed

src/Abstractions/TaskOptions.cs
torosent: fixed

src/Abstractions/TaskOptions.cs
```diff
@@ -1,103 +1,109 @@
...
+/// <param name="StartAt">
+/// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
+/// is specified, the orchestration instance will be scheduled immediately.
+/// </param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null)
+{
+    /// <summary>
+    /// Gets or sets the tags to associate with the orchestration instance. Tags are key-value pairs that can be used.
+    /// </summary>
+    public IDictionary<string, string>? Tags { get; init; } = new Dictionary<string, string>();
```
cgillum, line 108: 1. Remove the `?` - it's not needed now that you're assigning a default value.
2. Remove the `init`


```suggestion
    public IDictionary<string, string> Tags { get; } = new Dictionary<string, string>();
```

src/Client/Core/OrchestrationMetadata.cs
```diff
@@ -1,233 +1,238 @@
...
+    /// <summary>
+    /// Gets the orchestration instance's serialized custom status, if any, as a string value.
+    /// </summary>
+    /// <value>The serialized custom status or <c>null</c>.</value>
+    public string? SerializedCustomStatus { get; init; }
+
+    /// <summary>
+    /// Gets the tags associated with the orchestration instance.
+    /// </summary>
+    public IReadOnlyDictionary<string, string> Tags { get; init; } = new Dictionary<string, string>();
```
cgillum, line 85: Remove the `init` since we're providing the underlying dictionary.
```suggestion
    public IReadOnlyDictionary<string, string> Tags { get; } = new Dictionary<string, string>();
```

src/Abstractions/TaskOptions.cs
jviau, line 108: I love exposing tags on this API, but one of the goals of record types is immutability. Adding an `IDictionary` will undo that as it is immutable. Can we retain the immutability aspect? Additionally, I would like to avoid allocating a dictionary if no tags are provided.

Some recommendations:

1. Take in `IEnumerable<KeyValuePair<string, string>>` in the constructor (this follows the collections best practice of accept the lowest contract possible. Also make `Tags` this type.
2. Add new methods `AddTags` to allow for appending tags to a _new_ instance (does not affect current instance)

You can consider using `ImmutableDictionary` as the implementation type under the hood, but avoid exposing it directly.

src/Abstractions/TaskOptions.cs
```diff
@@ -1,103 +1,109 @@
...
+    /// <param name="instanceId">The instance ID to use.</param>
+    /// <returns>A new <see cref="SubOrchestrationOptions" />.</returns>
+    public SubOrchestrationOptions WithInstanceId(string instanceId) => new(this, instanceId);
+}
+
+/// <summary>
+/// Options that can be used to control the behavior of orchestrator task execution. This derived type can be used to
+/// supply extra options for orchestrations.
+/// </summary>
+public record SubOrchestrationOptions : TaskOptions
```
jviau, line 59: Can we add tags to sub orchestration options as well?

jviau (changes requested):
@torosent I think you have something with your git config causing newline changes in many files. Might be this? https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

src/Abstractions/TaskOptions.cs
torosent: We don't support it in the backend yet

src/Client/Core/OrchestrationMetadata.cs
torosent: Fixed but I can't use IReadOnlyDictionary

src/Abstractions/TaskOptions.cs
torosent: Fixed

torosent: Fixed

src/Client/Core/OrchestrationMetadata.cs
torosent: fixed

src/Abstractions/TaskOptions.cs
```diff
@@ -1,103 +1,143 @@
...
+/// <param name="StartAt">
+/// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
+/// is specified, the orchestration instance will be scheduled immediately.
+/// </param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null)
+{
+    /// <summary>
+    /// Gets the tags to associate with the orchestration instance. Tags are key-value pairs that can be used.
+    /// </summary>
+    public IEnumerable<KeyValuePair<string, string>> Tags { get; private set; } = System.Collections.Immutable.ImmutableDictionary.Create<string, string>();
```
cgillum, line 108: I feel bad that we're doing so much back and forth on this, but the current implementation isn't quite right. Please change it as follows:

```suggestion
    public IReadOnlyDictionary<string, string> Tags { get; init; } = ImmutableDictionary.Create<string, string>();
```

Explanation: We can't use `IEnumerable<KeyValuePair<string, string>>`, which allows multiple keys, because the underlying protobuf type is a `map<string, string>` *doesn't* support duplicate keys. An `I[ReadOnly]Dictionary<string, string>` is therefore more appropriate.

Also, I changed `private set` to `init`. Someone who wants to assign tags should just do so on creation (via `init`) rather than calling a method to add tags (the `AddTag*` methods defeat the purpose of using immutable collection here). So please remove the methods below. We only need to expose this one property.

src/Client/Core/OrchestrationMetadata.cs
```diff
@@ -1,233 +1,238 @@
...
+    /// <summary>
+    /// Gets the orchestration instance's serialized custom status, if any, as a string value.
+    /// </summary>
+    /// <value>The serialized custom status or <c>null</c>.</value>
+    public string? SerializedCustomStatus { get; init; }
+
+    /// <summary>
+    /// Gets the tags associated with the orchestration instance.
+    /// </summary>
+    public IReadOnlyDictionary<string, string> Tags { get; init; }
```
cgillum, line 85:
```suggestion
    public IReadOnlyDictionary<string, string> Tags { get; init; } = ImmutableDictionary.Create<string, string>()
```

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -1,523 +1,524 @@
...
+                    .Select(group => $"{group.Key} x{group.Count()}"));
+            }
+        }
+
+        async ValueTask<OrchestrationRuntimeState> BuildRuntimeStateAsync(
+            P.OrchestratorRequest orchestratorRequest,
+            ProtoUtils.EntityConversionState? entityConversionState,
+            CancellationToken cancellation)
+        {
+
```
cgillum, line 115: Can this whitespace-only change be reverted?

cgillum (changes requested):

src/Abstractions/TaskOptions.cs
```diff
@@ -1,103 +1,111 @@
...
+/// The unique ID of the orchestration instance to schedule. If not specified, a new GUID value is used.
+/// </param>
+/// <param name="StartAt">
+/// The time when the orchestration instance should start executing. If not specified or if a date-time in the past
+/// is specified, the orchestration instance will be scheduled immediately.
+/// </param>
+public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffset? StartAt = null)
+{
+    /// <summary>
+    /// Gets the tags to associate with the orchestration instance. Tags are key-value pairs that can be used.
```
cgillum, line 108: Looks like the second sentence is incomplete. Can be used for what?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -1,523 +1,523 @@
...
-
-            P.EntityBatchResult response = batchResult.ToEntityBatchResult(
-                completionToken,
-                operationInfos?.Take(batchResult.Results?.Count ?? 0));
-
-            await this.client.CompleteEntityTaskAsync(response, cancellationToken: cancellation);
-        }
-    }
-}
+// Copyright (c) Microsoft Corporation.
```
cgillum, line 1: Are there any changes to this file? If not, can it be reverted and removed from the PR?

test/Grpc.IntegrationTests/GrpcSidecarFixture.cs
```diff
@@ -1,67 +1,67 @@
...
-    public GrpcChannel Channel { get; }
-
-    public void Dispose()
-    {
-        this.Channel.ShutdownAsync().GetAwaiter().GetResult();
-        this.host.StopAsync().GetAwaiter().GetResult();
-        this.host.Dispose();
-    }
-}
+﻿// Copyright (c) Microsoft Corporation.
```
cgillum, line 1: Are there any changes to this file? If not, can it be reverted and removed from this PR?

test/Grpc.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -1,119 +1,119 @@
...
-                await this.host.StopAsync(this.cancellation);
-            }
-            finally
-            {
-                this.host.Dispose();
-            }
-        }
-    }
-}
+﻿// Copyright (c) Microsoft Corporation.
```
cgillum, line 1: Are there any changes to this file? If not, can it be reverted and removed from this PR?

cgillum (commented):
Code changes look good. A few more things to clean up before we merge.

Also, can you please add an entry for this PR to the CHANGELOG.md file?

test/Grpc.IntegrationTests/IntegrationTestBase.cs
torosent: fixed

test/Grpc.IntegrationTests/GrpcSidecarFixture.cs
torosent: fixed

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
torosent: fixed

src/Abstractions/TaskOptions.cs
torosent: fixed

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
torosent: fixed

src/Abstractions/TaskOptions.cs
jviau: These types aren't limited to DTS though

cgillum: Agreed we should add this option for sub-orchestrations. Created the following issue to track: https://github.com/microsoft/durabletask-dotnet/issues/399

PR #398, opened 2025-03-19
Add Exception Type `SubOrchestrationFailedException` 

nytian:
This PR partially solves issue https://github.com/Azure/azure-functions-durable-extension/issues/2689

This PR adds a new exception type `Microsoft.DurableTask.SubOrchestrationFailedException`. Previously, `SubOrchestrationFailedException` was included with the `TaskFailedException`, and it's now handled separately with this new type.

For example, if cx want custom handling logic specifically for suborchestration failures, they should write a catch block for SubOrchestrationFailedException instead of the broader TaskFailedException.

```csharp
try
{
    await ctx.CallActivityAsync("myactivity");
    await ctx.CallSubOrchestratorAsync("mysuborch");
}
catch (SubOrchestrationFailedException ex)
{
    // Handle suborchestration-specific exceptions here
}
catch (TaskFailedException ex)
{
    // Handle other task failures here, including Activities, non-deterministic etc
}
```

### DIFF
--- a/src/Abstractions/TaskFailureDetails.cs
+++ b/src/Abstractions/TaskFailureDetails.cs
@@ -5,7 +5,6 @@ using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using CoreFailureDetails = DurableTask.Core.FailureDetails;
 using CoreOrchestrationException = DurableTask.Core.Exceptions.OrchestrationException;
-using CoreSubOrchestrationFailedException = DurableTask.Core.Exceptions.SubOrchestrationFailedException;
 
 namespace Microsoft.DurableTask;
 
@@ -157,15 +156,6 @@ public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? 
 
         if (exception is CoreOrchestrationException coreEx)
         {
-            if (coreEx is CoreSubOrchestrationFailedException)
-            {
-                return new TaskFailureDetails(
-                    coreEx.GetType().ToString(),
-                    coreEx.Message,
-                    coreEx.StackTrace,
-                    FromCoreFailureDetailsRecursive(coreEx.FailureDetails?.InnerFailure) ?? FromExceptionRecursive(coreEx.InnerException));
-            }
-
             return new TaskFailureDetails(
                 coreEx.FailureDetails?.ErrorType ?? "(unknown)",
                 coreEx.FailureDetails?.ErrorMessage ?? "(unknown)",
--- a/src/Abstractions/TaskFailureDetails.cs
+++ b/src/Abstractions/TaskFailureDetails.cs
@@ -155,8 +155,17 @@ public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? 
             return null;
         }
 
-        if (exception is CoreOrchestrationException coreEx && exception is not CoreSubOrchestrationFailedException)
+        if (exception is CoreOrchestrationException coreEx)
         {
+            if (coreEx is CoreSubOrchestrationFailedException)
+            {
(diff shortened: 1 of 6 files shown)


src/Abstractions/TaskFailureDetails.cs
```diff
@@ -156,6 +157,15 @@ internal CoreFailureDetails ToCoreFailureDetails()
 
         if (exception is CoreOrchestrationException coreEx)
         {
+            if (coreEx is CoreSubOrchestrationFailedException)
+            {
+                return new TaskFailureDetails(
+                    coreEx.GetType().ToString(),
+                    coreEx.Message,
+                    coreEx.StackTrace,
+                    FromCoreFailureDetailsRecursive(coreEx.FailureDetails?.InnerFailure) ?? FromExceptionRecursive(coreEx.InnerException));
```
andystaples, line 166: Should this be
FromCoreFailureDetailsRecursive(coreEx.FailureDetails) ?? FromExceptionRecursive(coreEx.InnerException))
To prevent loss of the outer FailureDetails?

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
```diff
@@ -543,8 +545,8 @@ static void ValidateInnermostFailureDetailsChain(TaskFailureDetails? failureDeta
                     {
                         // Outer failure represents the orchestration failure
                         Assert.NotNull(ex.FailureDetails);
-                        Assert.True(ex.FailureDetails.IsCausedBy<TaskFailedException>());
-                        Assert.Contains("ThrowException", ex.FailureDetails.ErrorMessage);
```
andystaples, line 547: This change worries me - our tests were checking that the activity name was present in the exception message, which I think is important. Even if this change means that this information is in some InnerFailure, can we still make sure that this information is present somewhere in the exception details?

andystaples (commented):
Left some comments

test/Grpc.IntegrationTests/OrchestrationErrorHandling.cs
```diff
@@ -543,8 +545,8 @@ static void ValidateInnermostFailureDetailsChain(TaskFailureDetails? failureDeta
...
+                        Assert.True(ex.FailureDetails.IsCausedBy<CoreSubOrchestrationFailedException>());
```
cgillum, line 548: We don't want to expose any DurableTask.Core types in the .NET Isolated model. Also, this is a breaking change so I think we need to rethink this.

nytian: yeah what I’m mainly worried about with this change is that it looks like we don’t want to expose any core exception types here, though I’m not sure why. 
I’m totally fine with closing this. I made this change mainly because it’s what the issue/customer was asking for and with this change it seemed more in line with our in-proc model. But it seems like we just want to show that the sub-orchestrator failed in the error message, not the actual error type, is that correct?

jviau (changes requested):
We shouldn't expose any DurableTask.Core types to the user in this repo, we want to keep that purely an implementation detail. Please derive a new exception type in this repo and construct that from the `SubOrchestrationFailedException` instead

nytian:
After further discussing and consideration, this PR should be closed. The main reason is that our current approach—throwing `TaskFailedException` for all failed tasks—is consistent with the existing in-proc model, which always throws `FunctionsFailedException`. Changing this behavior might unexpectedly break errors handling of all Durable Functions. Also, detailed information about task failures or the specific functions that threw exceptions can already be obtained from the error message itself, and does not necessarily require differentiating exception types.

PR #400, opened 2025-03-24
Partial Purge Support

YunchuWang:
This pull request add portable sdk support for partial purge to dts backend.  PurgeInstancesResponse will return an optional flag "Iscomplete" to indicate whether purge is fully done or done partially.

Enhancements to `PurgeResult` class and related client updates:

* [`src/Client/Core/PurgeResult.cs`](diffhunk://#diff-125be6ccd02823bcbe25929c5cf6cdd7cac4b8263995f6ec49525015c2fde7c1L15-R33): Added a new `isComplete` parameter to the `PurgeResult` constructor and a corresponding `IsComplete` property to indicate whether the purge operation is complete.
* [`src/Client/Grpc/GrpcDurableTaskClient.cs`](diffhunk://#diff-1cbc56f55c712147e9731872d3513aebfdd938d74874d59ab6ddea60e75586b4L449-R449): Updated the `PurgeInstancesCoreAsync` method to pass the `IsComplete` value from the response to the `PurgeResult` constructor.
* [`src/Grpc/orchestrator_service.proto`](diffhunk://#diff-1da44a507b5d92ecb779a36d1bffcf6974159609ba30a0470958b0250c7782abR466): Added a new `isComplete` field to the `PurgeInstancesResponse` message to support the new `IsComplete` property in the `PurgeResult` class.

### DIFF
--- a/src/Client/Core/PurgeResult.cs
+++ b/src/Client/Core/PurgeResult.cs
@@ -12,11 +12,23 @@ public class PurgeResult
     /// Initializes a new instance of the <see cref="PurgeResult" /> class.
     /// </summary>
     /// <param name="count">The count of instances purged.</param>
-    /// <param name="isComplete">A value indicating whether the purge operation is complete. Default is null.</param>
-    public PurgeResult(int count, bool? isComplete = null)
+    public PurgeResult(int count)
     {
         Check.Argument(count >= 0, nameof(count), "Count must be non-negative");
         this.PurgedInstanceCount = count;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PurgeResult" /> class.
+    /// </summary>
+    /// <param name="count">The number of instances deleted.</param>
+    /// <param name="isComplete">A value indicating whether the purge operation is complete. 
+    /// If true, the purge operation is complete. All instances were purged.
+    /// If false, not all instances were purged. Please purge again.
+    /// If null, default to legacy purge behavior. All instances were purged.</param>
+    public PurgeResult(int count, bool? isComplete)
+        : this(count)
+    {
         this.IsComplete = isComplete;
     }
 
--- a/src/Client/Core/PurgeResult.cs
+++ b/src/Client/Core/PurgeResult.cs
@@ -22,7 +22,7 @@ public class PurgeResult
     /// Initializes a new instance of the <see cref="PurgeResult" /> class.
     /// </summary>
     /// <param name="count">The number of instances deleted.</param>
-    /// <param name="isComplete">A value indicating whether the purge operation is complete. 
+    /// <param name="isComplete">A value indicating whether the purge operation is complete.
     /// If true, the purge operation is complete. All instances were purged.
     /// If false, not all instances were purged. Please purge again.
     /// If null, whether or not all instances were purged is undefined.</param>
--- a/src/Client/Core/PurgeResult.cs
+++ b/src/Client/Core/PurgeResult.cs
@@ -25,7 +25,7 @@ public class PurgeResult
     /// <param name="isComplete">A value indicating whether the purge operation is complete. 
     /// If true, the purge operation is complete. All instances were purged.
     /// If false, not all instances were purged. Please purge again.
-    /// If null, default to legacy purge behavior. All instances were purged.</param>
+    /// If null, whether or not all instances were purged is undefined.</param>
     public PurgeResult(int count, bool? isComplete)
         : this(count)
     {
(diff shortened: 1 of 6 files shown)


src/Client/Core/PurgeResult.cs
```diff
@@ -12,15 +12,23 @@ public class PurgeResult
     /// Initializes a new instance of the <see cref="PurgeResult" /> class.
     /// </summary>
     /// <param name="count">The count of instances purged.</param>
-    public PurgeResult(int count)
+    /// <param name="isComplete">A value indicating whether the purge operation is complete. Default is null.</param>
+    public PurgeResult(int count, bool? isComplete = null)
```
cgillum, line 16: Same comment about this being a binary-breaking change. Let's create a second constructor overload that adds a non-nullable `bool isComplete` parameter.

test/Client/OrchestrationServiceClientShim.Tests/ShimDurableTaskClientTests.cs
```diff
@@ -174,6 +174,7 @@ public async Task PurgeInstanceMetadata()
         // assert
         this.orchestrationClient.VerifyAll();
         result.PurgedInstanceCount.Should().Be(1);
+        result.IsComplete.Should().BeNull();
```
cgillum, line 177: I would change this to assert that `IsComplete` should not be false. It's correct for it to be either null or true, depending on the backend implementation.

test/Grpc.IntegrationTests/GrpcDurableTaskClientIntegrationTests.cs
```diff
@@ -150,6 +150,80 @@ await server.Client.ScheduleNewOrchestrationInstanceAsync(
...
+
+        // Act
+        PurgeResult result = await server.Client.PurgeInstanceAsync(
+            instanceId,
+            new PurgeInstanceOptions { Recursive = true });
+
+        // Assert
+        result.Should().NotBeNull();
+        result.PurgedInstanceCount.Should().Be(1);
+        result.IsComplete.Should().BeNull();
```
cgillum, line 182: Same comment that this should be either null or true, so maybe we should assert that it's not false.

cgillum: Same comment for the other use in the next test method.

YunchuWang: updated

src/Client/Core/PurgeResult.cs
```diff
@@ -18,9 +18,29 @@ public PurgeResult(int count)
...
     /// Gets the number of purged instances.
     /// </summary>
     /// <value>The number of purged instances.</value>
     public int PurgedInstanceCount { get; }
+
+    /// <summary>
+    /// Gets a value indicating whether the purge operation is complete.
+    /// </summary>
+    /// <value>A value indicating whether the purge operation is complete.</value>
+    public bool? IsComplete { get; }
```
jviau, line 45: The comments indicate you are using `null` to indicate legacy behavior. Considering this is the same as `true`, do we need `null`? Or can we default this to `true`?

jviau (commented):
Please address the build warnings introduced by your changes.

src/Client/Core/PurgeResult.cs
jviau, line 45: Do we actually need to expose this to the caller? We could leave it part of the gRPC contract only, and the implementation of purge will make repeated calls until purge is complete.

YunchuWang: i think it is needed for customers who want to do partial purges themselves (ex. do partial purges following their own mechanism, adding custom loggings, quit early, etc)

YunchuWang: i can update to add purgeallinstances to do loop internally as alternative approach

YunchuWang: good question, i think both makes sense, null preserve the semantic of old and new versions better but adds more complexity for client to handle, @cgillum for feedback on this

cgillum: See my comment on the corresponding DTFx PR. It's best to document null as being "undefined" in terms of whether all instances got deleted since it depends on the implementation of the backend (for example, it's true for Azure Storage, but not necessarily for MSSQL).

cgillum: Depending on the backend, purge can be prohibitively expensive, so I don't think we should automate this loop.

PR #401, opened 2025-03-25
Introduce orchestration versioning into worker

halspang:
This commit gives the worker several ways to configure versioning. Included are, setting the version, how to match against incoming versions, and how to react to a failed match.

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -399,19 +399,6 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
         throw new NotSupportedException($"{this.GetType()} does not support purging of orchestration instances.");
     }
 
-    /// <summary>
-    /// Abandons a task orchestration work item to be re-enqueued at a later time.
-    /// </summary>
-    /// <param name="completionToken">The completion token that was given when the WorkItem was fetched.</param>
-    /// <param name="cancellation">
-    /// A <see cref="CancellationToken"/> that can be used to cancel the purge operation.
-    /// </param>
-    /// <returns>A task that completes when the WorkItem has been abandoned.</returns>
-    public virtual Task AbandonOrchestrationTask(string completionToken, CancellationToken cancellation = default)
-    {
-        throw new NotSupportedException($"{this.GetType()} does not support abandoning task orchestration work items.");
-    }
-
     // TODO: Create task hub
 
     // TODO: Delete task hub
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -399,6 +399,19 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
         throw new NotSupportedException($"{this.GetType()} does not support purging of orchestration instances.");
     }
 
+    /// <summary>
+    /// Abandons a task orchestration work item to be re-enqueued at a later time.
+    /// </summary>
+    /// <param name="completionToken">The completion token that was given when the WorkItem was fetched.</param>
+    /// <param name="cancellation">
+    /// A <see cref="CancellationToken"/> that can be used to cancel the purge operation.
+    /// </param>
+    /// <returns>A task that completes when the WorkItem has been abandoned.</returns>
+    public virtual Task AbandonOrchestrationTask(string completionToken, CancellationToken cancellation = default)
+    {
+        throw new NotSupportedException($"{this.GetType()} does not support abandoning task orchestration work items.");
+    }
+
     // TODO: Create task hub
 
     // TODO: Delete task hub
--- a/src/Worker/Core/DurableTaskWorkerOptions.cs
+++ b/src/Worker/Core/DurableTaskWorkerOptions.cs
@@ -10,7 +10,50 @@ namespace Microsoft.DurableTask.Worker;
 /// </summary>
 public class DurableTaskWorkerOptions
 {
+    readonly VersioningOptions versioning = new();
     DataConverter dataConverter = JsonDataConverter.Default;
+
+    /// <summary>
+    /// Defines the version matching strategy for the Durable Task worker.
+    /// </summary>
+    public enum VersionMatchStrategy
+    {
+        /// <summary>
+        /// Ignore Orchestration version, all work received is processed.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Worker will only process Tasks from Orchestrations with the same version as the worker.
+        /// </summary>
+        Strict = 1,
+
+        /// <summary>
+        /// Worker will process Tasks from Orchestrations whose version is less than or equal to the worker.
+        /// </summary>
+        CurrentOrOlder = 2,
+    }
+
+    /// <summary>
+    /// Defines the versioning failure strategy for the Durable Task worker.
+    /// </summary>
+    public enum VersionFailureStrategy
+    {
+        /// <summary>
+        /// Do not change the orchestration state if the version does not adhere to the matching strategy.
+        /// </summary>
+        Reject = 0,
+
+        /// <summary>
+        /// Suspend the orchestration if the version does not adhere to the matching strategy.
+        /// </summary>
+        Suspend = 1,
+
+        /// <summary>
+        /// Fail the orchestration if the version does not adhere to the matching strategy.
+        /// </summary>
+        Fail = 2,
+    }
 
     /// <summary>
     /// Gets or sets the data converter. Default value is <see cref="JsonDataConverter.Default" />.
@@ -94,6 +137,38 @@ public class DurableTaskWorkerOptions
     public ConcurrencyOptions Concurrency { get; } = new();
 
     /// <summary>
+    /// Gets or sets the versioning options for the Durable Task worker.
+    /// </summary>
+    /// <remarks>
+    /// Worker versioning controls how a worker will handle orchestrations of different versions. Defining both the
+    /// version of the worker, the versions that can be worked on, and what to do in case a version does not comply
+    /// with the given options.
+    /// </remarks>
+    public VersioningOptions Versioning
+    {
+        get => this.versioning;
+        set
+        {
+            if (value is not null)
+            {
+                this.IsVersioningSet = true;
+                this.Versioning.Version = value.Version;
+                this.Versioning.MatchStrategy = value.MatchStrategy;
+                this.Versioning.FailureStrategy = value.FailureStrategy;
+            }
+            else
+            {
+                this.IsVersioningSet = false;
+            }
+        }
+    }
+
+    /// <summary>
+    /// Gets a value indicating whether versioning is explicitly set or not.
+    /// </summary>
+    public bool IsVersioningSet { get; internal set; }
+
+    /// <summary>
     /// Gets a value indicating whether <see cref="DataConverter" /> was explicitly set or not.
     /// </summary>
     /// <remarks>
@@ -116,6 +191,7 @@ public class DurableTaskWorkerOptions
             other.DataConverter = this.DataConverter;
             other.MaximumTimerInterval = this.MaximumTimerInterval;
             other.EnableEntitySupport = this.EnableEntitySupport;
+            other.Versioning = this.Versioning;
         }
     }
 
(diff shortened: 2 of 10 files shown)


src/Client/Core/DurableTaskClient.cs
```diff
@@ -399,6 +399,19 @@ public virtual Task<PurgeResult> PurgeAllInstancesAsync(
...
 
+    /// <summary>
+    /// Abandons a task orchestration work item to be re-enqueued at a later time.
+    /// </summary>
+    /// <param name="completionToken">The completion token that was given when the WorkItem was fetched.</param>
+    /// <param name="cancellation">
+    /// A <see cref="CancellationToken"/> that can be used to cancel the purge operation.
+    /// </param>
+    /// <returns>A task that completes when the WorkItem has been abandoned.</returns>
+    public virtual Task AbandonOrchestrationTask(string completionToken, CancellationToken cancellation = default)
```
cgillum, line 410: It doesn't make sense for this kind of API to exist on `DurableTaskClient`. Clients don't interact with work items at all. Only workers do.

src/Worker/Core/DurableTaskWorkerOptions.cs
```diff
@@ -93,6 +136,38 @@ public DataConverter DataConverter
...
+            if (value is not null)
+            {
+                this.IsVersioningSet = true;
+                this.Versioning.Version = value.Version;
+                this.Versioning.MatchStrategy = value.MatchStrategy;
+                this.Versioning.FailureStrategy = value.FailureStrategy;
+            }
+            else
+            {
+                this.IsVersioningSet = false;
```
cgillum, line 161: Rather than having a flag indicating whether versioning is set, wouldn't it be simpler to just make this property nullable, and use the null value to signal whether the property is set? I think it would be a lot simpler, and also give you the benefit of compiler checks everywhere this property gets consumed.

src/Client/Core/DurableTaskClient.cs
halspang: Oh, you're right. I think I added this during my experimentation phase and then forgot to remove it. I don't believe it's being used anywhere, I'll remove it.

src/Worker/Core/DurableTaskWorkerOptions.cs
halspang: I was following the style that was already present in the options: https://github.com/microsoft/durabletask-dotnet/blob/e32fa7a4ff9e569120957299508a58e88c115398/src/Worker/Core/DurableTaskWorkerOptions.cs#L105

If you prefer null I can do that instead.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -361,6 +415,50 @@ async Task OnRunOrchestratorAsync(
...
+                                    OrchestrationStatus = P.OrchestrationStatus.Failed,
+                                    FailureDetails = failureDetails,
+                                },
+                            },
+                        },
+                    };
+                }
+                else if (this.worker.workerOptions.Versioning.FailureStrategy == DurableTaskWorkerOptions.VersionFailureStrategy.Suspend)
+                {
+                    await this.client.SuspendInstanceAsync(
```
cgillum, line 441: Using client commands from a worker like this is problematic on many levels. A few that I can think of off the top of my head:

- From an RBAC perspective, client operations and worker operations are expected to be treated as distinct, and this change would cause workers to sometimes operate as clients. This for sure won't work well with the DTS security model.
- From a consistency perspective, client operations like this are asynchronous, requiring another round-trip from the worker, so even though we're suspending it here, it won't actually be suspended until the worker processes the resulting "suspend yourself" work item, allowing for further processing to happen in the meantime.

The right way to suspend an orchestration from a worker is to use the `CompleteOrchestratorTaskAsync` API and include an action that transitions the orchestration into the suspended state. I don't think we actually have anything like this today, so we may need to introduce this as a new concept into DTFx itself and then surface it up through the other layers.

cgillum (changes requested):
A couple of blocking issues in this PR that may require a bit of offline discussion...

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -305,43 +306,96 @@ async Task OnRunOrchestratorAsync(
...
+                                    ErrorType = "VersionMismatch",
+                                    ErrorMessage = $"The orchestration version '{runtimeState.Version}' does not match the worker version '{this.worker.grpcOptions.Versioning.Version}'.",
+                                    IsNonRetriable = true,
+                                };
+                            }
+
+                            break;
+                        case DurableTaskWorkerOptions.VersionMatchStrategy.CurrentOrOlder:
+                            // Comparison > 0 indicates the orchestration version is greater than the worker version.
+                            if (versionComparison > 0)
```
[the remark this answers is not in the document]
halspang: This is basically a passthrough to the standard `CompareTo` methods which return in relation to the first argument, not the second. At least that's what my experimentation showed.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -361,6 +415,50 @@ async Task OnRunOrchestratorAsync(
                     completionToken,
                     entityConversionState);
             }
+            else if (failureDetails != null && failureDetails.ErrorType == "VersionMismatch")
```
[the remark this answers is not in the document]
halspang: I can add in a boolean instead of relying on the error type.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: The `CompleteOrchestratorTaskAsync` is also a client call though, no? Or are you saying that on the backend, it is handled differently than this would be? Given that we already had a client present in this class, it seemed safe to use. If it's not safe, we should document around it (if there is some and I missed it, please let me know) or we should separate the APIs that are safe/not-safe into different clients. Having them combined under one proto feels dangerous.

Is this the same for `Abandon`? Or can we just drop `Suspend` for now and implement that behavior without blocking this feature?

halspang: After speaking in person, I understand that `Suspend` is a very different operation than I thought it was. Removed for now and we can come back to it later if/when it's asked for and after we've assured `Suspend` works in a way that is compatible to this feature.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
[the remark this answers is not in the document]
cgillum: I double checked and I was wrong. What you have is good.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -298,50 +299,107 @@ async Task OnRunOrchestratorAsync(
...
-                    TaskOrchestration shim = this.shimFactory.CreateOrchestration(name, orchestrator, parent);
-                    TaskOrchestrationExecutor executor = new(
-                        runtimeState,
-                        shim,
-                        BehaviorOnContinueAsNew.Carryover,
-                        request.EntityParameters.ToCore(),
-                        ErrorPropagationMode.UseFailureDetails);
-                    result = executor.Execute();
+                    versionFailure = failureDetails != null;
                 }
```
cgillum, line 360: Non-blocking: Would it be easy to move this version checking logic into a helper method? This method is getting kind of long and harder to reason about the end-to-end flow. Breaking this logic out into a helper method would make the outer `OnRunOrchestratorAsync` method a little easier for humans to parse.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -361,6 +419,39 @@ async Task OnRunOrchestratorAsync(
...
+                                    OrchestrationStatus = P.OrchestrationStatus.Failed,
+                                    FailureDetails = failureDetails,
+                                },
+                            },
+                        },
+                    };
+                }
+                else
+                {
+                    await this.client.AbandonTaskOrchestratorWorkItemAsync(
```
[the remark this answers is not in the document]
halspang: Added logging for this and the general failure case.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: That's fair, moved the versioning check out.

src/Worker/Grpc/Logs.cs
```diff
@@ -51,5 +51,11 @@ static partial class Logs
 
         [LoggerMessage(EventId = 56, Level = LogLevel.Warning, Message = "Channel to backend has stopped receiving traffic, will attempt to reconnect.")]
         public static partial void ConnectionTimeout(this ILogger logger);
+
+        [LoggerMessage(EventId = 57, Level = LogLevel.Warning, Message = "Orchestration version did not meet worker versioning requirements. Error action = '{errorAction}'. Version error = '{versionError}'")]
+        public static partial void OrchestrationVersionFailure(this ILogger logger, string errorAction, string versionError);
+
+        [LoggerMessage(EventId = 58, Level = LogLevel.Information, Message = "Abandoning orchestration due to versioning mismatch.")]
+        public static partial void AbandoningOrchestrationDueToVersioning(this ILogger logger);
```
cgillum, line 59: Let's make this a generic abandon log message. We already know from the previous warning message why it was abandoned. Also, lets include details like the work item completion token so that we can correlate this log message to other log messages associated with this work item.

halspang: Added in the completion token and the instance id.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -69,562 +69,705 @@ public async Task ScheduleOrchesrationWithTags()
...
+                    return ctx.CallActivityAsync<string>("Versioned_Activity", ctx.Version);
+                })
+                .AddActivityFunc<string, string>("Versioned_Activity", (ctx, input) =>
+                {
+                    return $"Orchestration version: {input}";
+                }));
+            b.UseVersioning(new()
+            {
+                Version = workerVersion,
+                MatchStrategy = DurableTaskWorkerOptions.VersionMatchStrategy.None,
```
cgillum, line 645: Do we have a test validating that this is the default strategy? If not, I wonder if we should remove this line to confirm that it is indeed the default in terms of the runtime behavior.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -69,562 +69,705 @@ public async Task ScheduleOrchesrationWithTags()
...
+    public async Task OrchestrationVersioning_MatchTypeNone_NoVersionFailure()
+    {
+        var workerVersion = "0.1";
+        var clientVersion = "0.2";
+        await using HostTestLifetime server = await this.StartWorkerAsync(b =>
+        {
+            b.AddTasks(tasks => tasks
+                .AddOrchestratorFunc<string, string>("Versioned_Orchestration", (ctx, input) =>
+                {
+                    return ctx.CallActivityAsync<string>("Versioned_Activity", ctx.Version);
```
cgillum, line 636: Pro-tip: you don't need to call an activity - you can just do the following:

```suggestion
                    return Task.FromResult(ctx.Version);
```

...not that there's anything wrong with what you're doing, but I thought I'd mention it if you wanted to remove a few lines of code from these tests.

cgillum (approved):
Looks good. Just a couple non-blocking comments on the test code.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: I have copied this test and removed the default just so we can test it with nothing and with `None` :)

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: Good to know! I'll leave them as is for now as I don't mind some verbosity in a test.

cgillum (approved):

PR #404, opened 2025-03-28
Distributing Tracing for Entities (Isolated)

sophiatev:
This PR adds support for distributed tracing for entities in the .NET isolated framework. The actual tracing objects are created in the durabletask repo (see this [PR]()) but this repo is where the requests to entities are generated and executed in the isolated case. The changes in this PR are then mostly to propagate information correctly to the durabletask repo for the traces (for example, the time that a signal or call request to an entity was made, or the time that the request was accomplished, any associated error messages, etc.) Related to this propagation of information, the definition of proto objects related to signaling entities or entities starting orchestrations had to be populated with extra information (like the parent trace context of the request, or the time of the request, etc. For more details see this [PR](https://github.com/microsoft/durabletask-protobuf/pull/40) in the protobuf repo). The changes in the `ProtobufUtils` class focus on adding this extra information (similar changes occur in this [PR](https://github.com/Azure/azure-functions-durable-extension/pull/3076) in the WebJobs repo). 
- The changes in the `GrpcDurableEntityClient` and `ShimDurableEntityClient` classes are to propagate this additional information in the case of a client signaling an entity, 
- In `TaskEntityShim` they are for entities signaling other entities, entities starting orchestrations, and the actual execution of the entity requests, and 
- In `TaskOrchestrationEntityContext` for orchestrations signaling and calling entities

### DIFF
--- a/src/Client/Grpc/GrpcDurableEntityClient.cs
+++ b/src/Client/Grpc/GrpcDurableEntityClient.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics;
 using Microsoft.DurableTask.Client.Entities;
 using Microsoft.DurableTask.Entities;
 using Microsoft.Extensions.Logging;
@@ -56,6 +57,17 @@ class GrpcDurableEntityClient : DurableEntityClient
             Input = this.dataConverter.Serialize(input),
             ScheduledTime = scheduledTime?.ToTimestamp(),
         };
+
+        if (Activity.Current?.Id != null)
+        {
+            if (request.ParentTraceContext == null)
+            {
+                request.ParentTraceContext = new P.TraceContext();
+            }
+
+            request.ParentTraceContext.TraceParent = Activity.Current.Id;
+            request.ParentTraceContext.TraceState = Activity.Current.TraceStateString;
+        }
 
         // TODO this.logger.LogSomething
         try
--- a/src/Client/OrchestrationServiceClientShim/ShimDurableEntityClient.cs
+++ b/src/Client/OrchestrationServiceClientShim/ShimDurableEntityClient.cs
@@ -1,8 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics;
 using DurableTask.Core;
 using DurableTask.Core.Entities;
+using DurableTask.Core.Tracing;
 using Microsoft.DurableTask.Client.Entities;
 using Microsoft.DurableTask.Entities;
 
@@ -90,7 +92,8 @@ class ShimDurableEntityClient(string name, ShimDurableTaskClientOptions options)
             EntityMessageEvent.GetCappedScheduledTime(
                 DateTime.UtcNow,
                 this.options.Entities.MaxSignalDelayTimeOrDefault,
-                scheduledTime?.UtcDateTime));
+                scheduledTime?.UtcDateTime),
+            Activity.Current?.Id != null ? new DistributedTraceContext(Activity.Current.Id, Activity.Current.TraceStateString) : null);
 
         await this.options.Client!.SendTaskOrchestrationMessageAsync(eventToSend.AsTaskMessage());
     }
--- a/src/Grpc/orchestrator_service.proto
+++ b/src/Grpc/orchestrator_service.proto
@@ -75,6 +75,7 @@ message ExecutionStartedEvent {
     google.protobuf.Timestamp scheduledStartTimestamp = 6;
     TraceContext parentTraceContext = 7;
     google.protobuf.StringValue orchestrationSpanID = 8;
+    map<string, string> tags = 9;
 }
 
 message ExecutionCompletedEvent {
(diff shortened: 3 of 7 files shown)


src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -55,7 +56,18 @@ public override async Task SignalEntityAsync(
             Name = operationName,
             Input = this.dataConverter.Serialize(input),
             ScheduledTime = scheduledTime?.ToTimestamp(),
-        };
+        };
+
+        if (Activity.Current?.Id != null)
```
jviau, line 61: I like to use pattern matching:

```suggestion
        if (Activity.Current is { } activity)
```

Then you use `activity` throughout the if block. It is guaranteed non-null there.

jviau (commented):
Do you know where `Activity.Current` will be set per entity operation?

src/Client/Grpc/GrpcDurableEntityClient.cs
sophiatev: But this wouldn't catch if the `Id` is null, right? I don't see how `Activity.Current` could be non-null with a null `Id`, but I figured it's better to be safe rather than sorry and keep this sort of check instead

sophiatev:
> Do you know where `Activity.Current` will be set per entity operation?

This is actually done in DurableTask.Core in this [PR](https://github.com/Azure/durabletask/pull/1198) where the Activities are created (with the exception of a client signaling an entity, for which the `Activity` is created in `GrpcDurableEntityClientSignalEntityAsync.`. This was actually something I was wondering myself - `Activity.Current` seems to be set here already correctly and I use it to set the parent trace context of the signal request, but I'm not actually sure where it actually gets set...

src/Client/Grpc/GrpcDurableEntityClient.cs
jviau: `Activity.Current` won't have a null Id. `Activity.Current` must always be a 'running' activity, which means non-null ID.

https://github.com/dotnet/runtime/blob/4d14dc5b4c685eeca9de610298006f4552e43955/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1412

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
```
bachuv, line 1: Do you know why this is showing up in the diff?

src/Client/OrchestrationServiceClientShim/ShimDurableEntityClient.cs
```diff
@@ -90,7 +92,10 @@ public override async Task SignalEntityAsync(
             EntityMessageEvent.GetCappedScheduledTime(
                 DateTime.UtcNow,
                 this.options.Entities.MaxSignalDelayTimeOrDefault,
-                scheduledTime?.UtcDateTime));
+                scheduledTime?.UtcDateTime),
+            Activity.Current is { } activity ? new DistributedTraceContext(activity.Id!, activity.TraceStateString) : null,
+            DateTimeOffset.UtcNow,
+            createTrace: true);
```
bachuv, line 98: nit: can you indent these lines?

sophiatev: Should they be indented? It looks like they're on the same level as all the other parameters for the `ClientEntityHelpers.EmitOperationSignal` call

src/Client/Grpc/GrpcDurableTaskClient.cs
sophiatev: Unfortunately no, it doesn't seem like a line ending thing so I'm at a loss as to how to get rid of it...

bachuv: Ok, no worries. It should be fine.

src/Client/Grpc/GrpcDurableTaskClient.cs
bachuv, line 1: Can you fix the warnings in this file?

src/Client/Grpc/GrpcDurableEntityClient.cs
```diff
@@ -54,8 +55,16 @@ public override async Task SignalEntityAsync(
...
-        };
+            ScheduledTime = scheduledTime?.ToTimestamp(),
+            RequestTime = DateTimeOffset.UtcNow.ToTimestamp(),
+        };
+
+        if (Activity.Current is { } activity)
+        {
+            request.ParentTraceContext ??= new P.TraceContext();
+            request.ParentTraceContext.TraceParent = activity.Id;
+            request.ParentTraceContext.TraceState = activity.TraceStateString;
```
bachuv, line 66: Should we add a null check for `activity.Id` and `activity.TraceStateString`?

sophiatev: It looks like these are never null if `Activity.Current` is non-null as per the earlier [discussion](https://github.com/microsoft/durabletask-dotnet/pull/404#discussion_r2035874028) we had

bachuv (approved):

bachuv (approved):

PR #405, opened 2025-03-31
Basic orchestration and activity execution logs

cgillum:
We're missing some basic execution logs in this SDK. For Azure Functions, this was okay because the Functions host and worker layer would sufficiently take care of logging. However, for standalone use, the base logging was insufficient.

This PR adds basic execution starting, completion, and failure logs for orchestrations and activities.

### DIFF
--- a/src/Worker/Core/Logs.cs
+++ b/src/Worker/Core/Logs.cs
@@ -36,18 +36,18 @@ namespace Microsoft.DurableTask
         public static partial void ActivityFailed(this ILogger logger, Exception ex, string instanceId, string name);
 
         /// <summary>
-        /// Creates a logger named "Microsoft.DurableTask.Worker" with the specified subcategories.
+        /// Creates a logger named "Microsoft.DurableTask.Worker" with an optional subcategory.
         /// </summary>
         /// <param name="loggerFactory">The logger factory to use to create the logger.</param>
-        /// <param name="subcategories">The subcategory of the logger. For example, "Activities" or "Orchestrations".
+        /// <param name="subcategory">The subcategory of the logger. For example, "Activities" or "Orchestrations".
         /// </param>
         /// <returns>The generated <see cref="ILogger"/>.</returns>
-        internal static ILogger CreateWorkerLogger(ILoggerFactory loggerFactory, string? subcategories = null)
+        internal static ILogger CreateWorkerLogger(ILoggerFactory loggerFactory, string? subcategory = null)
         {
             string categoryName = "Microsoft.DurableTask.Worker";
-            if (!string.IsNullOrEmpty(subcategories))
+            if (!string.IsNullOrEmpty(subcategory))
             {
-                categoryName += "." + subcategories;
+                categoryName += "." + subcategory;
             }
 
             return loggerFactory.CreateLogger(categoryName);
--- a/src/Worker/Core/Logs.cs
+++ b/src/Worker/Core/Logs.cs
@@ -11,10 +11,16 @@ namespace Microsoft.DurableTask
     /// </summary>
     static partial class Logs
     {
-        /// <summary>
-        /// The common category name for worker logs.
-        /// </summary>
-        internal const string WorkerCategoryName = "Microsoft.DurableTask.Worker";
+        internal static ILogger CreateWorkerLogger(ILoggerFactory loggerFactory, params string[] subcategories)
+        {
+            string categoryName = "Microsoft.DurableTask.Worker";
+            if (subcategories.Length > 0)
+            {
+                categoryName += "." + string.Join(".", subcategories);
+            }
+
+            return loggerFactory.CreateLogger(categoryName);
+        }
 
         [LoggerMessage(EventId = 15, Level = LogLevel.Error, Message = "Unhandled exception in entity operation {entityInstanceId}/{operationName}.")]
         public static partial void OperationError(this ILogger logger, Exception ex, EntityInstanceId entityInstanceId, string operationName);
--- a/src/Worker/Core/Logs.cs
+++ b/src/Worker/Core/Logs.cs
@@ -11,17 +11,6 @@ namespace Microsoft.DurableTask
     /// </summary>
     static partial class Logs
     {
-        internal static ILogger CreateWorkerLogger(ILoggerFactory loggerFactory, params string[] subcategories)
-        {
-            string categoryName = "Microsoft.DurableTask.Worker";
-            if (subcategories.Length > 0)
-            {
-                categoryName += "." + string.Join(".", subcategories);
-            }
-
-            return loggerFactory.CreateLogger(categoryName);
-        }
-
         [LoggerMessage(EventId = 15, Level = LogLevel.Error, Message = "Unhandled exception in entity operation {entityInstanceId}/{operationName}.")]
         public static partial void OperationError(this ILogger logger, Exception ex, EntityInstanceId entityInstanceId, string operationName);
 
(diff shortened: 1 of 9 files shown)


philliphoff (approved):

src/Worker/Core/Logs.cs
```diff
@@ -9,15 +9,35 @@ namespace Microsoft.DurableTask
...
     /// </summary>
-    /// <remarks>
-    /// NOTE: Trying to make logs consistent with https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Logging/LogEvents.cs.
-    /// </remarks>
     static partial class Logs
     {
+        /// <summary>
+        /// The common category name for worker logs.
+        /// </summary>
+        internal const string WorkerCategoryName = "Microsoft.DurableTask.Worker";
```
jviau, line 17: Consider using more granular names under this prefix. Such as `Microsoft.DurableTask.Worker.Activities`, `Microsoft.DurableTask.Worker.Orchestrations`

src/Worker/Core/Logs.cs
```diff
@@ -9,15 +9,35 @@ namespace Microsoft.DurableTask
...
         [LoggerMessage(EventId = 55, Level = LogLevel.Information, Message = "{instanceId}: Evaluating custom retry handler for failed '{name}' task. Attempt = {attempt}.")]
         public static partial void RetryingTask(this ILogger logger, string instanceId, string name, int attempt);
+
+        [LoggerMessage(EventId = 600, Level = LogLevel.Information, Message = "'{Name}' orchestration with ID '{InstanceId}' started.")]
+        public static partial void OrchestrationStarted(this ILogger logger, string instanceId, string name);
+
+        [LoggerMessage(EventId = 601, Level = LogLevel.Information, Message = "'{Name}' orchestration with ID '{InstanceId}' completed.")]
+        public static partial void OrchestrationCompleted(this ILogger logger, string instanceId, string name);
+
+        [LoggerMessage(EventId = 602, Level = LogLevel.Information, Message = "'{Name}' orchestration with ID '{InstanceId}' failed.")]
```
jviau, line 31: Should this be error?

src/Worker/Core/Shims/TaskActivityShim.cs
```diff
@@ -34,11 +42,25 @@ public TaskActivityShim(DataConverter dataConverter, TaskName name, ITaskActivit
...
-        string? serializedOutput = this.dataConverter.Serialize(output);
-        return serializedOutput;
+        string instanceId = coreContext.OrchestrationInstance.InstanceId;
+        this.logger.ActivityStarted(instanceId, this.name);
+
+        try
+        {
+            object? output = await this.implementation.RunAsync(contextWrapper, deserializedInput);
+
+            this.logger.ActivityCompleted(instanceId, this.name);
```
jviau, line 53: recommend moving this after output serialization - that could fail and then they would see both a completed & failed message.

src/Worker/Core/Shims/TaskActivityShim.cs
```diff
@@ -11,17 +12,24 @@ namespace Microsoft.DurableTask.Worker.Shims;
...
     /// <param name="name">The name of the activity.</param>
     /// <param name="implementation">The activity implementation to wrap.</param>
-    public TaskActivityShim(DataConverter dataConverter, TaskName name, ITaskActivity implementation)
+    public TaskActivityShim(
+        ILoggerFactory loggerFactory,
+        DataConverter dataConverter,
+        TaskName name,
+        ITaskActivity implementation)
     {
+        this.logger = Check.NotNull(loggerFactory).CreateLogger(Logs.WorkerCategoryName);
```
jviau, line 32: Maybe the category name here should include _Invocation_ or _Execution_ in it somewhere? As well as _Activity_ or _Activities_ or _TaskActivity_.

`Microsoft.DurableTask.Worker.Invocation.TaskActivity`?

src/Worker/Core/Shims/TaskOrchestrationShim.cs
```diff
@@ -31,6 +33,8 @@ public TaskOrchestrationShim(
     {
         this.invocationContext = Check.NotNull(invocationContext);
         this.implementation = Check.NotNull(implementation);
+
+        this.logger = this.invocationContext.LoggerFactory.CreateLogger(Logs.WorkerCategoryName);
```
jviau, line 37: Similar suggestion for category name:

`Microsoft.DurableTask.Worker.Invocation.TaskOrchestrator`

src/Worker/Core/Logs.cs
```diff
@@ -9,15 +9,35 @@ namespace Microsoft.DurableTask
...
+        /// </summary>
+        internal const string WorkerCategoryName = "Microsoft.DurableTask.Worker";
+
         [LoggerMessage(EventId = 15, Level = LogLevel.Error, Message = "Unhandled exception in entity operation {entityInstanceId}/{operationName}.")]
         public static partial void OperationError(this ILogger logger, Exception ex, EntityInstanceId entityInstanceId, string operationName);
 
         [LoggerMessage(EventId = 55, Level = LogLevel.Information, Message = "{instanceId}: Evaluating custom retry handler for failed '{name}' task. Attempt = {attempt}.")]
         public static partial void RetryingTask(this ILogger logger, string instanceId, string name, int attempt);
+
+        [LoggerMessage(EventId = 600, Level = LogLevel.Information, Message = "'{Name}' orchestration with ID '{InstanceId}' started.")]
```
jviau, line 25: Maybe `Debug` level for all of the non-failure ones?

jviau (approved):
Good to have more logging. Do we have plans for distributed tracing in this library directly? These specific logs are more valuable as spans, as that will capture all of the same information but in a more usable way.

No harm in having these logs as well though - but I suggest putting the log level to `Debug` as if spans are ever added, these would be lower importance.

src/Worker/Core/Logs.cs
cgillum: I go back and forth on this one. It might be an error from the app's perspective, but it's not an error from the framework's perspective.

src/Worker/Core/Logs.cs
cgillum: I think of these logs as being similar to API logs, where each request gets logged at an informational level. They feel a bit more audit-like to me and thus debug would be too restrictive.

src/Worker/Core/Shims/TaskActivityShim.cs
cgillum: I went with simply `Microsoft.DurableTask.Worker.Activities` per the earlier suggestion. I don't see a need to distinguish between logs related to invocations and those which aren't.

src/Worker/Core/Shims/TaskOrchestrationShim.cs
cgillum: I went with simply Microsoft.DurableTask.Worker.Orchestrations per the earlier suggestion for the reason mentioned previously.

src/Worker/Core/Logs.cs
```diff
@@ -11,10 +11,16 @@ namespace Microsoft.DurableTask
     /// </summary>
     static partial class Logs
     {
-        /// <summary>
-        /// The common category name for worker logs.
-        /// </summary>
-        internal const string WorkerCategoryName = "Microsoft.DurableTask.Worker";
+        internal static ILogger CreateWorkerLogger(ILoggerFactory loggerFactory, params string[] subcategories)
```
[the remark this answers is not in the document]
cgillum: Fixed

jviau (approved):

src/Worker/Core/Logs.cs
```diff
@@ -9,15 +9,48 @@ namespace Microsoft.DurableTask
...
+        public static partial void ActivityCompleted(this ILogger logger, string instanceId, string name);
+
+        [LoggerMessage(EventId = 605, Level = LogLevel.Information, Message = "'{Name}' activity of orchestration ID '{InstanceId}' failed.")]
+        public static partial void ActivityFailed(this ILogger logger, Exception ex, string instanceId, string name);
+
+        /// <summary>
+        /// Creates a logger named "Microsoft.DurableTask.Worker" with the specified subcategories.
+        /// </summary>
+        /// <param name="loggerFactory">The logger factory to use to create the logger.</param>
+        /// <param name="subcategories">The subcategory of the logger. For example, "Activities" or "Orchestrations".
```
halspang, line 42: Is this `subcategories` because you can introduce more layers than just one? If so, can you include an example in the comment of multiple subcategory layers/how to delimit them? Or can we change it to subcategory?

cgillum: This is standard .NET logging stuff, so nothing here is specific to Durable. More information on categories and subcategories can be found [here](https://learn.microsoft.com/dotnet/core/extensions/logging?tabs=command-line#log-category).

halspang: Yes, I know it's just .NET logging, my complaint is more that we named the param `subcategories` then refer to it as `subcategory` in the comment without really extrapolating anything. So, I'd recommend we either rename the param to `subcategory` or update the comment to say `subcategories` and include an example that has multiple levels.

cgillum: Renamed to `subcategory`.

halspang (approved):

PR #413, opened 2025-04-18
Update logging for task hub NotFound

halspang:
This commit updates the logging for when a task hub connection returns NotFound. Previously it was logged as an unknown error, it is now a warn.

### DIFF
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -67,7 +67,12 @@ sealed partial class GrpcDurableTaskWorker
                 }
                 catch (RpcException ex) when (ex.StatusCode == StatusCode.NotFound)
                 {
-                    // Task hub is not found or insufficient permissions - retry
+                    // We retry on a NotFound for several reasons:
+                    // 1. It was the existing behavior through the UnexpectedError path.
+                    // 2. A 404 can be returned for a missing task hub or authentication failure. Authentication takes
+                    //    time to propagate so we should retry instead of making the user restart the application.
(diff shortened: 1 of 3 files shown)


src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -65,6 +65,11 @@ public async Task ExecuteAsync(CancellationToken cancellation)
                     // Sidecar is down - keep retrying
                     this.Logger.SidecarUnavailable();
                 }
+                catch (RpcException ex) when (ex.StatusCode == StatusCode.NotFound)
+                {
+                    // Task hub is not found or insufficient permissions - retry
```
cgillum, line 70: Typically, 404 Not Found responses are not supposed to be retried. Why would we want to retry in our case?

halspang: The two main reasons I had for this were:
1. A 404 is returned from the scheduler if the task hub isn't present or if there are insufficient permissions to access it. Given permissions can take time to propagate, I didn't want to have the customer have to go in and restart their worker to get things working again.
2. A scheduler and a task hub can be deployed in separate steps of a deployment. If the worker is mistakenly put in between them, it would mean the customer has to restart their worker again to get things going.
3. It's the behavior today (since we retry on unknown) and stopping that may be a breaking change. Who knows what customers depend on so I changed it to at least highlight the log better.

Let me know if you think those are valid reasons or if you think we should change the behavior or add additional sleeping.

cgillum: Sounds good to me! Might be worth mentioning this in the comments.

cgillum (approved):

cgillum (approved):

PR #415, opened 2025-04-22
Add New Property `Properties` to TaskOrchestrationContext

nytian:
This PR adds a new property IDictionay<string, object?> `Properties` at `TaskOrchestrationContext`, which allows `DurableTaskOptions` from WebJobs.Extensions.DurableTask can be passed through `Microsoft.DurableTask.Protobuf.OrchestratorRequest`. 

Related PR:
- https://github.com/Azure/azure-functions-durable-extension/pull/3092
- https://github.com/microsoft/durabletask-protobuf/pull/41


src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -38,9 +38,31 @@ public TaskOrchestrationContextWrapper(
...
+        IEnumerable<KeyValuePair<string, object?>> properties)
     {
         this.innerContext = Check.NotNull(innerContext);
         this.invocationContext = Check.NotNull(invocationContext);
+        if (properties is null)
+        {
+            throw new ArgumentNullException(nameof(properties));
+        }
+
+        this.Properties = properties.ToDictionary(pair => pair.Key, pair => pair.Value);
```
cgillum, line 65: Converting from `IEnumerable<KVP>` to a dictionary is dangerous because `IEnumerable<KVP>` allows duplicate keys whereas dictionary does not. We should change the `properties` parameter to be a dictionary instead of `IEnumerable<KVP>` to avoid these kinds of problems.

src/Abstractions/TaskOrchestrationContext.cs
```diff
@@ -65,6 +65,11 @@ public abstract class TaskOrchestrationContext
     /// </summary>
     public virtual string Version => string.Empty;
 
+    /// <summary>
+    /// Gets the configuration settings for the orchestration context.
+    /// </summary>
+    public virtual IDictionary<string, object?> Properties { get; } = new Dictionary<string, object?>();
```
cgillum, line 71: Can this be `IReadOnlyDictionary` instead of `IDictionary`? Seems like there's no scenario where we want customers adding items to this dictionary?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
jviau: It is safe if you use a `foreach` loop and set the key each time, so last key wins.

src/ScheduledTasks/Models/ScheduleState.cs
```diff
@@ -50,4 +50,4 @@ public void RefreshScheduleRunExecutionToken()
     {
         this.ExecutionToken = Guid.NewGuid().ToString("N");
     }
-}
```
nytian, line 53: This is unrelated to this PR, but just to fix some warnings

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
nytian: Thanks, Chris and Jacob. I updated this to be Dictionary. Because to write a foreach loop, I have to either grant Properties `private set`, or adding a new field to store this and then pass the value to Properties. Just feel like using Dictionary directly here seems like a more straightforward and cleaner way in this case.

jviau: I am okay with going back to dictionary, but I don't see why you need a `private set`. The following is valid without any setter on `Properties`:

``` CSharp
this.Properties = new();
foreach ((string key, object? value) in properties)
{
    this.Properties[key] = value;
}
```

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -84,7 +85,7 @@ public TaskOrchestrationContextWrapper(
     /// <summary>
     /// Gets the configuration settings for the orchestration.
     /// </summary>
-    public override IDictionary<string, object?> Properties => this.properties;
+    public override IDictionary<string, object?> Properties { get; } = new Dictionary<string, object?>();
```
jviau: You are always setting this in the ctor. No need to inline an initialization.

```suggestion
    public override IDictionary<string, object?> Properties { get; };
```

src/Worker/Core/Shims/TaskOrchestrationShim.cs
```diff
@@ -30,9 +31,24 @@ partial class TaskOrchestrationShim : TaskOrchestration
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TaskOrchestrationShim"/> class.
+    /// </summary>
+    /// <param name="invocationContext">The invocation context for this orchestration.</param>
+    /// <param name="implementation">The orchestration's implementation.</param>
+    /// <param name="properties">Configuration for the orchestration.</param>
+    public TaskOrchestrationShim(
+        OrchestrationInvocationContext invocationContext,
+        ITaskOrchestrator implementation,
+        Dictionary<string, object?> properties)
```
cgillum, line 47: Can this be read-only?
```suggestion
        IReadOnlyDictionary<string, object?> properties)
```

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -38,9 +38,27 @@ public TaskOrchestrationContextWrapper(
...
+    /// </summary>
+    /// <param name="innerContext">The inner orchestration context.</param>
+    /// <param name="invocationContext">The invocation context.</param>
+    /// <param name="deserializedInput">The deserialized input.</param>
+    /// <param name="properties">The configuration for context.</param>
+    public TaskOrchestrationContextWrapper(
+        OrchestrationContext innerContext,
+        OrchestrationInvocationContext invocationContext,
+        object? deserializedInput,
+        Dictionary<string, object?> properties)
```
cgillum, line 56: Can this be read-only?
```suggestion
        IReadOnlyDictionary<string, object?> properties)
```

cgillum (commented):
Just a couple of minor nits

cgillum (approved):

PR #416, opened 2025-04-22
Add version to TaskName

halspang:
This commit adds an explict set for Version in TaskName. It also includes a flag that is set so other code can determine if the TaskName is intended to be versioned. Finally, it updates the Version property in the TaskOrchestrationContext as that was always set to empty instead of being able to be used.

### DIFF
--- a/src/Abstractions/TaskOptions.cs
+++ b/src/Abstractions/TaskOptions.cs
@@ -90,6 +90,11 @@ public record SubOrchestrationOptions : TaskOptions
     /// Gets the orchestration instance ID.
     /// </summary>
     public string? InstanceId { get; init; }
+
+    /// <summary>
+    /// Gets the version to associate with the sub-orchestration instance.
+    /// </summary>
+    public TaskVersion Version { get; init; } = default!;
 }
 
 /// <summary>
@@ -108,4 +113,9 @@ public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffse
     /// Gets the tags to associate with the orchestration instance.
     /// </summary>
     public IReadOnlyDictionary<string, string> Tags { get; init; } = ImmutableDictionary.Create<string, string>();
+
+    /// <summary>
+    /// Gets the version to associate with the orchestration instance.
+    /// </summary>
+    public TaskVersion? Version { get; init; }
 }
--- a/src/Client/Grpc/GrpcDurableTaskClient.cs
+++ b/src/Client/Grpc/GrpcDurableTaskClient.cs
@@ -80,9 +80,9 @@ public sealed class GrpcDurableTaskClient : DurableTaskClient
 
         // We're explicitly OK with an empty version from the options as that had to be explicitly set. It should take precedence over the default.
         string version = string.Empty;
-        if (options != null && options.Version != null)
+        if (options?.Version is { } v)
         {
-            version = options.Version;
+            version = v;
         }
         else if (!string.IsNullOrEmpty(this.options.DefaultVersion))
         {
--- a/src/Client/Grpc/GrpcDurableTaskClient.cs
+++ b/src/Client/Grpc/GrpcDurableTaskClient.cs
@@ -78,10 +78,11 @@ public sealed class GrpcDurableTaskClient : DurableTaskClient
     {
         Check.NotEntity(this.options.EnableEntitySupport, options?.InstanceId);
 
+        // We're explicitly OK with an empty version from the options as that had to be explicitly set. It should take precedence over the default.
         string version = string.Empty;
-        if (!string.IsNullOrEmpty(orchestratorName.Version))
+        if (options != null && options.Version != null)
         {
-            version = orchestratorName.Version;
+            version = options.Version;
         }
         else if (!string.IsNullOrEmpty(this.options.DefaultVersion))
         {
--- a/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
@@ -177,10 +177,14 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
         static string? GetInstanceId(TaskOptions? options)
             => options is SubOrchestrationOptions derived ? derived.InstanceId : null;
         string instanceId = GetInstanceId(options) ?? this.NewGuid().ToString("N");
+        string defaultVersion = this.invocationContext.Options != null && this.invocationContext.Options.Versioning != null
+            ? this.invocationContext.Options.Versioning.DefaultVersion
+            : string.Empty;
+        string version = options != null && options is SubOrchestrationOptions subOptions ? subOptions.Version : defaultVersion;
 
         Check.NotEntity(this.invocationContext.Options.EnableEntitySupport, instanceId);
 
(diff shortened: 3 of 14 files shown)


halspang:
@jviau - I wasn't aware of that context, thanks for linking it. I'll go through it and see if I can determine a path forward there. In the meantime, I separated out the context change as I think this change should be isolated and that's the more important change that I need for versioning functionality in the extension.

halspang:
@jviau @cgillum 

This PR changed up a bit, but it has the updates from discussion had offline. The tl;dr of those conversations was to introduce a `TaskVersion` and utilize it in the Options of starting an orchestration. This means we don't worry about the round-trip of the `TaskName.Version` and don't have to make decisions about what divides the name/version. Overall, it's much safer and I feel a better pattern to follow going forward. I kept it in this PR to keep the history, though I can close and open a new one if we feel it should be redone.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -78,10 +78,11 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
     {
         Check.NotEntity(this.options.EnableEntitySupport, options?.InstanceId);
 
+        // We're explicitly OK with an empty version from the options as that had to be explicitly set. It should take precedence over the default.
         string version = string.Empty;
-        if (!string.IsNullOrEmpty(orchestratorName.Version))
+        if (options != null && options.Version != null)
```
jviau, line 83: I think pattern matching may be beneficial here? Not sure if I have the syntax right.
```suggestion
        if (options is { Version: v })
```

or null-operator:

``` CSharp
if (options?.Version is { } v)
```

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -177,10 +177,14 @@ public override async Task<TResult> CallSubOrchestratorAsync<TResult>(
         static string? GetInstanceId(TaskOptions? options)
             => options is SubOrchestrationOptions derived ? derived.InstanceId : null;
         string instanceId = GetInstanceId(options) ?? this.NewGuid().ToString("N");
+        string defaultVersion = this.invocationContext.Options != null && this.invocationContext.Options.Versioning != null
```
jviau, line 180: Null operators may help here:

```suggestion
        string defaultVersion = this.invocationContext.Options?.Versioning?.DefaultVersion ?? string.Empty;
```

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -177,10 +177,14 @@ public override async Task<TResult> CallSubOrchestratorAsync<TResult>(
...
+            ? this.invocationContext.Options.Versioning.DefaultVersion
+            : string.Empty;
+        string version = options != null && options is SubOrchestrationOptions subOptions ? subOptions.Version : defaultVersion;
```
jviau, line 183: `options is SubOrchestrationOptions subOptions` will already do a null-check, so no need for `options != null`

jviau (commented):
Please address all the warnings from adding `[Obsolete]` to the version property.

halspang (commented):
Updated the warnings for Obsolete/changed the reference where appropriate.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -177,10 +179,12 @@ public override async Task<TResult> CallSubOrchestratorAsync<TResult>(
         static string? GetInstanceId(TaskOptions? options)
             => options is SubOrchestrationOptions derived ? derived.InstanceId : null;
         string instanceId = GetInstanceId(options) ?? this.NewGuid().ToString("N");
+        string defaultVersion = this.invocationContext.Options?.Versioning?.DefaultVersion ?? string.Empty;
+        string version = options is SubOrchestrationOptions subOptions ? subOptions.Version : defaultVersion;
```
jviau, line 183: I just want to leave a comment regarding what we discussed offline here. I don't think it is a good idea to conflate worker version and orchestration version. I see these as separate concerns.

Worker version is purely for routing and supporting rolling updates.
Orchestration version is for enabling an API-versioning-esque scheme for backcompat.

Having worker version apply as a default to orchestration version unnecessarily complicates that scenario. Now customers need to think about how their individual orchestrations version in relation to their worker version as a whole. Which one takes precedence? How does a worker version less than an orchestration version behave? What if they aren't comparable at all?

All of these questions will need to be considered, answered, and documented if we keep these intertangled. Whereas if we just keep them entirely separate then answering all of that becomes much simpler.

src/Abstractions/TaskOptions.cs
```diff
@@ -90,6 +90,11 @@ public SubOrchestrationOptions(TaskOptions options, string? instanceId = null)
     /// Gets the orchestration instance ID.
     /// </summary>
     public string? InstanceId { get; init; }
+
+    /// <summary>
+    /// Gets the version to associate with the sub-orchestration instance.
+    /// </summary>
+    public TaskVersion Version { get; init; } = default!;
```
jviau, line 97: Have we evaluated versioning for `TaskActivity`? DurableTask.Core has it. I personally haven't used it much myself, unlike orchestration versioning which I used extensively. Still wondering if it is planned?

jviau (approved):

src/Abstractions/TaskOptions.cs
halspang: I also haven't used it much, the initial thought here is that differently versioned orchestrations can have the different activities and different version activities in the same orchestration were less likely. 

That being said, we've left the door open for it, so if it's a requested feature or something we agree to as a team we can still implement it down the road.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
halspang: I appreciate the comment and the conversation we had offline about this! I'll make sure that this is all clearly documented/explained.

> Having worker version apply as a default to orchestration version unnecessarily complicates that scenario. Now customers need to think about how their individual orchestrations version in relation to their worker version as a whole. Which one takes precedence? How does a worker version less than an orchestration version behave? What if they aren't comparable at all?

The default version only really applies to sub-orchestrations from a Worker point of view. It's matching the client side's default version. In my opinion, the worker shouldn't be starting sub-orchestrations without the client but since it does, we provide a means of setting it. For functions apps, this is the same `host.json` value. For apps just using this SDK, it does look a little weird as you have to supply it in both the client and the worker options. 

In regard to what takes precedence, it's always the version set by the options when scheduling an orchestration. The worker version is only used for the matching/accepting of an orchestration. The way they interact is set with the version matching strategy and the version failure strategy. The docs aren't published to learn yet as we wanted support across the SDKs and DTFx first, but this sample readme talks about the interaction: [Orchestration versioning sample](https://github.com/Azure-Samples/Durable-Task-Scheduler/tree/main/samples/durable-task-sdks/dotnet/OrchestrationVersioning).

cgillum: Not sure if this was covered in the offline discussion, but it's probably best to not think of this as a "worker version", but rather as a "default orchestration version" which happens to be something that can be configured at a client and/or worker level.

PR #417, opened 2025-04-24
Add User-Agent Header to gRPC Metadata

torosent:
This PR adds a user-agent header to the gRPC metadata in the DurableTaskSchedulerClientOptions.CreateChannel() method.

#### **Why add a user-agent header?**

- **Telemetry & Diagnostics:** Including a user-agent string allows the Durable Task Scheduler backend to identify requests coming from the durabletask-dotnet SDK. This helps with usage tracking, diagnostics, and troubleshooting.
- **Version Awareness:** The user-agent includes the SDK version (via `DurableTaskUserAgentUtil.GetUserAgent()`), making it easier to correlate issues or behaviors with specific client versions.
- **Best Practices:** Sending a user-agent is a common best practice for SDKs and clients interacting with cloud services, enabling better support and analytics on the service side.

#### **Summary of changes**

- The gRPC call interceptor now adds a `user-agent` header with the durabletask-dotnet SDK name and version to every outgoing request.

This change improves observability and supportability for both SDK users and service operators.


src/Shared/Core/DurableTaskVersionUtil.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.DurableTask;
+
+public static class DurableTaskUserAgentUtil
+{
+    static string SdkName => "durabletask-dotnet";
+
+    public static string GetUserAgent()
+    {
+        var version = typeof(DurableTaskUserAgentUtil).Assembly.GetName().Version;
```
cgillum, line 11: This code gets the assembly version, which is different from the DLL version and isn't really what we want. Use this instead:
```suggestion
        string version = FileVersionInfo.GetVersionInfo(typeof(DurableTaskUserAgentUtil).Assembly.Location).FileVersion;
```

cgillum: Also, make this into a `readonly static` value so that we read it only once and cache it rather than for every gRPC request.

```csharp
static readonly string PackageVersion = FileVersionInfo.GetVersionInfo(typeof(DurableTaskUserAgentUtil).Assembly.Location).FileVersion;
```

src/Shared/Core/DurableTaskVersionUtil.cs
```diff
@@ -0,0 +1,14 @@
+using System.Reflection;
+
+namespace Microsoft.DurableTask;
+
+public static class DurableTaskUserAgentUtil
+{
+    static string SdkName => "durabletask-dotnet";
```
cgillum, line 16:
```suggestion
    const string SdkName => "durabletask-dotnet";
```

src/Shared/Core/DurableTaskVersionUtil.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT License.
+
+using System.Reflection;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Utility class for generating the user agent string for the Durable Task SDK.
+/// </summary>
+public static class DurableTaskUserAgentUtil
```
cgillum, line 11: Why is this public and why is it in the shared code? Since this is only used by the `AzureManaged` package, shouldn't it go into that project?

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -96,6 +96,9 @@ this.Credential is not null
             async (context, metadata) =>
             {
                 metadata.Add("taskhub", taskHubName);
+
+                // Add user agent header with durabletask-dotnet and DLL version from util
+                metadata.Add("user-agent", $"{DurableTaskUserAgentUtil.GetUserAgent()}");
```
cgillum, line 101: This covers the client APIs, but shouldn't we also do this for the worker API calls?

src/Shared/Core/DurableTaskVersionUtil.cs
torosent: fixed

src/Shared/Core/DurableTaskVersionUtil.cs
torosent: fixed

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
torosent: done

src/Shared/Core/DurableTaskVersionUtil.cs
torosent: I moved it to the shared/azuremanaged

cgillum (approved):

PR #426, opened 2025-05-10
Activity tag support

YunchuWang:


src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -90,7 +91,9 @@ internal static HistoryEvent ConvertHistoryEvent(P.HistoryEvent proto, EntityCon
                     proto.EventId,
                     proto.TaskScheduled.Name,
                     proto.TaskScheduled.Version,
-                    proto.TaskScheduled.Input);
+                    proto.TaskScheduled.Input) {
```
cgillum, line 94: nit: please fix whitespace (opening curly brace goes on its own line in C#)

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -5,6 +5,7 @@
 using System.Buffers.Text;
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
```
cgillum, line 8: Not used?

src/Abstractions/TaskOptions.cs
```diff
@@ -52,6 +52,13 @@ public TaskOptions(TaskRetryOptions? retry = null)
     /// <param name="instanceId">The instance ID to use.</param>
     /// <returns>A new <see cref="SubOrchestrationOptions" />.</returns>
     public SubOrchestrationOptions WithInstanceId(string instanceId) => new(this, instanceId);
+
+    /// <summary>
+    /// Returns a new <see cref="CallActivityOptions" /> with the provided tags.
+    /// </summary>
+    /// <param name="tags">The tags to associate with the activity.</param>
+    /// <returns>A new <see cref="CallActivityOptions" />.</returns>
+    public CallActivityOptions WithTags(IDictionary<string, string> tags) => new(this, tags);
```
cgillum, line 61: Returning `CallActivityOptions` means we can't use this for sub-orchestrations in the future, correct? However, I think supporting sub-orchestrations will be important since we support it already for top-level orchestrations. How can we modify the design to support both?

samples/ScheduleWebApp/Activities/CacheClearingActivity.cs
```diff
@@ -0,0 +1,15 @@
...
+using Microsoft.DurableTask;
+
+namespace ScheduleWebApp.Activities;
+
+[DurableTask] // Optional: enables code generation for type-safe calls
+public class CacheClearingActivity : TaskActivity<object, string>
+{
+    public override Task<string> RunAsync(TaskActivityContext context, object input)
+    {
+        return Task.FromResult("hello");
```
cgillum, line 13: Move the `Task.Delay` and the code comment that you removed from the orchestrator into this code block.

cgillum (changes requested):
I think we need to revisit the public API change. In particular, I don't think we should limit tags to activity calls.

src/Abstractions/TaskOptions.cs
YunchuWang: moved tags to taskoptions as common optional field since it is common to all subclasses supported currently. suborchoptions can later add new ctor to create with tags. also removed withtags fluent api (not extensible if in future we add more and more fields to subtypes, need to create lots of withXXYYZZ functions) and let customers create callactivityoptions directly or using c9 record with api.

src/Abstractions/TaskOptions.cs
```diff
@@ -109,3 +125,29 @@ public record StartOrchestrationOptions(string? InstanceId = null, DateTimeOffse
     /// </summary>
     public IReadOnlyDictionary<string, string> Tags { get; init; } = ImmutableDictionary.Create<string, string>();
 }
+
+/// <summary>
+/// Options for calling activities from an orchestrator.
+/// </summary>
+public record CallActivityOptions : TaskOptions
```
cgillum, line 132: Why do we need this class if the base class already supports Tags? Shouldn't the `CallActivityAsync` implementation just reference `TaskOptions.Tags` instead of `CallActivityOptions.Tags`?

YunchuWang: mainly keeping this for semantic separation and clarity. context.CallActivityAsync(..., CallActivityOptions options) is self documenting and consistent with suborchoptions (even now callactivityoptions dont have unique fields in itself)

YunchuWang:
@cgillum Hi Chris, all tests passed (tested locally works). 
image i see the inmemoryorchservice used in integ test do not support getorchestrationhistory, i will add the integration test to dts after this is released.
can I get approval? thanks

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -138,27 +138,51 @@ public override async Task<T> CallActivityAsync<T>(
 
         try
         {
+            IDictionary<string, string>? tags = new Dictionary<string, string>();
```
cgillum: For performance reasons, we should not allocate this dictionary unless it's needed, especially since this is a common code path but using tags is not a common case.

src/Abstractions/TaskOptions.cs
cgillum: Adding new public types creates technical debt and maintenance overhead, including documentation. I don't think we should add any public APIs if they aren't needed for anything. SubOrchestrationOptions was created because it has options specific to sub-orchestrations. That's not the case here.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
YunchuWang: @cgillum   in released durabletask core https://github.com/Azure/durabletask/blob/e3ecb29e6b945e621c07ec8288d5ecaa547edd5b/src/DurableTask.Core/TaskOrchestrationContext.cs#L94 scheduletaskoptions is made to build with non-null tags. passing it as null (tested) would throw param cant be null exception.

agree that while allocating a new Dictionary() is functionally valid (cpu time on order of nanoseconds), it creates unnecessary memory allocations that lead to garbage collection pressure in high-throughput environments. 
i am changing to use the static ImmutableDictionary.Empty singleton avoids this cost entirely while unblocks the activity tags feature

samples/ScheduleWebApp/Orchestrations/CacheClearingOrchestrator.cs
```diff
@@ -14,8 +15,12 @@ public override async Task<string> RunAsync(TaskOrchestrationContext context, st
...
             logger.LogInformation("Starting CacheClearingOrchestration for schedule ID: {ScheduleId}", scheduleId);
 
-            // Simulate cache clearing
-            await Task.Delay(TimeSpan.FromSeconds(5));
+            TaskOptions options = new TaskOptions(tags: new Dictionary<string, string>
+            {
+                { "scheduleId", scheduleId }
+            });
+
+            await context.CallActivityAsync(nameof(CacheClearingActivity), options);
```
nytian, line 23: I am confused what is this scheduleId used for when calling CacheClearingActivity? Can you add some comments to describe about this test? like how this tag is used.

nytian (commented):
Can we add a integration tests, e.g.  in here https://github.com/microsoft/durabletask-dotnet/blob/main/test/Grpc.IntegrationTests/OrchestrationPatterns.cs to simulate calling activity with tags? Just want to make sure this behavior is ensured, as we don't run samples in the CI usually.

samples/ScheduleWebApp/Orchestrations/CacheClearingOrchestrator.cs
YunchuWang: as discussed offline, test to run the activity with tag added but to verify tag retrieval (as mentioned in pr description) this repo does not have the support. i will add the retrieval test to dtmb.

nytian (approved):

PR #427, opened 2025-05-10
Add extension methods for registering entities by type

cgillum:
This PR introduces functionality for registering `ITaskEntity` implementations in the `DurableTaskRegistry`.

### Improved support for entity registration in `DurableTaskRegistry`:
* [`src/Abstractions/DurableTaskRegistry.Entities.cs`](diffhunk://#diff-9a292b16090e0a41d380420e949f390fe04bf62316886ead992f0b03177dc87fR1-R77): Added several methods to `DurableTaskRegistry` for registering entities by type, name, and instance, enabling more convenient entity registration.
* [`src/ScheduledTasks/Extension/DurableTaskWorkerBuilderExtensions.cs`](diffhunk://#diff-b89fc591cada9bd5657c1e1f41beaf925df78706c474aa7057ac50762b75f43aL21-R20): Simplified the `UseScheduledTasks` method by replacing a lambda-based entity registration with the new `AddEntity<Schedule>()` method.

### DIFF
--- a/src/Abstractions/DurableTaskRegistry.Entities.cs
+++ b/src/Abstractions/DurableTaskRegistry.Entities.cs
@@ -1,0 +1,77 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Container for registered <see cref="ITaskOrchestrator" />, <see cref="ITaskActivity" />,
+/// and <see cref="ITaskEntity"/> implementations.
+/// </summary>
+public partial class DurableTaskRegistry
+{
+    /// <summary>
+    /// Registers an entity factory.
+    /// </summary>
+    /// <param name="name">The name of the entity to register.</param>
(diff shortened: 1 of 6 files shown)


src/Abstractions/DurableTaskRegistry.Entities.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Container for registered <see cref="ITaskOrchestrator" />, <see cref="ITaskActivity" />,
+/// and <see cref="ITaskEntity"/> implementations.
+/// </summary>
+public partial class DurableTaskRegistry
```
YunchuWang, line 13: nitpick: in comments, we can group functions by registration types to enhance readability? ex.
// --- Type(Factory)-based Registration ---
// --- Generic Type-based Registration ---
// --- Singleton Instance Registration ---

cgillum: I don't think this is necessary. There are only 6 methods and they're all quite short, so it's not difficult to read as-is.

PR #437, opened 2025-06-03
Update sub-orchestration default versioning

halspang:
This commit updates the sub-orchestration default version handling to incorporate properties from TaskOrchestrationContext. These properties are generated and passed into the context by systems using Durable Functions.

### DIFF
--- a/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
@@ -179,9 +179,8 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
         static string? GetInstanceId(TaskOptions? options)
             => options is SubOrchestrationOptions derived ? derived.InstanceId : null;
         string instanceId = GetInstanceId(options) ?? this.NewGuid().ToString("N");
-        string defaultVersion = this.invocationContext.Options?.Versioning?.DefaultVersion ?? string.Empty;
-        string version = options is SubOrchestrationOptions subOptions ? subOptions.Version : defaultVersion;
-
+        string defaultVersion = this.GetDefaultVersion();
+        string version = options is SubOrchestrationOptions { Version: { } v } ? v.Version : defaultVersion;
         Check.NotEntity(this.invocationContext.Options.EnableEntitySupport, instanceId);
 
         // if this orchestration uses entities, first validate that the suborchestration call is allowed in the current context
@@ -481,5 +480,23 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
                 }
             }
         }
+    }
+
+    // The default version can come from two different places depending on the context of the invocation.
+    string GetDefaultVersion()
+    {
+        // Preferred choice.
+        if (this.invocationContext.Options.Versioning?.DefaultVersion is { } v)
+        {
+            return v;
+        }
+
+        // Secondary choice.
+        if (this.Properties.TryGetValue("defaultVersion", out var propVersion) && propVersion is string v2)
+        {
+            return v2;
+        }
+
+        return string.Empty;
     }
 }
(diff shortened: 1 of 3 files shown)


src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -179,9 +179,8 @@ public override async Task<TResult> CallSubOrchestratorAsync<TResult>(
         static string? GetInstanceId(TaskOptions? options)
             => options is SubOrchestrationOptions derived ? derived.InstanceId : null;
         string instanceId = GetInstanceId(options) ?? this.NewGuid().ToString("N");
-        string defaultVersion = this.invocationContext.Options?.Versioning?.DefaultVersion ?? string.Empty;
-        string version = options is SubOrchestrationOptions subOptions ? subOptions.Version : defaultVersion;
-
+        string defaultVersion = this.GetDefaultVersion();
+        string version = options is SubOrchestrationOptions { Version: { } v } ? v.Version : defaultVersion;
```
cgillum, line 183: I don't understand this C# syntax, but I'll take your word for it that it's good and correct.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -482,4 +481,32 @@ async Task<T> InvokeWithCustomRetryHandler<T>(
...
         }
     }
+
+    string GetDefaultVersion()
+    {
+        // The default version can come from two different places depending on the context of the invocation.
+        string? propertiesDefault;
+        string? optionsDefault;
+
+        if (this.invocationContext.Options.Versioning?.DefaultVersion is { } v)
```
cgillum, line 489: Could this just be `if (!string.IsNullOrEmpty(this.invocationContext.Options.Versioning?.DefaultVersion))`?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -482,4 +481,32 @@ async Task<T> InvokeWithCustomRetryHandler<T>(
             }
         }
     }
+
+    string GetDefaultVersion()
+    {
+        // The default version can come from two different places depending on the context of the invocation.
+        string? propertiesDefault;
+        string? optionsDefault;
```
cgillum: Could the `else` conditions below be removed by just assigning these variables a `null` value here?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -482,4 +481,32 @@ async Task<T> InvokeWithCustomRetryHandler<T>(
...
+
+        if (this.invocationContext.Options.Versioning?.DefaultVersion is { } v)
+        {
+            optionsDefault = v;
```
cgillum: nit: if the options default is preferred, it would be more efficient to just return here rather than continue on to the next if check and pay the cost of a dictionary lookup.

cgillum (approved):

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
halspang: .NET has some fancy null checking/casting nowadays :) You can thank Jacob for showing it to me!

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
halspang: So this check is actually just doing the null check and casting it out to `v`. We can't do a null or empty check because empty is actually a valid version and different from a null version (not set).

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
halspang: They sure could, that was a little verbose of me.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
halspang: True! Then I can also eliminate the extra variables and the null coalescing.

PR #441, opened 2025-06-12
Add orchestration execution tracing

### LINKED ISSUE #439: Tracing parity with Durable Task Framework (DTFx)
The Durable Task Framework (DTFx) SDK provides Durable Task specific tracing across orchestration execution. The Durable Task .NET SDK provides no tracing (i.e. separate from any provided when used with Azure Durable Functions).

The Durable Task .NET SDK should match DTFx tracing so that, when used on its own, clients can get a complete picture of orchestration execution including how their own traces dovetail with DT execution. 

Work in progress is here, basically a port of the DTFx behavior:

https://github.com/philliphoff/durabletask-dotnet/tree/philliphoff-orchestration-tracing

Requires some changes to the [DT protos](https://github.com/microsoft/durabletask-protobuf/pull/44)

philliphoff:
> NOTE: **!!!** Do not merge until the associated `protobuf` PR is committed and this PR has been updated to match. **!!!**

Ports Durable Task Framework (DTFx) tracing behavior to the Durable Task .NET SDK gRPC layer to allow end-to-end tracing of orchestrations from creation to completion, including:

 - Client-side
   - Creating orchestrations
   - Raising events
 - Worker-side
   - Executing orchestrations (where all activity is parented to a single "logical" span)
   - Creating suborchestrations
   - Creating timers
   - Sending events
   
With this change, users can get a complete picture of orchestration execution including how their own traces dovetail with Durable Task execution, as shown in the following example, which follows creation of an orchestration starting from a web API, to execution by a DT worker, with suborchestrations, and back to the web API as part of task execution.

Screenshot 2025-06-12 at 16 36 03

 This change depends on corresponding changes to the [Durable Task protobuf definitions](https://github.com/microsoft/durabletask-protobuf/pull/44).  It also assumes any backend properly accepts and returns the parent trace contexts given it by the SDK.
 
 Resolves #439

### DIFF
--- a/src/Grpc/versions.txt
+++ b/src/Grpc/versions.txt
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch main at 2025-06-02 21:12:34 UTC
-https://raw.githubusercontent.com/microsoft/durabletask-protobuf/fd9369c6a03d6af4e95285e432b7c4e943c06970/protos/orchestrator_service.proto
+# The following files were downloaded from branch main at 2025-08-08 16:46:11 UTC
+https://raw.githubusercontent.com/microsoft/durabletask-protobuf/e88acbd07ae38b499dbe8c4e333e9e3feeb2a9cc/protos/orchestrator_service.proto


Directory.Packages.props
sophiatev, line 1: It looks like this file has no actual updates, can you remove the changes?

src/Grpc/orchestrator_service.proto
sophiatev, line 1: I think we don't want to merge individual proto file changes like this, but rather wait until your protobuf changes are merged into the protobuf repo, run the command to update this proto file/versions.txt, and then merge the versions.txt and orchestrator_service.proto changes together (I think)

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+            newActivity.SetTag(Schema.Task.InstanceId, createInstanceRequest.InstanceId);
+            newActivity.SetTag(Schema.Task.ExecutionId, createInstanceRequest.ExecutionId);
+
+            if (!string.IsNullOrEmpty(createInstanceRequest.Version))
+            {
+                newActivity.SetTag(Schema.Task.Version, createInstanceRequest.Version);
+            }
+        }
+
+        if (Activity.Current?.Id != null || Activity.Current?.TraceStateString != null)
```
sophiatev: I'm assuming you got this from `GrpcDurableTaskClient`, but these checks are actually unnecessary. If `Activity.Current` is non-null then it has a valid `Id`. And it doesn't make sense to check for the `TraceStateString` separately either. So just a check for if `Activity.Current` is non-null should suffice 

```c#
        {
            createInstanceRequest.ParentTraceContext ??= new P.TraceContext();
            createInstanceRequest.ParentTraceContext.TraceParent = activity.Id;
            createInstanceRequest.ParentTraceContext.TraceState = activity.TraceStateString;
        }
```

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -279,32 +281,52 @@ internal static Timestamp ToTimestamp(this DateTime dateTime)
     /// <exception cref="NotSupportedException">When an orchestrator action is unknown.</exception>
     internal static P.OrchestratorResponse ConstructOrchestratorResponse(
         string instanceId,
+        string executionId,
```
sophiatev, line 286: Can you fix the warnings for this file and add XML comments?

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -378,6 +407,12 @@ internal static P.OrchestratorResponse ConstructOrchestratorResponse(
                             Name = sendEventAction.EventName,
                             Data = sendEventAction.EventData,
                         };
+
+                        // Distributed Tracing: start a new trace activity derived from the orchestration
+                        // for an EventRaisedEvent (external event)
+                        using Activity? traceActivity = TraceHelper.StartTraceActivityForEventRaisedFromWorker(sendEventAction,instanceId, executionId);
```
sophiatev: Nit: missing a space between `sendEventAction` and `instanceId`

src/Shared/Grpc/Tracing/DiagnosticActivityExtensions.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Copyright (c) Microsoft Corporation.
```
sophiatev, line 1: Can you fix the warnings in this file?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
             string completionToken,
             CancellationToken cancellationToken)
         {
+            var executionStartedEvent =
```
sophiatev, line 365: You can just call `runtimeState.ExecutionStartedEvent` once the `OrchestrationRuntimeState` is created in the `try` block below (unless you want to do this outside of the try block?)

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
...
+                                request.InstanceId,
+                                subOrchestrationEvent,
+                                subOrchestrationEvent.SubOrchestrationInstanceCreated,
+                                newEvent.SubOrchestrationInstanceFailed);
+                            break;
+                        }
+
+                    case P.HistoryEvent.EventTypeOneofCase.TaskCompleted:
+                        {
+                            var taskScheduledEvent = request.PastEvents.Where(x => x.EventTypeCase == P.HistoryEvent.EventTypeOneofCase.TaskScheduled).LastOrDefault(x => x.EventId == newEvent.TaskCompleted.TaskScheduledId);
```
sophiatev: Nit: Can you break this up into multiple lines like you did for the suborchestration cases?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
...
+
+            foreach (var newEvent in request.NewEvents)
+            {
+                switch (newEvent.EventTypeCase)
+                {
+                    case P.HistoryEvent.EventTypeOneofCase.SubOrchestrationInstanceCompleted:
+                        {
+                            var subOrchestrationEvent = request.PastEvents
+                                .Where(x => x.EventTypeCase == P.HistoryEvent.EventTypeOneofCase.SubOrchestrationInstanceCreated)
+                                .LastOrDefault(x => x.EventId == newEvent.SubOrchestrationInstanceCompleted.TaskScheduledId);
```
sophiatev: It looks like the equivalent method in DT.Core calls `FirstOrDefault`. I don't think it should matter since presumably there should be only one `SubOrchestrationInstanceCreatedEvent` with the same Id as this `SubOrchestrationInstanceCompletedEvent`, but I just want to make sure the logic is consistent

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
...
+                        {
+                            var taskScheduledEvent = request.PastEvents.Where(x => x.EventTypeCase == P.HistoryEvent.EventTypeOneofCase.TaskScheduled).LastOrDefault(x => x.EventId == newEvent.TaskFailed.TaskScheduledId);
+                            TraceHelper.EmitTraceActivityForTaskFailed(request.InstanceId, taskScheduledEvent, taskScheduledEvent.TaskScheduled, newEvent.TaskFailed);
+                            break;
+                        }
+
+                    case P.HistoryEvent.EventTypeOneofCase.TimerFired:
+                        {
+                            // We immediately publish the activity span for this timer by creating the activity and immediately calling Dispose() on it.
+                            TraceHelper.EmitTraceActivityForTimer(request.InstanceId, executionStartedEvent?.Name, newEvent.Timestamp.ToDateTime(), newEvent.TimerFired);
```
sophiatev: Can you fix the warning here?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+// NOTE: Modified from https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Tracing/TraceHelper.cs
+
+using System.Diagnostics;
+using DurableTask.Core.Command;
+using Google.Protobuf.WellKnownTypes;
+using P = Microsoft.DurableTask.Protobuf;
+
+namespace Microsoft.DurableTask.Tracing;
+
+public class TraceHelper
```
sophiatev: Can you add a class comment to fix the warning?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+{
+    const string Source = "Microsoft.DurableTask";
+
+    static readonly ActivitySource ActivityTraceSource = new ActivitySource(Source);
+
+    /// <summary>
+    /// Starts a new trace activity for scheduling an orchestration from the client.
+    /// </summary>
+    /// <param name="startEvent">The orchestration's execution started event.</param>
```
sophiatev: Can you fix these method comments to match the modified parameters you have here (relative to DT.Core)?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+    /// <returns>
+    /// Returns a newly started <see cref="Activity"/> with orchestration-specific metadata.
+    /// </returns>
+    internal static Activity? StartActivityForNewOrchestration(P.CreateInstanceRequest createInstanceRequest)
+    {
+        Activity? newActivity = ActivityTraceSource.StartActivity(
```
sophiatev, line 32: Is there a reason you're not setting `Activity.Current` as the parent trace context of the `newActivity` you create here as is done in DT.Core? Is `Activity.Current` null?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+        activity.SetTag(Schema.Task.InstanceId, startEvent.OrchestrationInstance.InstanceId);
+
+        if (!string.IsNullOrEmpty(startEvent.Version))
+        {
+            activity.SetTag(Schema.Task.Version, startEvent.Version);
+        }
+
+        if (startEvent.OrchestrationSpanID != null)
+        {
+            activity.SetSpanId(startEvent.OrchestrationSpanID!);
```
sophiatev: Can you explain why you only set the span ID here (and in the else block below) and not the Id as well like is done in DT.Core?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,449 @@
...
+            parentContext: Activity.Current?.Context ?? default);
+
+        if (activity == null)
+        {
+            return null;
+        }
+
+        if (createdEvent.ParentTraceContext != null)
+        {
+            if (ActivityContext.TryParse(createdEvent.ParentTraceContext.TraceParent, createdEvent.ParentTraceContext?.TraceState, out ActivityContext parentContext))
```
sophiatev: You don't need the null check on `createdEvent.ParentTraceContext?.TraceState` here. Also, is there a reason you didn't just combine this with the previous `if` statement with a `&&`?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
...
+                                .LastOrDefault(x => x.EventId == newEvent.SubOrchestrationInstanceCompleted.TaskScheduledId);
+
+                            TraceHelper.EmitTraceActivityForSubOrchestrationCompleted(
+                                request.InstanceId,
+                                subOrchestrationEvent,
+                                subOrchestrationEvent.SubOrchestrationInstanceCreated);
+                            break;
+                        }
+
+                    case P.HistoryEvent.EventTypeOneofCase.SubOrchestrationInstanceFailed:
```
sophiatev: For these suborchestration-related events and for the timer, is there a reason you don't have the `if (Activity.Current != null)` check that DT.Core has [here](https://github.com/Azure/durabletask/blob/750d78f12d660f518f8a8b412ba6abf22314d203/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L885)? Presumably it's because the timer/suborchestration events rely on `Activity.Current` as the parent trace context

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,70 @@ async Task OnRunOrchestratorAsync(
...
+                    .FirstOrDefault();
+
+            Activity? traceActivity = TraceHelper.StartTraceActivityForOrchestrationExecution(executionStartedEvent);
+
+            foreach (var newEvent in request.NewEvents)
+            {
+                switch (newEvent.EventTypeCase)
+                {
+                    case P.HistoryEvent.EventTypeOneofCase.SubOrchestrationInstanceCompleted:
+                        {
```
sophiatev: Nit: I don't think you actually need the curly braces for these `case` statements?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -493,6 +562,18 @@ await this.client.AbandonTaskOrchestratorWorkItemAsync(
...
+            var completeOrchestrationAction = response.Actions.FirstOrDefault(a => a.CompleteOrchestration is not null);
+
+            if (completeOrchestrationAction is not null)
+            {
+                if (completeOrchestrationAction.CompleteOrchestration.OrchestrationStatus == P.OrchestrationStatus.Failed)
+                {
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, completeOrchestrationAction.CompleteOrchestration.Result);
+                }
+
+                traceActivity?.Stop();
```
sophiatev: I think you actually want to call `traceActivity?.Dispose()` here

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -493,6 +562,18 @@ await this.client.AbandonTaskOrchestratorWorkItemAsync(
                 };
             }
 
+            var completeOrchestrationAction = response.Actions.FirstOrDefault(a => a.CompleteOrchestration is not null);
+
+            if (completeOrchestrationAction is not null)
+            {
+                if (completeOrchestrationAction.CompleteOrchestration.OrchestrationStatus == P.OrchestrationStatus.Failed)
+                {
+                    traceActivity?.SetStatus(ActivityStatusCode.Error, completeOrchestrationAction.CompleteOrchestration.Result);
```
sophiatev: Is there a reason you opted to call `SetStatus` here as opposed to `SetTag(Schema.Task.Status, runtimeStatus)` like DT.Core does using `runtimeState.OrchestrationStatus.ToString()`? It looks like it does this in all cases (not just failures)

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -562,6 +649,9 @@ async Task OnRunActivityAsync(P.ActivityRequest request, string completionToken,
                 CompletionToken = completionToken,
             };
 
+            // Stop the trace activity here to avoid including the completion time in the latency calculation
+            traceActivity?.Stop();
```
sophiatev, line 725: Again I think you want to call `traceActivity?.Dispose()`

src/Worker/Grpc/GrpcOrchestrationRunner.cs
```diff
@@ -117,10 +117,13 @@ public static string LoadAndRun(
 
         P.OrchestratorResponse response = ProtoUtils.ConstructOrchestratorResponse(
             request.InstanceId,
+            request.ExecutionId,
             result.CustomStatus,
             result.Actions,
             completionToken: string.Empty, /* doesn't apply */
-            entityConversionState: null);
+            entityConversionState: null,
+            // TODO: Should this activity be created?
```
sophiatev: This might lead to redundancy in the case that this is being used by Durable Functions (since DT.Core will also create the trace activity in that case, I think)

test/Grpc.IntegrationTests/GrpcSidecar/Dispatcher/GrpcSubOrchestrationInstanceCreatedEvent.cs
```diff
@@ -0,0 +1,19 @@
...
+public class GrpcSubOrchestrationInstanceCreatedEvent : SubOrchestrationInstanceCreatedEvent
+{
+    public GrpcSubOrchestrationInstanceCreatedEvent(int eventId)
+        : base(eventId)
+    {
+    }
+
+    [DataMember]
+    public DistributedTraceContext? ParentTraceContext { get; set; }
+}
```
sophiatev: Can you add a new line at the end of the file?

test/Grpc.IntegrationTests/GrpcSidecar/Dispatcher/TaskOrchestrationDispatcher.cs
```diff
@@ -89,6 +89,14 @@ protected override async Task ExecuteWorkItemAsync(TaskOrchestrationWorkItem wor
                 continue;
             }
 
+            ExecutionStartedEvent? executionStartedEvent = workItem.OrchestrationRuntimeState.NewEvents.OfType<ExecutionStartedEvent>().FirstOrDefault();
```
sophiatev: You can call `workItem.OrchestrationRuntimeState.ExecutionStartedEvent` here

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -279,32 +281,52 @@ internal static Timestamp ToTimestamp(this DateTime dateTime)
...
+        EntityConversionState? entityConversionState,
+        Activity? parentActivity)
     {
         Check.NotNull(actions);
         var response = new P.OrchestratorResponse
         {
             InstanceId = instanceId,
             CustomStatus = customStatus,
             CompletionToken = completionToken,
+            OrchestrationSpanID = parentActivity?.SpanId.ToString(),
```
sophiatev: Can you help explain the flow here - it looks like you save this information to the `OrchestratorResponse`, but then attempt to extract it from the `ExecutionStartedEvent` in the `TraceHelper.StartTraceActivityForOrchestrationExecution`? So somewhere the information is getting transferred from this to the `ExecutionStartedEvent`, but I'm missing where that happens

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -324,6 +346,13 @@ internal static P.OrchestratorResponse ConstructOrchestratorResponse(
                         InstanceId = subOrchestrationAction.InstanceId,
                         Name = subOrchestrationAction.Name,
                         Version = subOrchestrationAction.Version,
+                        ParentTraceContext = clientActivityContext is not null
```
sophiatev: Rather than having to creating a whole `TraceContext` object and having to create these strings manually, can you just save the span ID? Since it looks like that is the only information `TraceHelper.CreateTraceActivityForSchedulingSubOrchestration` and `TraceHelper.StartTraceActivityForSchedulingTask` actually use

philliphoff:
> Just out of curiosity: do we plan to do the same thing for entities at some point?

My first step was just to reach parity with DTFx but I think we'd definitely want to extended to entities/schedules.

Directory.Packages.props
philliphoff: Looks like the tools formatted it; will revert it.

src/Grpc/orchestrator_service.proto
philliphoff: Exactly, it was useful/necessary to make the changes here to help provide context as to why some of the protobuf changes were necessary (as that came up in that PR). 

I'll add a note to the top the PR to ensure this PR's not merged until that happens.

src/Shared/Grpc/ProtoUtils.cs
philliphoff: Will do.  (It looks like I had a bunch of doc/warning fixes in progress when I had to switch to something else.  These should now be taken care of.)

src/Shared/Grpc/ProtoUtils.cs
philliphoff: Update.

src/Shared/Grpc/Tracing/DiagnosticActivityExtensions.cs
philliphoff: Should now be fixed.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: Updated.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: That's a good catch. I agree it shouldn't make a difference but best to be consistent. Updated.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: AFAIK, `ActivityTraceSource.StartActivity()` will automatically parent any current activity to the started activity (unless using the overload that specifies an explicit parent, which we are *not*).  I don't see DT.Core doing this either.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: If I understand correctly, `Activity.Id` is derived from `Activity.SpanId`, so passing/setting both is redundant and prevents us from needing to add that additional field to the protos.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Yes, you're right.  (I didn't realize `TraceState` was a "nullable" proto string.) Updated.

test/Grpc.IntegrationTests/GrpcSidecar/Dispatcher/GrpcSubOrchestrationInstanceCreatedEvent.cs
philliphoff: Updated.

test/Grpc.IntegrationTests/GrpcSidecar/Dispatcher/TaskOrchestrationDispatcher.cs
philliphoff: Updated.

src/Shared/Grpc/ProtoUtils.cs
philliphoff: Unless I misunderstand, while it's true that `TraceHelper` only use a subset of `TraceContext`, the protos for those history events themselves offer a parent `TraceContext` so this is necessary to satisfy that original contract (e.g. across all platforms).  Currently, for example, DTS is passing the original parent context at orchestration creation to those events, which is be better than nothing, but results in gaps in the trace.  Allowing for the context to round trip gives the SDKs flexibility in adding layers to the trace.

philliphoff: That said, I did do some cleanup to simplify/consolidate creation of these context objects.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: Because they create fields (with the same name), braces are required.  I've done some consolidation as well as removed the one case where they were *not* needed, though.

src/Shared/Grpc/Tracing/TraceHelper.cs
sophiatev: Ah okay, my guess was that the point of this logic (when I was looking at it a few months ago) in DT.Core was to ensure that every "new Activity" we make as part of the orchestration execution when this method gets called actually all corresponds to the same trace span in app insights (or whatever else is being used to look at the traces). But if this is sufficient to achieve that then sounds good to me

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: In this case, the activity is already disposed at the end of the method (via the `using`); we just want to stop before disposal to not include the last time.  While we *could* call `Dispose()`, as it does an implicit `Stop()`, the `Stop()` feels more direct.

src/Shared/Grpc/Tracing/TraceHelper.cs
sophiatev: Actually: could there be a situation where a difference in `TraceFlags` leads to unexpected behavior here? As I understand it, the `Activity.Id` is based off of the trace id, span id, and trace flags. The trace id should be the same since we manually set the parent trace context, but I wonder if we could somehow get into a weird state where we end up with a different trace flag in one of the iterations of creating this Activity for the orchestration execution. Not sure how that would be possible but thought to mention it just in case

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: `Stop()` *should* be sufficient, but `Dispose()` could work here, too.  (We don't generally dispose of the activity in this method to avoid the implicit `Stop()`, which we don't want because the activity needs to span multiple calls to this method.  *Not* disposing *should* be ok, as the base `Activity` doesn't actually release any resources.)

src/Shared/Grpc/ProtoUtils.cs
sophiatev: Yeah I see what you mean, it just makes me quite nervous that we're having to build these manually, and that we're altering the parent trace context to now hold an injected client span id rather than the parent trace activity's span id. An alternative approach is to add an additional field to the proto object corresponding to the `clientSpanId`, and just set the `TraceParent` as `Activity.Id` as you normally would. If anything, I'd say that this is truer to the original contract: other SDKs accessing this object would expect it hold the span id of the parent trace Activity, not an injected client span id. 

All that being said, I do see that DT.Core builds a traceparent manually in an identical [way](https://github.com/Azure/durabletask/blob/750d78f12d660f518f8a8b412ba6abf22314d203/src/DurableTask.Core/ISupportsDurableTraceContext.cs#L65) so maybe it's not a cause for concern

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: Good catch; I think that was just missed as the logic was moved a bit from where it is in DTFx.  Added (with tests).

src/Shared/Grpc/ProtoUtils.cs
philliphoff: Right, like DTFx, the `ExecutionStartedEvent` event is intended to house the "logical" orchestration execution span properties.  (My protobuf change adds one missing field.)  In DTFx, on the first orchestration execution, that information is persisted in the existing event which is sent back to the scheduler (to be persisted and returned on subsequent runs).  However, for DT, there's no current provision for sending that information back to the scheduler (as the `ExecutionStartedEvent` is not round-tripped).  

My protobuf change adds those missing fields to the `OrchestratorResponse` message instead.  The *scheduler* is then obligated to ensure it makes its way back into subsequent orchestration executions' `ExecutionStartedEvent`.  DTS will store it...somewhere--I haven't completely decided.  My PoC stuffs it in a orchestration tag, for example, and then pulls it out on subsequent executions.  If a scheduler does *not* do so, then each execution acts as the "first" and I think you should get a series of subspans beneath the orchestration client span--not as pretty but should still be useful.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: I want/need the protobuf version of the event with the extended properties, not the DTFx Core version.  I tried using the latter and it didn't seem to work out.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: To be honest, I couldn't see why DTFx had that check; AFAICT none of the sub-activities really depend on an existing `Activity.Current`; the spans are just as valid, if not as pretty, in a tracelog.  If tracing is enabled for the source, we should generally always have a parent orchestration-level activity.

src/Shared/Grpc/ProtoUtils.cs
philliphoff: That's where I got it from, so I figured it ought to work at least as well here and I was trying to minimize the changes needed to the protos.  That said, I'm also not crazy about having to build the values by hand.

src/Worker/Grpc/GrpcOrchestrationRunner.cs
philliphoff: I was primarily targeting these changes at DT used separately from DF. I will see if I can test the two together.  I would expect that DF uses a differently-named activity source, though, so overlap should be ok.  I wouldn't want to *not* trace something in DT *because* it's being traced in DF.

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,452 @@
...
+    public static Activity? StartTraceActivityForOrchestrationExecution(
+        P.ExecutionStartedEvent? startEvent,
+        P.OrchestrationTraceContext? orchestrationTraceContext)
+    {
+        if (startEvent == null)
+        {
+            return null;
+        }
+
+        if (startEvent.ParentTraceContext is null || !ActivityContext.TryParse(startEvent.ParentTraceContext.TraceParent, startEvent.ParentTraceContext.TraceState, out ActivityContext activityContext))
```
cgillum: nit: can you wrap these long lines? Per the .editorconfig file (which not all IDEs support natively, though VSCode.dev seems to), we try to keep line length to a maximum of 120 characters.

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,452 @@
...
+
+namespace Microsoft.DurableTask.Tracing;
+
+/// <summary>
+/// Methods for starting and managing trace activities related to Durable Task operations.
+/// </summary>
+/// <remarks>
+/// Adapted from "https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Tracing/TraceHelper.cs".
+/// </remarks>
+class TraceHelper
```
cgillum: nit: make this `static class`.

src/Shared/Grpc/Tracing/TraceActivityConstants.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Modified from https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Tracing/TraceActivityConstants.cs.
+namespace Microsoft.DurableTask.Tracing;
+
+/// <summary>
+/// Constants for trace activity names used in Durable Task Framework.
+/// </summary>
+class TraceActivityConstants
```
cgillum: nit: make this `static class`

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -0,0 +1,452 @@
...
+            newActivity.SetTag(Schema.Task.InstanceId, createInstanceRequest.InstanceId);
+            newActivity.SetTag(Schema.Task.ExecutionId, createInstanceRequest.ExecutionId);
+
+            if (!string.IsNullOrEmpty(createInstanceRequest.Version))
+            {
+                newActivity.SetTag(Schema.Task.Version, createInstanceRequest.Version);
+            }
+        }
+
+        if (Activity.Current is not null)
```
cgillum: nit: rather than checking separately for `Activity.Current` being null, could we just move this logic into the previous `newActivity != null` code block?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -356,6 +359,91 @@ async Task OnRunOrchestratorAsync(
...
+                    .Select(e => e.ExecutionStarted)
+                    .FirstOrDefault();
+
+            Activity? traceActivity = TraceHelper.StartTraceActivityForOrchestrationExecution(executionStartedEvent, request.OrchestrationTraceContext);
+
+            if (executionStartedEvent is not null)
+            {
+                P.HistoryEvent? GetSuborchestrationInstanceCreatedEvent(int eventId)
+                {
+                    var subOrchestrationEvent =
```
cgillum, line 381: nit: per the existing conventions in this repo, please try to use explicit type names instead of `var`. It makes the code easier to understand.

cgillum (commented):
A few nit-picks as I go through the PR. I have not yet reviewed everything.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceActivityConstants.cs
philliphoff: Updated.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: Yes, this makes sense.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
philliphoff: I believe I've updated all of these.

src/Shared/Grpc/Tracing/TraceHelper.cs
philliphoff: I believe I've gotten all of these.

philliphoff:
@sophiatev Is there more feedback on this change?

sophiatev (approved):
Looks good to me, but just a few questions:
1. How did we actually get this working? When I was trying to do add distributed tracing for entities I had to do a lot of work-arounds and add most of it to DT.Core because we seemingly couldn't add an Activity source to the dotnet SDK at the time. Did that change? (I wish I would have done my work after you did then :D. It would have improved my implementation by quite a bit)
2. Does the user have to enable distributed tracing in their settings somehow to get this tracing? 
3. Do we want to make an Activity in the `GrpcOrchestrationRunner` after all?

philliphoff:
> How did we actually get this working? When I was trying to do add distributed tracing for entities I had to do a lot of work-arounds and add most of it to DT.Core because we seemingly couldn't add an Activity source to the dotnet SDK at the time. Did that change?

@sophiatev I got it working/tested by creating an app that references a private build of the SDK and enables this new activity source (`Microsoft.DurableTask`), and then a backend that respects the protobuf changes that allow the context to be properly passed (changes coming to DTS once this goes in).  I'm not sure what issues you ran into before, but AFAIK, any 3rd party can add activity sources.

> Does the user have to enable distributed tracing in their settings somehow to get this tracing?

Yes, users collecting traces would have to add `Microsoft.DurableTask` as a source. 

> Do we want to make an Activity in the GrpcOrchestrationRunner after all?

Are you referring to the earlier comment about Durable Functions also having an activity for that method, or something else?  If the former, I'd say yes; the DF traces are from a different source so we'd want to ensure it's logged when DT is used on its own.

@sophiatev With tentative approval here, we can now commit the corresponding protobuf changes.  That will then mean a small update to this PR to bring them across the "proper way", which will need a re-approval.

sophiatev:
> Yes, users collecting traces would have to add Microsoft.DurableTask as a source.

Perhaps a silly question, but have you tested _without_ doing this to make sure that nothing breaks? I made the mistake of not doing so (in my case, not trying a run without distributed tracing enabled) which ended up in a null exception for customers. Want to make sure other's don't repeat my mistakes :D 

> Are you referring to the earlier comment about Durable Functions also having an activity for that method, or something else? If the former, I'd say yes; the DF traces are from a different source so we'd want to ensure it's logged when DT is used on its own.

Yes, I am. But it looks like the `GrpcOrchestrationRunner` is still not creating an `Activity`? Doesn't this imply that it should? (Although I'm not sure this is called when DT is used on its own, maybe there's a scenario where it will be in the future?)

philliphoff:
> Perhaps a silly question, but have you tested without doing this to make sure that nothing breaks? I made the mistake of not doing so (in my case, not trying a run without distributed tracing enabled) which ended up in a null exception for customers. Want to make sure other's don't repeat my mistakes :D

@sophiatev That's a good test case; yes, I've verified that execution proceeds when tracing has not been enabled.

> Yes, I am. But it looks like the GrpcOrchestrationRunner is still not creating an Activity? Doesn't this imply that it should? (Although I'm not sure this is called when DT is used on its own, maybe there's a scenario where it will be in the future?)

I verified that `GrpcOrchestrationRunner` (and `GrpcEntityRunner`) are *not* called within the SDK and *only* called by the DF extension (and no other external libraries as far as GH can find).  At least for now, I'm inclined to leave tracing out of them, given the specific focus of this PR on DT, and it'd be a different source from whatever AF/DF was doing anyway, plus there doesn't seem to be an equivalent for activities so there would still be gaps in the DT traces?  If and when it is needed to provide a coherent tracing story for DF, I'd be open to adding them, but I feel like that's a completely different set of tests and fixes.

philliphoff:
@sophiatev Is there a regular schedule/process for new releases?

PR #443, opened 2025-07-01
Add ability to filter orchestrations at worker

halspang:
This change allows a user to set a lambda that acts a filter for orchestrations. If the lambda returns false, the orchestration will be abandoned, leaving it to be attempted again later.

In this initial commit, the filter can work off of the orchestration name and the orchestration tags.

### DIFF
--- a/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
@@ -109,4 +109,21 @@ public static class DurableTaskWorkerBuilderExtensions
         });
         return builder;
     }
+
+    /// <summary>
+    /// Adds an orchestration filter to the specified <see cref="IDurableTaskWorkerBuilder"/>.
+    /// </summary>
+    /// <param name="builder">The builder to set the builder target for.</param>
+    /// <param name="orchestrationFilter">The filter function that determines whether an orchestration should be processed.</param>
+    /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+    public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTaskWorkerBuilder builder, Func<OrchestrationInfo, bool> orchestrationFilter)
+    {
+        Check.NotNull(builder);
+        Check.NotNull(orchestrationFilter);
+        builder.Services.Configure<DurableTaskWorkerOptions>(builder.Name, options =>
+        {
+            options.OrchestrationFilter = orchestrationFilter;
+        });
+        return builder;
+    }
 }
--- a/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
@@ -116,7 +116,6 @@ public static class DurableTaskWorkerBuilderExtensions
     /// <param name="builder">The builder to set the builder target for.</param>
     /// <typeparam name="TOrchestrationFilter">The implementation of a <see cref="IOrchestrationFilter"/> that will be bound.</typeparam>
     /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
-
     [Obsolete("Experimental")]
     public static IDurableTaskWorkerBuilder UseOrchestrationFilter<TOrchestrationFilter>(this IDurableTaskWorkerBuilder builder) where TOrchestrationFilter : class, IOrchestrationFilter
     {
--- a/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
@@ -116,6 +116,8 @@ public static class DurableTaskWorkerBuilderExtensions
     /// <param name="builder">The builder to set the builder target for.</param>
     /// <typeparam name="TOrchestrationFilter">The implementation of a <see cref="IOrchestrationFilter"/> that will be bound.</typeparam>
     /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+
+    [Obsolete("Experimental")]
     public static IDurableTaskWorkerBuilder UseOrchestrationFilter<TOrchestrationFilter>(this IDurableTaskWorkerBuilder builder) where TOrchestrationFilter : class, IOrchestrationFilter
     {
         Check.NotNull(builder);
@@ -129,6 +131,7 @@ public static class DurableTaskWorkerBuilderExtensions
     /// <param name="builder">The builder to set the builder target for.</param>
     /// <param name="filter">The instance of an <see cref="IOrchestrationFilter"/> to use.</param>
     /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+    [Obsolete("Experimental")]
     public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTaskWorkerBuilder builder, IOrchestrationFilter filter)
     {
         Check.NotNull(builder);
--- a/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
@@ -114,16 +114,25 @@ public static class DurableTaskWorkerBuilderExtensions
     /// Adds an orchestration filter to the specified <see cref="IDurableTaskWorkerBuilder"/>.
     /// </summary>
     /// <param name="builder">The builder to set the builder target for.</param>
-    /// <param name="orchestrationFilter">The filter function that determines whether an orchestration should be processed.</param>
+    /// <typeparam name="TOrchestrationFilter">The implementation of a <see cref="IOrchestrationFilter"/> that will be bound.</typeparam>
     /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
-    public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTaskWorkerBuilder builder, Func<OrchestrationInfo, bool> orchestrationFilter)
+    public static IDurableTaskWorkerBuilder UseOrchestrationFilter<TOrchestrationFilter>(this IDurableTaskWorkerBuilder builder) where TOrchestrationFilter : class, IOrchestrationFilter
     {
         Check.NotNull(builder);
-        Check.NotNull(orchestrationFilter);
-        builder.Services.Configure<DurableTaskWorkerOptions>(builder.Name, options =>
-        {
-            options.OrchestrationFilter = orchestrationFilter;
-        });
+        builder.Services.AddSingleton<IOrchestrationFilter, TOrchestrationFilter>();
+        return builder;
+    }
+
+    /// <summary>
+    /// Adds an orchestration filter to the specified <see cref="IDurableTaskWorkerBuilder"/>.
+    /// </summary>
+    /// <param name="builder">The builder to set the builder target for.</param>
+    /// <param name="filter">The instance of an <see cref="IOrchestrationFilter"/> to use.</param>
+    /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+    public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTaskWorkerBuilder builder, IOrchestrationFilter filter)
+    {
+        Check.NotNull(builder);
+        builder.Services.AddSingleton(filter);
         return builder;
     }
 }
--- a/src/Worker/Core/DurableTaskWorkerOptions.cs
+++ b/src/Worker/Core/DurableTaskWorkerOptions.cs
@@ -50,6 +50,22 @@ public class DurableTaskWorkerOptions
     }
 
     /// <summary>
+    /// Struct representation of orchestration information.
+    /// </summary>
+    public struct OrchestrationInfo
+    {
+        /// <summary>
+        /// Gets the name of the orchestration.
+        /// </summary>
+        public string Name { get; init; }
+
+        /// <summary>
+        /// Gets the tags associated with the orchestration.
+        /// </summary>
+        public Dictionary<string, string> Tags { get; init; }
+    }
+
+    /// <summary>
     /// Gets or sets the data converter. Default value is <see cref="JsonDataConverter.Default" />.
     /// </summary>
     /// <remarks>
@@ -157,6 +173,11 @@ public class DurableTaskWorkerOptions
     internal bool DataConverterExplicitlySet { get; private set; }
 
     /// <summary>
+    /// Gets or sets a callback function that determines whether an orchestration should be accepted for work.
+    /// </summary>
+    public Func<OrchestrationInfo, bool>? OrchestrationFilter { get; set; }
+
+    /// <summary>
     /// Applies these option values to another.
     /// </summary>
     /// <param name="other">The other options object to apply to.</param>
@@ -169,6 +190,7 @@ public class DurableTaskWorkerOptions
             other.MaximumTimerInterval = this.MaximumTimerInterval;
             other.EnableEntitySupport = this.EnableEntitySupport;
             other.Versioning = this.Versioning;
+            other.OrchestrationFilter = this.OrchestrationFilter;
         }
     }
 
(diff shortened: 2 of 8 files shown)


cgillum:
@halspang can you add a bit more context on how this is intended to be used in real-world situations?

Normally a filtering feature would be built on the server-side so that you don't waste cycles delivering and abandoning messages repeatedly (or infinitely in some cases). Is there a plan to add such functionality to DTS? And would the design change at all if/when we go down that path?

halspang:
> @halspang can you add a bit more context on how this is intended to be used in real-world situations?
> 
> Normally a filtering feature would be built on the server-side so that you don't waste cycles delivering and abandoning messages repeatedly (or infinitely in some cases). Is there a plan to add such functionality to DTS? And would the design change at all if/when we go down that path?

@cgillum - The general goal of this is to give users the ability to react to events that require they temporarily pause or block work. It's being done this way as it allows for users to hook in their own custom logic for how that filtering is achieved. The main scenario that I was presented to make this change was the ability to add a dynamic kill switch to a specific worker. 

I agree that it would be good to do this at a server level, and I've long wanted to add more worker knowledge to the backend. That's the more correct path, but also the much longer path. Especially so when it comes to integrating this kind of filtering as it would require persisting data between the worker and scheduler. I have ideas for this but, again, it'll just take a bit longer. 

Let me know what you think in terms of, should we do this to get the functionality out there and then follow up with backend support, or if we should just aim for the server-side solution from the start.

cgillum:
> The main scenario that I was presented to make this change was the ability to add a dynamic kill switch to a specific worker.

I think the scenario is interesting, but I'm curious to understand how one would operationally implement a kill switch that only targets a specific worker. Can we clarify what "specific worker" means? For example, in an AKS context, is this a specific _deployment_, a specific _pod_, or perhaps a specific pod _replica_ within the deployment?

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
```diff
@@ -109,4 +109,33 @@ public static IDurableTaskWorkerBuilder UseVersioning(this IDurableTaskWorkerBui
...
         return builder;
     }
+
+    /// <summary>
+    /// Adds an orchestration filter to the specified <see cref="IDurableTaskWorkerBuilder"/>.
+    /// </summary>
+    /// <param name="builder">The builder to set the builder target for.</param>
+    /// <typeparam name="TOrchestrationFilter">The implementation of a <see cref="IOrchestrationFilter"/> that will be bound.</typeparam>
+    /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+
```
cgillum: nit: remove extra whitespace

src/Worker/Core/DurableTaskWorkerOptions.cs
```diff
@@ -145,6 +145,11 @@ public DataConverter DataConverter
     /// </summary>
     public bool IsVersioningSet { get; internal set; }
 
+    /// <summary>
+    /// Gets or sets a callback function that determines whether an orchestration should be accepted for work.
+    /// </summary>
+    public IOrchestrationFilter? OrchestrationFilter { get; set; }
```
cgillum, line 152: This needs `[Obsolete("Experimental")]` as well.

src/Worker/Core/IOrchestrationFilter.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.Worker;
+
+/// <summary>
+/// Defines a filter for validating orchestrations.
+/// </summary>
+public interface IOrchestrationFilter
```
cgillum, line 10: This needs `[Obsolete("Experimental")]`. Same for all new public APIs.

src/Worker/Core/IOrchestrationFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <param name="info">The information on the orchestration to validate.</param>
+    /// <param name="cancellationToken">The cancellation token for the request to timeout.</param>
+    /// <returns><code>true</code> if the orchestration is valid <code>false</code> otherwise.</returns>
+    Task<bool> IsOrchestrationValidAsync(OrchestrationInfo info, CancellationToken cancellationToken = default);
+}
+
+/// <summary>
+/// Struct representation of orchestration information.
+/// </summary>
+public struct OrchestrationInfo
```
cgillum: Two things:
1. This public type needs `[Obsolete("Experimental")]`. Same for all new public APIs.
2. This type name is too generic. Consider a more specific name, like `OrchestrationFilter` or `OrchestrationFilterParameters`.

src/Worker/Core/IOrchestrationFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Gets the name of the orchestration.
+    /// </summary>
+    public string Name { get; init; }
+
+    /// <summary>
+    /// Gets the tags associated with the orchestration.
+    /// </summary>
+    public Dictionary<string, string> Tags { get; init; }
```
cgillum: This property should be declared as nullable if it's not required. Also, in order for this to be truly "read-only", you should use `IReadOnlyDictionary<string, string>` instead of `Dictionary<string, string>` since the latter is read-write.
```suggestion
    public IReadOnlyDictionary<string, string>? Tags { get; init; }
```

src/Worker/Core/IOrchestrationFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+
+/// <summary>
+/// Struct representation of orchestration information.
+/// </summary>
+public struct OrchestrationInfo
+{
+    /// <summary>
+    /// Gets the name of the orchestration.
+    /// </summary>
+    public string Name { get; init; }
```
cgillum: I assume this is a required property? If so, you should mark it as `required`.
```suggestion
    public required string Name { get; init; }
```

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -374,6 +377,31 @@ async Task OnRunOrchestratorAsync(
...
                     cancellationToken);
 
+                bool filterPassed = true;
+                if (this.orchestrationFilter != null)
+                {
+                    filterPassed = await this.orchestrationFilter.IsOrchestrationValidAsync(
+                        new OrchestrationInfo
+                        {
+                            Name = runtimeState.Name,
+                            Tags = new Dictionary<string, string>(runtimeState.Tags ?? ImmutableDictionary<string, string>.Empty),
```
cgillum: We shouldn't allocate a new dictionary object if `runtimeState.Tags` is null. Better to just leave the `Tags` property unassigned.

src/Worker/Core/IOrchestrationFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+/// </summary>
+public interface IOrchestrationFilter
+{
+    /// <summary>
+    /// Validate the orchestration against the filter represented by this interface.
+    /// </summary>
+    /// <param name="info">The information on the orchestration to validate.</param>
+    /// <param name="cancellationToken">The cancellation token for the request to timeout.</param>
+    /// <returns><code>true</code> if the orchestration is valid <code>false</code> otherwise.</returns>
+    Task<bool> IsOrchestrationValidAsync(OrchestrationInfo info, CancellationToken cancellationToken = default);
```
cgillum: Consider using `ValueTask<bool>` instead of `Task<bool>` since many implementations may not actually require an async implementation, making `ValueTask<T>` a more appropriate choice for performance.

src/Worker/Grpc/GrpcDurableTaskWorker.cs
```diff
@@ -48,7 +49,8 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
         await using AsyncDisposable disposable = this.GetCallInvoker(out CallInvoker callInvoker, out string address);
         this.logger.StartingTaskHubWorker(address);
-        await new Processor(this, new(callInvoker)).ExecuteAsync(stoppingToken);
+        IOrchestrationFilter? filter = this.services.GetService<IOrchestrationFilter>();
```
cgillum: Since `IOrchestrationFilter` is registered as a singleton, it will make more sense and be more efficient to resolve this as part of the constructor. As far as I can tell, the `GrpcDurableTaskWorker` is created via dependency-injection, so you can just add `IOrchestrationFilter` as an optional constructor parameter.

src/Worker/Grpc/Logs.cs
```diff
@@ -57,5 +57,8 @@ static partial class Logs
 
         [LoggerMessage(EventId = 58, Level = LogLevel.Information, Message = "Abandoning orchestration. InstanceId = '{instanceId}'. Completion token = '{completionToken}'")]
         public static partial void AbandoningOrchestrationDueToVersioning(this ILogger logger, string instanceId, string completionToken);
+
+        [LoggerMessage(EventId = 59, Level = LogLevel.Information, Message = "Abandoning orchestration due to filtering. InstanceId = '{instanceId}'. Completion token = '{completionToken}'")]
+        public static partial void AbandoningOrchestrationDueToOrchestrationFilter(this ILogger logger, string instanceId, string completionToken);
```
cgillum, line 62: Should we include information about why it was filtered for debugging purposes? For example, because it matched name "Foo" or tag "Bar"?

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -970,16 +969,99 @@ public async Task RunActivityWithTags()
...
+        });
+        OrchestrationMetadata metadata = await server.Client.WaitForInstanceCompletionAsync(
+            instanceId, this.TimeoutToken);
+
+        Assert.NotNull(metadata);
+        Assert.Equal(instanceId, metadata.InstanceId);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, metadata.RuntimeStatus);
+
+        // Update the filter and re-enqueue. We should see the orchestration denied.
+        orchestrationFilter.TagDenyDict.Add("test", "true");
```
cgillum, line 1078: We need to add another test which validates the non-match scenario. For example:
```csharp
orchestrationFilter.TagDenyDict.Add("test", "false");
```

src/Worker/Core/IOrchestrationFilter.cs
cgillum: Question: What are the intended semantics as far as case-sensitivity for keys and values in this dictionary?

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
halspang: Done.

src/Worker/Core/DurableTaskWorkerOptions.cs
halspang: Done.

src/Worker/Core/IOrchestrationFilter.cs
halspang: Done.

src/Worker/Core/IOrchestrationFilter.cs
halspang: Done, went for `OrchestrationFilterParameters`.

src/Worker/Core/IOrchestrationFilter.cs
halspang: Updated.

src/Worker/Core/IOrchestrationFilter.cs
halspang: Actually, it should be nullable like the Tags. You could match on name, tags, or both. It's up to the user to decide what info to use.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
halspang: Updated.

src/Worker/Core/IOrchestrationFilter.cs
halspang: This is good to know, I often forget about the difference between the two.

src/Worker/Grpc/Logs.cs
halspang: We won't actually know what it matched as that happens in the user implementation of the interface.

src/Worker/Core/IOrchestrationFilter.cs
halspang: Again, this is up to the user. We provide an interface and the information, they can make whatever decision they want with it.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: Done.

src/Worker/Grpc/GrpcDurableTaskWorker.cs
halspang: Updated.

halspang:
> > The main scenario that I was presented to make this change was the ability to add a dynamic kill switch to a specific worker.
> 
> I think the scenario is interesting, but I'm curious to understand how one would operationally implement a kill switch that only targets a specific worker. Can we clarify what "specific worker" means? For example, in an AKS context, is this a specific _deployment_, a specific _pod_, or perhaps a specific pod _replica_ within the deployment?

That all depends on how they implement the interface. Deployment is easiest as you could just be deploying a different image with specific logic about how to approve/deny orchestrations. You could also pass in the pod/replica ID as an env var and use that in your service. We left it very open so the user can make all the choices here.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -1009,6 +1009,41 @@ public async Task FilterOrchestrationsByName()
...
+        string instanceId = await server.Client.ScheduleNewOrchestrationInstanceAsync(orchestratorName);
+        OrchestrationMetadata metadata = await server.Client.WaitForInstanceCompletionAsync(
+            instanceId, this.TimeoutToken);
+
+        Assert.NotNull(metadata);
+        Assert.Equal(instanceId, metadata.InstanceId);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, metadata.RuntimeStatus);
+
+        // Update the filter and re-enqueue. We should see the orchestration denied.
+        orchestrationFilter.NameDenySet.Add($"not-{orchestratorName}");
```
cgillum, line 1036: Is this correct? It seems to me that the comment doesn't match the code. You're adding a Deny rule for an orchestrator name that doesn't exist, so I wouldn't expect anything to be denied.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -1049,16 +1084,62 @@ public async Task FilterOrchestrationsByTag()
...
+        });
+        OrchestrationMetadata metadata = await server.Client.WaitForInstanceCompletionAsync(
+            instanceId, this.TimeoutToken);
+
+        Assert.NotNull(metadata);
+        Assert.Equal(instanceId, metadata.InstanceId);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, metadata.RuntimeStatus);
+
+        // Update the filter and re-enqueue. We should see the orchestration denied.
+        orchestrationFilter.TagDenyDict.Add("test", "false");
```
cgillum, line 1119: Same comment as above, it seems like the orchestration would NOT be denied since it doesn't match the filter, and the filter is described as a deny rule.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: Oops, yes. Copy-pasted and didn't update the comment.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
halspang: Same.

test/Grpc.IntegrationTests/OrchestrationPatterns.cs
```diff
@@ -970,16 +969,180 @@ public async Task RunActivityWithTags()
...
+        Assert.NotNull(metadata);
+        Assert.Equal(instanceId, metadata.InstanceId);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, metadata.RuntimeStatus);
+
+        // Update the filter and re-enqueue. We should see the orchestration denied.
+        orchestrationFilter.NameDenySet.Add(orchestratorName);
+
+        // This should throw as the work is denied.
+        instanceId = await server.Client.ScheduleNewOrchestrationInstanceAsync(orchestratorName);
+        await Assert.ThrowsAsync<OperationCanceledException>(async () => await server.Client.WaitForInstanceCompletionAsync(instanceId, new CancellationTokenSource(TimeSpan.FromSeconds(5)).Token));
```
cgillum: `CancellationTokenSource` is disposable, so you should always write your code in a way that disposes it after you're doing with it. For example:
```csharp
using CancellationTokenSource timeoutCts = new(TimeSpan.FromSeconds(5));
await Assert.ThrowsAsync<OperationCanceledException>(async () => await server.Client.WaitForInstanceCompletionAsync(instanceId, timeoutCts.Token));
```
The same applies in your other test cases.

halspang: Well that's just an oversight on my part :( Fixed.

cgillum (approved):

PR #444, opened 2025-07-02
Adding Analyzer to build and release

bachuv:
This PR includes the Analyzer in the build and release process. It also updates src/Analyzers/AnalyzerReleases.Shipped.md to list which rules are shipped.


src/Analyzers/Analyzers.csproj
```diff
@@ -12,12 +12,12 @@
 
   <PropertyGroup>
     <VersionPrefix>0.0.1</VersionPrefix>
-    <VersionSuffix>preview.1</VersionSuffix>
+    <VersionSuffix>preview</VersionSuffix>
```
bachuv: Are we releasing this as a preview package for now or should I remove `preview`?

src/Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -1,20 +1,2 @@
...
-DURABLE0006 | Orchestration | Warning | EnvironmentOrchestrationAnalyzer
-DURABLE0007 | Orchestration | Warning | CancellationTokenOrchestrationAnalyzer
-DURABLE0008 | Orchestration | Warning | OtherBindingsOrchestrationAnalyzer
-DURABLE1001 | Attribute Binding | Error | OrchestrationTriggerBindingAnalyzer
-DURABLE1002 | Attribute Binding | Error | DurableClientBindingAnalyzer
-DURABLE1003 | Attribute Binding | Error | EntityTriggerBindingAnalyzer
-DURABLE2001 | Activity | Warning | MatchingInputOutputTypeActivityAnalyzer
-DURABLE2002 | Activity | Warning | MatchingInputOutputTypeActivityAnalyzer
\ No newline at end of file
+; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
```
nytian: I am not sure what this md is used for and why we keep this md file. Can't we just remove it and only keep the shipped md?

src/Analyzers/AnalyzerReleases.Shipped.md
```diff
@@ -1,3 +1,20 @@
 ﻿; Shipped analyzer releases
 ; https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
 
+### New Rules
```
nytian, line 6: Please fix the warnings

src/Analyzers/Analyzers.csproj
nytian: if we say GA this analyzer, I think we should remove the preview.

bachuv: @lilyjma - are we releasing the .NET Isolated Analyzer as preview or GA?

src/Analyzers/AnalyzerReleases.Unshipped.md
bachuv: For analyzers, it's standard to have 2 .md files for shipped and unshipped rules. Reference: https://github.com/dotnet/roslyn-analyzers/tree/7633ceddc23f29ac9a6c844c6e5fad5f52e1ebc9/src/NetAnalyzers/Core

src/Analyzers/AnalyzerReleases.Shipped.md
bachuv: fixed

src/Analyzers/Analyzers.csproj
lilyjma: Is there any breaking/major change or additions we anticipate?

bachuv: I don't know of any major changes that we would need to make. From what I understand, we would just keep adding analyzer rules to the existing implementation.

bachuv: I updated the version to 0.1.0.

src/Analyzers/Analyzers.csproj
```diff
@@ -11,13 +11,13 @@
...
 
   <PropertyGroup>
-    <VersionPrefix>0.0.1</VersionPrefix>
-    <VersionSuffix>preview.1</VersionSuffix>
+    <VersionPrefix>0.1.0</VersionPrefix>
+    <VersionSuffix></VersionSuffix>
     <PackageDescription>.NET Analyzers for the Durable Task SDK.</PackageDescription>
     <NeutralLanguage>en</NeutralLanguage>
-    <IncludeBuildOutput>false</IncludeBuildOutput> <!-- Do not include the analyzer as a lib dependency -->
+    <IncludeBuildOutput>true</IncludeBuildOutput>
```
cgillum: What does this change do?

bachuv: This change includes the compiled outputs (.dll, .xml, .pdb, etc.) in the NuGet package under the `lib/` directory. I actually reverted this change because it's not recommended for analyzers. For analyzers, it's recommended to put the analyzer DLL under `analyzers/` so the analyzer shows up as a live analyzer instead of a reference library (we do this on line 31 of this .csproj).

I also checked the DF .NET In Proc analyzer and we set `IncludeBuildOutput` to `false` - https://github.com/Azure/azure-functions-durable-extension/blob/9b39310c869c8e02e3d3b7963e70803c9d53cc9d/src/WebJobs.Extensions.DurableTask.Analyzers/WebJobs.Extensions.DurableTask.Analyzers.csproj#L5

Note: I tested the analyzer with `IncludeBuildOutput` set to `true` in a DF .NET Isolated app where I didn't follow some orchestrator code constrations and I saw warnings so there must be some other features that don't work when `IncludeBuildOutput` is set to `true`.

cgillum (approved):

PR #445, opened 2025-07-08
Move from net6 to net8 target

halspang:
.NET 6 reached end of life in November of 2024. Updating the target frameworks to reference .NET 8.


halspang:
> Looks great so far. I think the next step is to take a look at our Directory.Packages.props file at the repo root and replace all the `System.` and `Microsoft.` packages that target 6.0 versions and replace them with the latest 8.0 versions.

Didn't think about that past the framework! Updated :)

cgillum (approved):

PR #446, opened 2025-07-10
Removing breaking change for TaskOptions

sophiatev:
A breaking change was introduced in `TaskOptions` where there are now two constructors which both have all optional parameters. Now creating a new `TaskOption()` with no arguments is ambiguous between the two constructors. One constructor was changed to have its parameter be non-optional, whereas the other constructor remains as having all optional parameters. This should make creating a new `TaskOption()` unambiguous (it will correspond to the second constructor with two optional parameters).

### DIFF
--- a/.github/workflows/validate-build.yml
+++ b/.github/workflows/validate-build.yml
@@ -23,13 +23,13 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v3
+    - name: Setup .NET 8.0
+      uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: '8.0.x'
 
     - name: Setup .NET from global.json
-      uses: actions/setup-dotnet@v3
+      uses: actions/setup-dotnet@v4
       with:
         global-json-file: global.json
 
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,15 +9,15 @@
 
   <!-- Microsoft.Extensions.* Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="6.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="8.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="8.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.2" />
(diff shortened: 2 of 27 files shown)


cgillum:
@sophiatev thanks for this PR! Looking at the code history, I think there might be a more correct change. The first constructor was originally the only constructor, so I think a safer fix would be to just change the second constructor (which is the new one) to make both parameters required. Your current fix solves the ambiguity problem, but I'm not 100% certain that it will solve the binary compatibility problem.

Also adding @YunchuWang for visibility since it was #426 that introduced this change.

sophiatev:
> @sophiatev thanks for this PR! Looking at the code history, I think there might be a more correct change. The first constructor was originally the only constructor, so I think a safer fix would be to just change the second constructor (which is the new one) to make both parameters required. Your current fix solves the ambiguity problem, but I'm not 100% certain that it will solve the binary compatibility problem.
> 
> Also adding @YunchuWang for visibility since it was #426 that introduced this change.

@YunchuWang is there ever a situation where you would want to allow just specifying a `tags` parameter but not a `retry` parameter? Just want to double-check that it's okay if we make both parameters required for the new `TaskOptions` constructor you introduced. 

Also adding @jviau since he suggested the initial fix and has a much better understanding of this binary compatibility stuff than I do.

src/Abstractions/TaskOptions.cs
bachuv, line 1: Can you add a test for this change?

src/Abstractions/TaskOptions.cs
```diff
@@ -14,7 +14,7 @@ public record TaskOptions
     /// Initializes a new instance of the <see cref="TaskOptions"/> class.
     /// </summary>
     /// <param name="retry">The task retry options.</param>
-    public TaskOptions(TaskRetryOptions? retry = null)
+    public TaskOptions(TaskRetryOptions? retry)
```
jviau, line 17: You could also change this to:


```suggestion
    public TaskOptions(TaskRetryOptions? retry)
      : this(retry, null)
    {
    }
```

jviau (approved):

jviau:
>  but I'm not 100% certain that it will solve the binary compatibility problem.

There was no binary compatibility problem with the original change. Only a new ambiguity issue at compilation time. The proposed changes resolve the ambiguity problem and do not affect the binary compatibility (so all good there).

With that said, you would prefer the 2nd ctor to have optional params removed that is fine as well. Its a question of, do you want `new(tags: <...>)` to work or not? But honestly, tags have `init` on them, so it isn't a big deal to have it via ctor. You can always do `new() { Tags = <...> }`

test/Abstractions.Tests/TaskOptionsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        SubOrchestrationOptions subOptions = new();
+        subOptions.Retry.Should().BeNull();
+        subOptions.Tags.Should().BeNull();
+        subOptions.InstanceId.Should().BeNull();
+
+        StartOrchestrationOptions startOptions = new();
+        startOptions.Version.Should().BeNull();
+        startOptions.InstanceId.Should().BeNull();
+        startOptions.StartAt.Should().BeNull();
+        startOptions.Tags.Should().BeEmpty();
```
cgillum, line 24: Weird that tags defaults to empty here but defaults to null in the *Options classes. Oh well - not something to deal with in this PR.

bachuv (approved):

PR #447, opened 2025-07-11
Expose gRPC client retry options

halspang:
This commit allows for users to specify their own retry options instead of relying on the defaults we provide. The general use case here is to enable longer retry periods, more retries, and different status codes to retry on.

### DIFF
--- a/src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
+++ b/src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
@@ -41,35 +41,6 @@ public static class DurableTaskSchedulerClientExtensions
     }
 
     /// <summary>
-    /// Configures Durable Task client to use the Azure Durable Task Scheduler service.
-    /// </summary>
-    /// <param name="builder">The Durable Task client builder to configure.</param>
-    /// <param name="endpointAddress">The endpoint address of the Durable Task Scheduler resource. Expected to be in the format "https://{scheduler-name}.{region}.durabletask.io".</param>
-    /// <param name="taskHubName">The name of the task hub resource associated with the Durable Task Scheduler resource.</param>
-    /// <param name="credential">The credential used to authenticate with the Durable Task Scheduler task hub resource.</param>
-    /// <param name="retryOptions">The options that determine how and when a request will be retried.</param>
-    /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerClientOptions.</param>
-    public static void UseDurableTaskScheduler(
-        this IDurableTaskClientBuilder builder,
-        string endpointAddress,
-        string taskHubName,
-        TokenCredential credential,
-        DurableTaskSchedulerClientOptions.ClientRetryOptions retryOptions,
-        Action<DurableTaskSchedulerClientOptions>? configure = null)
-    {
-        ConfigureSchedulerOptions(
-            builder,
-            options =>
-            {
-                options.EndpointAddress = endpointAddress;
-                options.TaskHubName = taskHubName;
-                options.Credential = credential;
-                options.RetryOptions = retryOptions;
-            },
-            configure);
-    }
-
-    /// <summary>
     /// Configures Durable Task client to use the Azure Durable Task Scheduler service using a connection string.
     /// </summary>
     /// <param name="builder">The Durable Task client builder to configure.</param>
@@ -89,33 +60,6 @@ public static class DurableTaskSchedulerClientExtensions
                 options.TaskHubName = connectionOptions.TaskHubName;
                 options.Credential = connectionOptions.Credential;
                 options.AllowInsecureCredentials = connectionOptions.AllowInsecureCredentials;
-            },
-            configure);
-    }
-
-    /// <summary>
-    /// Configures Durable Task client to use the Azure Durable Task Scheduler service using a connection string.
-    /// </summary>
-    /// <param name="builder">The Durable Task client builder to configure.</param>
-    /// <param name="connectionString">The connection string used to connect to the Durable Task Scheduler service.</param>
-    /// /// <param name="retryOptions">The options that determine how and when a request will be retried.</param>
-    /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerClientOptions.</param>
-    public static void UseDurableTaskScheduler(
-        this IDurableTaskClientBuilder builder,
-        string connectionString,
-        DurableTaskSchedulerClientOptions.ClientRetryOptions retryOptions,
-        Action<DurableTaskSchedulerClientOptions>? configure = null)
-    {
-        var connectionOptions = DurableTaskSchedulerClientOptions.FromConnectionString(connectionString);
-        ConfigureSchedulerOptions(
-            builder,
-            options =>
-            {
-                options.EndpointAddress = connectionOptions.EndpointAddress;
-                options.TaskHubName = connectionOptions.TaskHubName;
-                options.Credential = connectionOptions.Credential;
-                options.AllowInsecureCredentials = connectionOptions.AllowInsecureCredentials;
-                options.RetryOptions = retryOptions;
             },
             configure);
     }
--- a/src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
+++ b/src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
@@ -7,7 +7,7 @@ using Azure.Identity;
 using Grpc.Core;
 using Grpc.Net.Client;
 using Microsoft.DurableTask.Client;
-using grpcConfig = Grpc.Net.Client.Configuration;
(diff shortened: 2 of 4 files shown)


src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
```diff
@@ -40,15 +40,70 @@ public static void UseDurableTaskScheduler(
...
+    /// <param name="taskHubName">The name of the task hub resource associated with the Durable Task Scheduler resource.</param>
+    /// <param name="credential">The credential used to authenticate with the Durable Task Scheduler task hub resource.</param>
+    /// <param name="retryOptions">The options that determine how and when a request will be retried.</param>
+    /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerClientOptions.</param>
+    public static void UseDurableTaskScheduler(
+        this IDurableTaskClientBuilder builder,
+        string endpointAddress,
+        string taskHubName,
+        TokenCredential credential,
+        DurableTaskSchedulerClientOptions.ClientRetryOptions retryOptions,
```
jviau: Are these overloads necessary? Caller can easily include it within the `configure` callback already present.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -6,8 +6,8 @@
 using Azure.Identity;
 using Grpc.Core;
 using Grpc.Net.Client;
-using Microsoft.DurableTask;
 using Microsoft.DurableTask.Client;
+using grpcConfig = Grpc.Net.Client.Configuration;
```
cgillum: nit: Let's use Pascal casing here, which is the standard for namespaces.
```suggestion
using GrpcConfig = Grpc.Net.Client.Configuration;
```

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -46,6 +46,11 @@ public class DurableTaskSchedulerClientOptions
     /// </summary>
     public bool AllowInsecureCredentials { get; set; }
 
+    /// <summary>
+    /// Gets or sets the options that determine how and when calls made to the scheduler will be retried.
+    /// </summary>
+    internal ClientRetryOptions? RetryOptions { get; set; }
```
cgillum: Any reason not to make this public?

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -109,6 +114,48 @@ this.Credential is not null
...
+            };
+
+            serviceConfig = new grpcConfig.ServiceConfig
+            {
+                MethodConfigs = { methodConfig },
+            };
+        }
+        else
+        {
+            serviceConfig = GrpcRetryPolicyDefaults.DefaultServiceConfig;
```
cgillum: nit: consider doing this assignment on line 117 so that you can drop this `else` condition.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -109,6 +114,48 @@ this.Credential is not null
...
+                        continue;
+                    }
+
+                    retryPolicy.RetryableStatusCodes.Add(statusCode);
+                }
+            }
+
+            grpcConfig.MethodConfig methodConfig = new grpcConfig.MethodConfig
+            {
+                Names = { grpcConfig.MethodName.Default },
```
cgillum: Might be worth adding a comment explaining why what we're doing here is correct

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
cgillum: Rather than creating a new overload with a dedicated parameter for `retryOptions`, customers should just use the existing overload above and specify retry options in the `configure` action.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
jviau: Any specific reason to not make this `public`? Having it public will allow users to configure this through general `IOptions` patterns. Keeping it internal limits them to the explicit methods you offer, which don't support reading from config, depending on other options, etc.

If you do make it public, consider also having it be non-null, giving it a default value at the start. This way callers don't need to null-check before assigning properties to it.

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
```diff
@@ -40,15 +40,70 @@ public static void UseDurableTaskScheduler(
...
     /// Configures Durable Task client to use the Azure Durable Task Scheduler service using a connection string.
     /// </summary>
     /// <param name="builder">The Durable Task client builder to configure.</param>
     /// <param name="connectionString">The connection string used to connect to the Durable Task Scheduler service.</param>
+    /// /// <param name="retryOptions">The options that determine how and when a request will be retried.</param>
     /// <param name="configure">Optional callback to dynamically configure DurableTaskSchedulerClientOptions.</param>
     public static void UseDurableTaskScheduler(
         this IDurableTaskClientBuilder builder,
         string connectionString,
+        DurableTaskSchedulerClientOptions.ClientRetryOptions retryOptions,
```
cgillum: Same comment as above. I don't think we need a new `UseDurableTaskScheduler` method overload for this. The existing methods should be usable as-is.

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
halspang: I was following what I thought was the convention of adding extra constructors for ease of use. If we don't need that, happy to use the `configure`.

halspang: I was following what I thought was the convention of adding extra constructors for ease of use. If we don't need that, happy to use the `configure`.

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
halspang: See above.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
halspang: Oops, fixed.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
halspang: My original change had this as the actual gRPC retry policy, which I wanted to obscure from users with the extension methods. I forgot to remake this public after I changed to using the options you see now.

halspang: My original change had this as the actual gRPC retry policy, which I wanted to obscure from users with the extension methods. I forgot to remake this public after I changed to using the options you see now.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
halspang: Done.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
halspang: Done.

src/Client/AzureManaged/DurableTaskSchedulerClientExtensions.cs
cgillum: Adding extra constructors in most cases is only for preserving backwards binary compatibility. In general, it's better to have fewer APIs (fewer things to document) so we try to add new overloads only when strictly necessary.

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
```diff
@@ -173,4 +217,35 @@ this.Credential is not null
...
+
+        /// <summary>
+        /// Gets or sets the backoff multiplier for exponential backoff.
+        /// </summary>
+        public double? BackoffMultiplier { get; set; }
+
+        /// <summary>
+        /// Gets or sets the list of status codes that can be retried.
+        /// </summary>
+        public List<StatusCode>? RetryableStatusCodes { get; set; }
```
cgillum: Optional: Consider using an interface here, like `IList<StatusCode>?` or `ICollection<StatusCode>?`. This increases the number of list/collection types that customers can use to configure this property.

cgillum (approved):

src/Client/AzureManaged/DurableTaskSchedulerClientOptions.cs
halspang: Done :)

cgillum (approved):

PR #448, opened 2025-07-11
Add System.Text.Json to Shared.csproj

andystaples:
NuGet dynamically resolves System.Text.Json to 6.0.0 in Shared.csproj even though it is defined at 8.0.6 in CPM. 
This causes flags in S360 even though this projects dependencies are not built into any of the released components. 
This change intentionally resolves the CPM version of this dependency for this project

### DIFF
--- a/src/Shared/Shared.csproj
+++ b/src/Shared/Shared.csproj
@@ -17,8 +17,6 @@
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
     <PackageReference Include="Microsoft.Extensions.Hosting" />
-    <!-- For some reason, this project resolves this at 6.0.0 instead of the centrally defined 8.0.6. 
-         Since 6.0.0 has known vulnerabilities, we override to 6.0.11 to avoid major version changes here. -->
     <PackageReference Include="System.Text.Json"/>
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


src/Shared/Shared.csproj
```diff
@@ -17,6 +17,9 @@
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
     <PackageReference Include="Microsoft.Extensions.Hosting" />
+    <!-- For some reason, this project resolves this at 6.0.0 instead of the centrally defined 8.0.6. 
+         Since 6.0.0 has known vulnerabilities, we override to 6.0.11 to avoid major version changes here. -->
+    <PackageReference Include="System.Text.Json" VersionOverride="6.0.11"/>
```
cgillum: Do we need `VersionOverride` or can we just mention the package name and rely on the central package management config to get the version?

andystaples: Removing VersionOverride, you get 8.0.6. This is a major version change from the previously resolved 6.0.0, but if we're not concerned about that, we can do it

cgillum: We should be on the 8.0.x versions now since Hal recently migrated this repo to using .NET 8 instead of .NET 6.

andystaples: Done

andystaples: Done

cgillum (approved):

PR #456, opened 2025-08-05
Add RestartAsync API Support at DurableTaskClient

nytian:
This PR is part of support of issue https://github.com/Azure/azure-functions-durable-extension/issues/2903

This PR adds a new RestartAsync api at DurableTaskClient that allows restarting orchestration instances with provided instanceId or a new instanceId.

Related PR: 

- https://github.com/microsoft/durabletask-protobuf/pull/46
- https://github.com/Azure/azure-functions-durable-extension/pull/3147

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -399,6 +399,46 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
         throw new NotSupportedException($"{this.GetType()} does not support purging of orchestration instances.");
     }
 
+    /// <summary>
+    /// Restarts an orchestration instance with the same or a new instance ID.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This method restarts an existing orchestration instance. If <paramref name="restartWithNewInstanceId"/> is <c>true</c>,
+    /// a new instance ID will be generated for the restarted orchestration. If <c>false</c>, the original instance ID will be reused.
+    /// </para><para>
+    /// The restarted orchestration will use the same input data as the original instance. If the original orchestration
+    /// instance is not found, an <see cref="ArgumentException"/> will be thrown.
+    /// </para><para>
+    /// Note that this operation is backend-specific and may not be supported by all durable task backends.
+    /// If the backend does not support restart operations, a <see cref="NotSupportedException"/> will be thrown.
+    /// </para>
+    /// </remarks>
+    /// <param name="instanceId">The ID of the orchestration instance to restart.</param>
+    /// <param name="restartWithNewInstanceId">
+    /// If <c>true</c>, a new instance ID will be generated for the restarted orchestration.
+    /// If <c>false</c>, the original instance ID will be reused.
+    /// </param>
+    /// <param name="cancellation">
+    /// The cancellation token. This only cancels enqueueing the restart request to the backend.
+    /// Does not abort restarting the orchestration once enqueued.
+    /// </param>
+    /// <returns>
+    /// A task that completes when the orchestration instance is successfully restarted.
+    /// The value of this task is the instance ID of the restarted orchestration instance.
+    /// </returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found.
+    /// </exception>
+    /// <exception cref="NotSupportedException">
+    /// Thrown if the backend does not support restart operations.
+    /// </exception>
+    public virtual Task<string> RestartAsync(
+        string instanceId,
+        bool restartWithNewInstanceId = false,
+        CancellationToken cancellation = default)
+        => throw new NotSupportedException($"{this.GetType()} does not support orchestration restart.");
+
     // TODO: Create task hub
 
     // TODO: Delete task hub
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -428,14 +428,13 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// The value of this task is the instance ID of the restarted orchestration instance.
     /// </returns>
     /// <exception cref="ArgumentException">
-    /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found.
-    /// </exception>
+    /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found. </exception>
     /// <exception cref="InvalidOperationException">
     /// Thrown when attempting to restart an instance using the same instance Id
-    /// while the instance has not yet reached a completed or terminal state.
+    /// while the instance has not yet reached a completed or terminal state. </exception>
     /// <exception cref="NotSupportedException">
-    /// Thrown if the backend does not support restart operations.
-    /// </exception>
+    /// Thrown if the backend does not support restart operations. </exception>
+    [Obsolete("Experimental")]
     public virtual Task<string> RestartAsync(
         string instanceId,
         bool restartWithNewInstanceId = false,
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -430,6 +430,9 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <exception cref="ArgumentException">
     /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found.
     /// </exception>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when attempting to restart an instance using the same instance Id
+    /// while the instance has not yet reached a completed or terminal state.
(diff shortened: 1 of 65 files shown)


src/Client/Core/DurableTaskClient.cs
```diff
@@ -399,6 +399,46 @@ public virtual Task<PurgeResult> PurgeAllInstancesAsync(
...
+    /// <exception cref="ArgumentException">
+    /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found.
+    /// </exception>
+    /// <exception cref="NotSupportedException">
+    /// Thrown if the backend does not support restart operations.
+    /// </exception>
+    public virtual Task<string> RestartAsync(
+        string instanceId,
+        bool restartWithNewInstanceId = false,
+        CancellationToken cancellation = default)
```
nytian, line 440: I am not sure if a cancellation token is needed. We don't have this at in-proc restart api. But it seems all other api here has this so I added it.

src/Grpc/orchestrator_service.proto
```diff
@@ -471,6 +471,15 @@ message PurgeInstancesResponse {
     google.protobuf.BoolValue isComplete = 2;
 }
 
+message RestartInstanceRequest {
```
nytian, line 485: This should be updated once we merge PR https://github.com/microsoft/durabletask-protobuf/pull/46. But we will merge these related PR together, so I just manually updated it

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -254,6 +254,47 @@ public override async Task<OrchestrationMetadata> WaitForInstanceStartAsync(
...
+    public override async Task<string> RestartAsync(
+        string instanceId,
+        bool restartWithNewInstanceId = false,
+        CancellationToken cancellation = default)
+    {
+        Check.NotNullOrEmpty(instanceId);
+        cancellation.ThrowIfCancellationRequested();
+
+        // Get the current orchestration status to retrieve the name and input
+        OrchestrationMetadata? status = await this.GetInstanceAsync(instanceId, getInputsAndOutputs: true, cancellation);
```
YunchuWang, line 267: any race condition if get instance returns instance but it got deleted before restarting?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -408,6 +408,38 @@ public override Task<PurgeResult> PurgeAllInstancesAsync(
...
+        string instanceId,
+        bool restartWithNewInstanceId = false,
+        CancellationToken cancellation = default)
+    {
+        Check.NotNullOrEmpty(instanceId);
+        Check.NotEntity(this.options.EnableEntitySupport, instanceId);
+
+        var request = new P.RestartInstanceRequest
+        {
+            InstanceId = instanceId,
```
YunchuWang, line 409: why not keeping one field only instanceid, if it is not set, use a new id; else restart with it?

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -254,6 +254,47 @@ public override async Task<OrchestrationMetadata> WaitForInstanceStartAsync(
...
+        };
+
+        // Use the original serialized input directly to avoid double serialization
+        TaskMessage message = new()
+        {
+            OrchestrationInstance = instance,
+            Event = new ExecutionStartedEvent(-1, status.SerializedInput)
+            {
+                Name = status.Name,
+                OrchestrationInstance = instance,
```
nytian, line 302: note here we don't support Version. Because OrchestrationMetadata doesn't have version property.

src/Client/Grpc/GrpcDurableTaskClient.cs
nytian: This is to keep parity with the current implementation at in-proc model.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
nytian: I don't think this will be in conflict. We only need the state as we need to check its status, and get it's input and instanceid. I think it doesn't matter if it's deleted or not later.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
bachuv: > note here we don't support Version. Because OrchestrationMetadata doesn't have version property.

Should we add this as a comment in the code?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -395,6 +395,38 @@ public override Task<PurgeResult> PurgeAllInstancesAsync(
         return this.PurgeInstancesCoreAsync(request, cancellation);
     }
 
+    /// <inheritdoc/>
+    public override async Task<string> RestartAsync(
```
torosent, line 399: Please add `[Obsolete("Experimental")]`

nytian: Do you want to add this too at ShimDurableTaskClient? Also can I know why? If we mark it as this it this will looks to customer as they are not encouraged to use this api?

src/Client/Core/DurableTaskClient.cs
```diff
@@ -399,6 +399,49 @@ public virtual Task<PurgeResult> PurgeAllInstancesAsync(
...
+    /// A task that completes when the orchestration instance is successfully restarted.
+    /// The value of this task is the instance ID of the restarted orchestration instance.
+    /// </returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown if an orchestration with the specified <paramref name="instanceId"/> was not found.
+    /// </exception>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when attempting to restart an instance using the same instance Id
+    /// while the instance has not yet reached a completed or terminal state.
+    /// <exception cref="NotSupportedException">
```
cgillum, line 435: Missing `</exception>` tag here, which is being called out by the warnings surfaced here in GitHub.

src/Client/Core/DurableTaskClient.cs
cgillum: We should have a `CancellationToken` parameter for all async client APIs as a best practice, so this is good.

src/Client/Grpc/GrpcDurableTaskClient.cs
cgillum: This is only needed in the interface definition, not in the implementation.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
cgillum: We should open a bug to track fixing this.

test/Client/OrchestrationServiceClientShim.Tests/ShimDurableTaskClientTests.cs
```diff
@@ -326,6 +326,78 @@ public async Task ScheduleNewOrchestrationInstance_IdProvided_TagsProvided()
...
+        // Verify the captured message details
+        capturedMessage.Should().NotBeNull();
+        capturedMessage!.Event.Should().BeOfType<ExecutionStartedEvent>();
+        
+        var startedEvent = (ExecutionStartedEvent)capturedMessage.Event;
+        startedEvent.Name.Should().Be(orchestratorName);
+        startedEvent.Input.Should().Be(serializedInput);
+        startedEvent.OrchestrationInstance.InstanceId.Should().Be(restartedInstanceId);
+        startedEvent.OrchestrationInstance.ExecutionId.Should().NotBeNullOrEmpty();
+        startedEvent.OrchestrationInstance.ExecutionId.Should().NotBe(originalState.OrchestrationInstance.ExecutionId);
```
cgillum, line 383: Let's add a TODO mentioning that we need to confirm that the version is correct.

src/Client/Grpc/GrpcDurableTaskClient.cs
nytian: I added this in the interface as well as the two implementations,

cgillum: Sorry, why did we decide that the `[Obsolete]` attribute is needed? Isn't Restart a GA feature?

nytian: oh I thought you voted for it too. My bad. I will remove it. :) I’m not entirely sure why it was needed, but if @torosent prefers it, I can always add it back.

torosent: The idea was that it's experimental but I'm ok with not adding it

cgillum: Maybe you're thinking of the Rewind API? IIRC, the Restart API isn't experimental.

cgillum (approved):

torosent (approved):
LGTM

PR #462, opened 2025-08-26
Fix encoding of entity unlock events

sebastianburckhardt:
This has been just wrong from the beginning. 

We eventually noticed this because it causes nondeterminism failures on orchestrators. During replay the events are compared and they don't match because of the incorrect encoding.


src/Shared/Grpc/EntityConversions.cs
```diff
@@ -138,7 +138,7 @@ public static DTCore.History.HistoryEvent EncodeUnlockSent(
        OrchestrationInstance? instance)
     {
         P.EntityUnlockSentEvent unlockSentEvent = protoEvent.EntityUnlockSent;
-        string name = EncodeEventName(null);
+        string name = "release";
```
halspang, line 141: I trust that this is correct, but can we get context on it? It seems the other encodes use the same `EncodeEventName`, why is this one different?

If this has always been wrong, can we also get a test for this path that shows it is working? Or is there already a test that has traditionally been flaky and we haven't investigated it?

sophiatev: I think maybe this [PR](https://github.com/microsoft/durabletask-dotnet/pull/464) adds the tests

sebastianburckhardt: > I think maybe this [PR](https://github.com/microsoft/durabletask-dotnet/pull/464) adds the tests

Yes, specifically the test in `TwoCriticalSections.cs`

sebastianburckhardt: > I trust that this is correct, but can we get context on it? 

The event name must be encoded the same way as it is decoded on line https://github.com/microsoft/durabletask-dotnet/blob/65bbff1126d3818488f221cfef3835814a3bf618/src/Shared/Grpc/EntityConversions.cs#L299.

The format we use here to encode entity messages as events is the same as in DurableTask.Core (for example, event names are defined in 
https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Entities/EventFormat/EntityMessageEventNames.cs) but it is not technically required to be the same and we do not want to introduce a dependency on the latter, so the class `EntityConversions.cs` is designed to be self-contained, and now, hopefully, self-consistent.

halspang (approved):

PR #464, opened 2025-09-02
Standalone app that uses the portable SDK for testing entities.

sebastianburckhardt:
I have used these tests extensively while developing entity support in DTS. However, they have not been integrated into any unit test environment yet. At the moment, to run these tests, just set DTS_CONNECTION_STRING and start the app. Then you can list and execute the tests using http commands (see file demo.http for examples).


sebastianburckhardt:
build was still failing. I think I got it this time though, waiting for CI. Fingers crossed.

update: build has succeeded.

PR #467, opened 2025-09-16
Abandon workitem if processing workitem failed 

YunchuWang:
This pull request improves the error handling and logging for background task processing in the gRPC worker, ensuring that when an error occurs, the system properly abandons the relevant work item and logs the process for orchestrator, activity, and entity work items. It also introduces comprehensive tests to verify the new logging behavior.

**Enhanced error handling and logging:**

* Added logic to `RunBackgroundTask` in `GrpcDurableTaskWorker.Processor` to explicitly abandon orchestrator, activity, and both versions of entity work items when an error occurs, with detailed logging for each step. Throws an exception if the work item type is unknown.
* Introduced new logger messages in `Logs.cs` for "abandoning" and "abandoned" states for orchestrator, activity, and entity work items, each with distinct event IDs and message formats.

**Testing improvements:**

* Added a new test suite `RunBackgroundTaskLoggingTests` that verifies the correct logging and abandonment behavior for orchestrator, activity, and both versions of entity work items when errors are encountered during background task processing.

### DIFF
--- a/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
+++ b/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
@@ -276,7 +276,8 @@ sealed partial class GrpcDurableTaskWorker
                             () => this.OnRunOrchestratorAsync(
                                 workItem.OrchestratorRequest,
                                 workItem.CompletionToken,
-                                cancellation));
+                                cancellation),
+                            cancellation);
                     }
                     else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.ActivityRequest)
                     {
@@ -285,13 +286,15 @@ sealed partial class GrpcDurableTaskWorker
                             () => this.OnRunActivityAsync(
                                 workItem.ActivityRequest,
                                 workItem.CompletionToken,
-                                cancellation));
+                                cancellation),
+                            cancellation);
                     }
                     else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.EntityRequest)
                     {
                         this.RunBackgroundTask(
                             workItem,
-                            () => this.OnRunEntityBatchAsync(workItem.EntityRequest.ToEntityBatchRequest(), cancellation));
+                            () => this.OnRunEntityBatchAsync(workItem.EntityRequest.ToEntityBatchRequest(), cancellation),
+                            cancellation);
                     }
                     else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.EntityRequestV2)
                     {
@@ -305,7 +308,8 @@ sealed partial class GrpcDurableTaskWorker
                                 batchRequest,
                                 cancellation,
                                 workItem.CompletionToken,
-                                operationInfos));
+                                operationInfos),
+                             cancellation);
                     }
                     else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.HealthPing)
                     {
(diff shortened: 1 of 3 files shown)


src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -351,8 +351,75 @@ void RunBackgroundTask(P.WorkItem? workItem, Func<Task> handler)
...
+
+                    if (workItem?.OrchestratorRequest != null)
+                    {
+                        this.Logger.AbandoningOrchestratorWorkItem(instanceId, workItem?.CompletionToken ?? string.Empty);
+                        await this.client.AbandonTaskOrchestratorWorkItemAsync(
+                            new P.AbandonOrchestrationTaskRequest
+                            {
+                                CompletionToken = workItem?.CompletionToken,
+                            },
+                            cancellationToken: CancellationToken.None);
```
sebastianburckhardt: I think we should use the cancellation token of the worker here, just add it to the argument list of `RunBackgroundTask`

(and same for the other ones below)

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -351,8 +351,75 @@ void RunBackgroundTask(P.WorkItem? workItem, Func<Task> handler)
...
+                        this.Logger.AbandonedOrchestratorWorkItem(instanceId, workItem?.CompletionToken ?? string.Empty);
+                    }
+                    else if (workItem?.ActivityRequest != null)
+                    {
+                        this.Logger.AbandoningActivityWorkItem(
+                            instanceId,
+                            workItem.ActivityRequest.Name,
+                            workItem.ActivityRequest.TaskId,
+                            workItem?.CompletionToken ?? string.Empty);
+                        await this.client.AbandonTaskActivityWorkItemAsync(
```
sebastianburckhardt: The abandon calls could themselves throw exeptions, which have to be caught because this is a background task. Also, we need to retry the 'Unavailable' grpc exceptions.

sebastianburckhardt (changes requested):

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -351,8 +351,75 @@ void RunBackgroundTask(P.WorkItem? workItem, Func<Task> handler)
...
+                                CompletionToken = workItem?.CompletionToken,
+                            },
+                            cancellationToken: CancellationToken.None);
+                        this.Logger.AbandonedEntityWorkItem(
+                            workItem.EntityRequestV2.InstanceId,
+                            workItem?.CompletionToken ?? string.Empty);
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("Unknown work item type");
```
sebastianburckhardt: not a good idea to throw an exception here because it is a background task. I think you can just ignore this case since we already logged the exception.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
YunchuWang: updated

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
YunchuWang: updated

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
YunchuWang: updated, yes we have retry added to all grpc calls from this grpclient on unavaiable status code in the default gRPC retry policy but it is only applied when using the Azure Managed Durable Task Scheduler options src\Client\AzureManaged\DurableTaskSchedulerClientOptions.cs
src\Worker\AzureManaged\DurableTaskSchedulerWorkerOptions.cs

YunchuWang: will cut another pr to have grpc retry config for all portable grpc user cases.

PR #468, opened 2025-09-17
Large payload azure blob externalization support 

YunchuWang:
[Copilot is generating a summary...]


src/Extensions/AzureBlobPayloads/Interceptors/PayloadInterceptor.cs
```diff
@@ -0,0 +1,225 @@
...
+        if (string.IsNullOrEmpty(value) || !this.payloadStore.IsKnownPayloadToken(value))
+        {
+            return;
+        }
+
+        string resolved = await this.payloadStore.DownloadAsync(value, cancellation);
+        assign(resolved);
+    }
+
+    sealed class TransformingStreamReader<T> : IAsyncStreamReader<T>
```
halspang: Naive comment: What does this do?

src/Extensions/AzureBlobPayloads/Interceptors/BasePayloadInterceptor.cs
```diff
@@ -0,0 +1,225 @@
...
+    protected abstract Task ResolveResponsePayloadsAsync<TResponse>(TResponse response, CancellationToken cancellation);
+
+    /// <summary>
+    /// Externalizes a payload if it exceeds the threshold.
+    /// </summary>
+    /// <param name="assign">Action to assign the externalized token.</param>
+    /// <param name="value">The value to potentially externalize.</param>
+    /// <param name="cancellation">Cancellation token.</param>
+    /// <returns>A task representing the async operation.</returns>
+    protected Task MaybeExternalizeAsync(Action<string?> assign, string? value, CancellationToken cancellation)
```
halspang: This is a kind of weird interface. We take a function to assign the value back to something? It's fairly opaque what this is doing and I wonder if it's better to return the string in the `Task` and let the caller assign it. The end goal is the same but, to me, the control feels a bit inverted this way.

src/Extensions/AzureBlobPayloads/Options/LargePayloadStorageOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    public LargePayloadStorageOptions()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="LargePayloadStorageOptions"/> class.
+    /// </summary>
+    /// <param name="connectionString">The Azure Storage connection string to the customer's storage account.</param>
+    public LargePayloadStorageOptions(string connectionString)
```
halspang, line 43: Should users be able to connect to storage via an identity?

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -0,0 +1,215 @@
...
+    readonly BlobContainerClient containerClient;
+    readonly LargePayloadStorageOptions options;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BlobPayloadStore"/> class.
+    /// </summary>
+    /// <param name="options">The options for the blob payload store.</param>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="options"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown when <paramref name="options.ConnectionString"/> is null or empty.</exception>
+    public BlobPayloadStore(LargePayloadStorageOptions options)
```
halspang, line 33: Same comment here, I think we should expose identity connections unless you can specify that in the connection string.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -0,0 +1,215 @@
...
+        };
+        BlobServiceClient serviceClient = new(options.ConnectionString, clientOptions);
+        this.containerClient = serviceClient.GetBlobContainerClient(options.ContainerName);
+    }
+
+    /// <inheritdoc/>
+    public async Task<string> UploadAsync(ReadOnlyMemory<byte> payloadBytes, CancellationToken cancellationToken)
+    {
+        // One blob per payload using GUID-based name for uniqueness (stable across retries)
+        string timestamp = DateTimeOffset.UtcNow.ToString("yyyy/MM/dd/HH/mm/ss", CultureInfo.InvariantCulture);
```
halspang: I'm not a huge fan of this naming scheme. We might want to make it configurable so customers can control if it overlaps with their own naming schemes. Regardless, I think we may want to put it under a DTS indicated prefix like `dts/payloads` or something similar. Then I'd just a GUID or hash for the name. Since this should be stored in the input/output of the activities, it doesn't need to be super human readable.

However, this is coming from my own background so if people don't agree we can keep it this way.

halspang: Also, if we rerun this step a lot, will it upload a bunch of new files? It'd be good if it was idempotent to the same key.

src/Extensions/AzureBlobPayloads/Interceptors/PayloadInterceptor.cs
```diff
@@ -0,0 +1,225 @@
...
+    /// <summary>
+    /// Resolves a payload token if it's known to the store.
+    /// </summary>
+    /// <param name="assign">Action to assign the resolved value.</param>
+    /// <param name="value">The value to potentially resolve.</param>
+    /// <param name="cancellation">Cancellation token.</param>
+    /// <returns>A task representing the async operation.</returns>
+    protected async Task MaybeResolveAsync(Action<string?> assign, string? value, CancellationToken cancellation)
+    {
+        if (string.IsNullOrEmpty(value) || !this.payloadStore.IsKnownPayloadToken(value))
```
halspang: Is `IsKnownPayloadToken` essentially being used as a versioning mechanism? It seems like it may cause problems if we try and extend it in the future. Or is the hope to use it to distinguish between different storage backends?

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -0,0 +1,215 @@
...
+        int sep = rest.IndexOf(':');
+        if (sep <= 0 || sep >= rest.Length - 1)
+        {
+            throw new ArgumentException("Invalid external payload token format.", nameof(token));
+        }
+
+        return (rest.Substring(0, sep), rest.Substring(sep + 1));
+    }
+
+    static async Task<T> WithTransientRetryAsync<T>(Func<CancellationToken, Task<T>> operation, CancellationToken cancellationToken)
```
halspang: Does the base client not support retries? I'd love to not reinvent this here.

src/Extensions/AzureBlobPayloads/PayloadStore/IPayloadStore.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Abstraction for storing and retrieving large payloads out-of-band.
+/// </summary>
+public interface IPayloadStore
```
halspang: Super nit: I would classify this more as an abstract class than an interface. It's not something you would think would neat multiple inheritance and might have some base behavior in the future (like with the token stuff). 

In general, you can think of an interface as a label which you can apply to classes. `IDisposable`, it can be disposed. `IComparable` it can be compared. In this case you wouldn't really consider things as 'It can be a PayloadStore' and also other things. 

Anyway, that's a personal preference on how things look, feel free to ignore me :)

src/Grpc.AzureManagedBackend/backend_service.proto
```diff
@@ -0,0 +1,282 @@
...
+import "orchestrator_service.proto";
+
+
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+import "google/protobuf/empty.proto";
+
+// gRPC service used by Durable Task Framework (DTFx) backend implementations.
+// The RPCs in this service are used by DTFx backends to manage orchestration state.
+service BackendService {
```
halspang: I'm hesitant to add a new proto just for this feature. Is it possible that the middleware for this should exist elsewhere?

halspang (commented):
Nothing opposing the core functionality, but my main concern is with the fact that this has to be setup on the client and the worker. I'm worried that users could provide different configurations and break their payload flow.

Do you think it'd be possible to instead make the client/worker extensions depend directly on inheriting an interface/abstract class that we could instantiate instead? By having the user instantiate the class itself and then pass that around, we make chances for misconfiguration go down.

YunchuWang:
> Nothing opposing the core functionality, but my main concern is with the fact that this has to be setup on the client and the worker. I'm worried that users could provide different configurations and break their payload flow.
> 
> Do you think it'd be possible to instead make the client/worker extensions depend directly on inheriting an interface/abstract class that we could instantiate instead? By having the user instantiate the class itself and then pass that around, we make chances for misconfiguration go down.

Hi 

> Nothing opposing the core functionality, but my main concern is with the fact that this has to be setup on the client and the worker. I'm worried that users could provide different configurations and break their payload flow.
> 
> Do you think it'd be possible to instead make the client/worker extensions depend directly on inheriting an interface/abstract class that we could instantiate instead? By having the user instantiate the class itself and then pass that around, we make chances for misconfiguration go down.

good point, i updated to shared option and store

src/Extensions/AzureBlobPayloads/Interceptors/PayloadInterceptor.cs
YunchuWang: This class is just a thin wrapper around gRPC’s IAsyncStreamReader<T>.
Normally, IAsyncStreamReader<T> gives you the raw protobuf messages from the server stream. In our case, those messages may contain “payload tokens” (like blob:v1:...) instead of the actual data.

TransformingStreamReader<T> intercepts each element as it’s read (MoveNext), applies a transformation function (in our case ResolveResponsePayloadsAsync), and then exposes the transformed object as Current.

So effectively:

Without it: the consumer would get unresolved payload tokens.

With it: the consumer always sees hydrated/decoded messages (no tokens).

src/Extensions/AzureBlobPayloads/Interceptors/PayloadInterceptor.cs
YunchuWang: its mainly for distinguish backend, appending version is to support token schema change if any in future. I think it is needed to ensure token schema can be safely change without breaking.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
YunchuWang: it wont generate a new blob name every retry at blob client level, but can be at grpc retry level. if we want to make this fully idempotent, we would need payload content hash. problem with that is it is error prone for customer to delete blobs as they need to know if the blob is referred by any orchestration. I will make this guid based.

src/Extensions/AzureBlobPayloads/PayloadStore/IPayloadStore.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Abstraction for storing and retrieving large payloads out-of-band.
+/// </summary>
+public abstract class IPayloadStore
```
AnatoliB: Since this is not an interface anymore, prefixing it with "I" is probably not a good idea. Consider renaming it to `PayloadStoreBase` or `AbstractPayloadStore` then. (Personally, I wouldn't even mind keeping it an interface to begin with, my only problem is that the current name breaks the convention.)

src/Extensions/AzureBlobPayloads/Examples/SharedPayloadStoreExample.cs
AnatoliB, line 1: Do we need this file to be completely commented out? Also, something is wrong with this file: GitHub show it as binary on the PR

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// <inheritdoc/>
+    public override async Task<string> UploadAsync(ReadOnlyMemory<byte> payloadBytes, CancellationToken cancellationToken)
+    {
+        // One blob per payload using GUID-based name for uniqueness (stable across retries)
+        string blobName = $"{Guid.NewGuid():N}";
+        BlobClient blob = this.containerClient.GetBlobClient(blobName);
+
+        byte[] payloadBuffer = payloadBytes.ToArray();
```
AnatoliB: This creates a temporary in-memory copy of a potentially large chunk of data. Since this data remains read-only, perhaps there is a way to get rid of this copy. `Stream.WriteAsync` accepts `ReadOnlyMemory<byte>` directly, so you may be able to *also* skip wrapping this data into a `MemoryStream` and just do `blobStream.WriteAsync(payloadBytes, ...)` or `compressedBlobStream(payloadBytes, ...)`.

src/Extensions/AzureBlobPayloads/Interceptors/AzureBlobPayloadsManagedBackendInterceptor.cs
```diff
@@ -0,0 +1,495 @@
...
+
+using P = Microsoft.DurableTask.AzureManagedBackend.Protobuf;
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// gRPC interceptor that externalizes large payloads to an <see cref="IPayloadStore"/> on requests
+/// and resolves known payload tokens on responses for Azure Managed Backend.
+/// </summary>
+public sealed class AzureBlobPayloadsManagedBackendInterceptor(IPayloadStore payloadStore, LargePayloadStorageOptions options)
```
AnatoliB: Quite a lot of code duplication between this and `AzureBlobPayloadsSideCarInterceptor` - wondering if there is a way to extract the common code to a shared component.

src/Extensions/AzureBlobPayloads/PayloadStore/PayloadStore.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <returns>Opaque reference token.</returns>
+    public abstract Task<string> UploadAsync(ReadOnlyMemory<byte> payloadBytes, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// Downloads the payload referenced by the token.
+    /// </summary>
+    /// <param name="token">The opaque reference token.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>Payload string.</returns>
+    public abstract Task<string> DownloadAsync(string token, CancellationToken cancellationToken);
```
AnatoliB, line 25: `UploadAsync` receives `ReadOnlyMemory<byte>`, but `DownloadAsync` returns a string. One could naively expect symmetry here. It looks like `UploadAsync` is always invoked with a result of `Encoding.UTF8.GetBytes(value)`. If this is indeed the case, perhaps pass a string to `UploadAsync` and do the encoding inside?

src/Grpc.AzureManagedBackend/backend_service.proto
AnatoliB: I share this concern, but I don't have good understanding of what alternatives are available. Let's list our options. To start:
- Can we add support for this feature to the exising Grpc project?
- Where can we move this middleware (so that it wouldn't require this gRPC update, if this is what Hal means)?

I'm not strictly against doing it the way it's currently implemented, but let's understand and compare our options.

src/Extensions/AzureBlobPayloads/Examples/SharedPayloadStoreExample.cs
YunchuWang: will clean up this file

src/Extensions/AzureBlobPayloads/PayloadStore/IPayloadStore.cs
YunchuWang: good catch, I checked "abstract" prefix for abstract classes is not used in this repo. will update to PayloadStore to be consistent with same convention.

src/Extensions/AzureBlobPayloads/PayloadStore/PayloadStore.cs
YunchuWang: will update

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
YunchuWang: updated

src/Grpc.AzureManagedBackend/backend_service.proto
YunchuWang: - Can we add support for this feature to the exising Grpc project?
Unfortunately not, dotnet cant generate same proto compiled types under multiple namespaces in same project assembly. we need two projects to generate proto types under two namespaces respectively.
- Where can we move this middleware (so that it wouldn't require this gRPC update, if this is what Hal means)?
We can move the azuremanaged interceptor to azuremanaged sdk in DTMB repo . It would create coupling not make sense much to put the blob which is opt in extension logic like intereceptor into existing azuremanaged packages directly. if we move the middleware to there, we would still need a new extension package.

the other alternative is creating multiple packages:
**Microsoft.DurableTask.Extensions.AzureBlobPayloads.Abstractions** (common large payload apis including payloadstore, baseinterceptor,etc shared by other packages)
**Microsoft.DurableTask.Extensions.AzureBlobPayloads.Sidecar** (sidecar interceptorfor portable dotnet sdk connecting to sidecar grpc )
**Microsoft.DurableTask.Extensions.AzureBlobPayloads.AzureManaged** (azuremanaged intereceptor for df cases)

@AnatoliB

src/Extensions/AzureBlobPayloads/Interceptors/AzureBlobPayloadsManagedBackendInterceptor.cs
YunchuWang: will address once we confirm on the package approach

src/Extensions/AzureBlobPayloads/DependencyInjection/DurableTaskClientBuilderExtensions.AzureBlobPayloads.cs
```diff
@@ -0,0 +1,61 @@
...
+/// </summary>
+public static class DurableTaskClientBuilderExtensionsAzureBlobPayloads
+{
+    /// <summary>
+    /// Enables externalized payload storage using a pre-configured shared payload store.
+    /// This overload helps ensure client and worker use the same configuration.
+    /// </summary>
+    /// <param name="builder">The builder to configure.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskClientBuilder UseExternalizedPayloads(
```
halspang, line 25: Should there be a hard check somewhere in this path to check if an `IPayloadStore` has been set in the services? It would be good if we could detect the case that `AddExternalizedPayloadStore` hasn't been called and give an error stating that.

Alternatives to that could be:
- Adding a `IPayloadStore` into the method parameters to force something like `builder.Services.UseExternalizedPayloads(builder.Services.GetRequiredService<IPayloadStore>())`
- Require a specifically configured section in the app's configuration that the extension keys into directly. This might make it a bit more SDK independent since I don't know if all SDKs support extensions, but it could also get messy if we introduce a lot more storage options.

src/Extensions/AzureBlobPayloads/DependencyInjection/DurableTaskWorkerBuilderExtensions.AzureBlobPayloads.cs
```diff
@@ -0,0 +1,84 @@
...
+        return UseExternalizedPayloadsCore(builder);
+    }
+
+    /// <summary>
+    /// Enables externalized payload storage using a pre-configured shared payload store.
+    /// This overload helps ensure client and worker use the same configuration.
+    /// </summary>
+    /// <param name="builder">The builder to configure.</param>
+    /// <returns>The original builder, for call chaining.</returns>
+    public static IDurableTaskWorkerBuilder UseExternalizedPayloads(
```
halspang, line 48: Same comment as the client builder here.

src/Extensions/AzureBlobPayloads/DependencyInjection/ServiceCollectionExtensions.AzureBlobPayloads.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        Check.NotNull(services);
+        Check.NotNull(configure);
+
+        // Apply once to ALL names (IConfigureOptions<T> hits every named options instance),
+        // so monitor.Get(builder.Name) in the client/worker extensions will see the same config.
+        services.Configure(configure);
+
+        // Provide a single shared PayloadStore instance built from the default options.
+        services.AddSingleton<PayloadStore>(sp =>
```
halspang, line 34: Should we not add it as the base interface (`IPayloadStore`)?

halspang: I've just realized that the file name is `IPayloadStore` but it's actually an abstract class, nevermind.

src/Extensions/AzureBlobPayloads/PayloadStore/PayloadStore.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask;
+
+/// <summary>
+/// Abstraction for storing and retrieving large payloads out-of-band.
+/// </summary>
+public abstract class PayloadStore
```
halspang, line 9: Can you make sure the filename and classname match? The filename implies it's an interface. Unless that's just github being weird, if so, disregard.

src/Extensions/AzureBlobPayloads/DependencyInjection/ServiceCollectionExtensions.AzureBlobPayloads.cs
YunchuWang: updated filename

src/Grpc.AzureManagedBackend/backend_service.proto
YunchuWang: discussed offline, will go with two packages

src/Extensions/AzureBlobPayloads/DependencyInjection/DurableTaskClientBuilderExtensions.AzureBlobPayloads.cs
YunchuWang: i tested the dotnet already throw a clear exception if payloadstore is not registered at runtime. I think it is fine to keep as is (dotnet validate it exists right before using it)
image

src/Extensions/AzureBlobPayloads/DependencyInjection/DurableTaskWorkerBuilderExtensions.AzureBlobPayloads.cs
YunchuWang: same as above

samples/LargePayloadConsoleApp/Program.cs
```diff
@@ -0,0 +1,198 @@
...
+Console.WriteLine($"Deserialized input length: {deserializedInput.Length}");
+
+// Run entity samples
+Console.WriteLine();
+Console.WriteLine("Running LargeEntityOperationInput...");
+string largeEntityInput = new string('E', 700 * 1024); // 700KB
+string entityInputInstance = await client.ScheduleNewOrchestrationInstanceAsync("LargeEntityOperationInput");
+OrchestrationMetadata entityInputResult = await client.WaitForInstanceCompletionAsync(entityInputInstance, getInputsAndOutputs: true, cts.Token);
+int entityInputLength = entityInputResult.ReadOutputAs<int>();
+Console.WriteLine($"Status: {entityInputResult.RuntimeStatus}, Output length: {entityInputLength}");
```
halspang: Nit: Is this talking about input or output?

samples/LargePayloadConsoleApp/Program.cs
```diff
@@ -0,0 +1,198 @@
...
+string entityStateInstance = await client.ScheduleNewOrchestrationInstanceAsync("LargeEntityState");
+OrchestrationMetadata entityStateOrch = await client.WaitForInstanceCompletionAsync(entityStateInstance, getInputsAndOutputs: true, cts.Token);
+Console.WriteLine($"Status: {entityStateOrch.RuntimeStatus}");
+EntityMetadata<string>? state = await client.Entities.GetEntityAsync<string>(new EntityInstanceId(nameof(StateEntity), "1"), includeState: true);
+int stateLength = state?.State?.Length ?? 0;
+Console.WriteLine($"State length: {stateLength}");
+Console.WriteLine($"Deserialized state equals original: {state?.State == largeEntityState}");
+
+
+
```
halspang: Nit, this is a lot of extra space.

src/Extensions/AzureBlobPayloads/Options/LargePayloadStorageOptions.cs
```diff
@@ -0,0 +1,93 @@
...
+        Check.NotNull(accountUri, nameof(accountUri));
+        Check.NotNull(credential, nameof(credential));
+        this.AccountUri = accountUri;
+        this.Credential = credential;
+    }
+
+    /// <summary>
+    /// Gets or sets the threshold in bytes at which payloads are externalized. Default is 900_000 bytes.
+    /// </summary>
+    public int ExternalizeThresholdBytes { get; set; } = 900_000; // leave headroom below 1MB
```
halspang: Should we enforce a max this can be? Cause a customer could set it >1MB and have their requests fail, right?

test/Grpc.IntegrationTests/LargePayloadTests.cs
```diff
@@ -0,0 +1,647 @@
...
+            if (DateTime.UtcNow >= deadline)
+            {
+                Assert.NotNull(status);
+                Assert.Equal(OrchestrationRuntimeStatus.Running, status!.RuntimeStatus);
+            }
+
+            await Task.Delay(TimeSpan.FromSeconds(1), this.TimeoutToken);
+        }
+
+
```
halspang, line 227: Nit: Extra space.

test/Grpc.IntegrationTests/LargePayloadTests.cs
```diff
@@ -0,0 +1,647 @@
...
+                // Override store with in-memory test double
+                client.Services.AddSingleton<PayloadStore>(fakeStore);
+            },
+            services =>
+            {
+                services.AddExternalizedPayloadStore(opts =>
+                {
+                    opts.ExternalizeThresholdBytes = 1024;
+                    opts.ContainerName = "test";
+                    opts.ConnectionString = "UseDevelopmentStorage=true";
```
halspang, line 53: Nit, you use a direct `PayloadStore` registration and the new extension method. Can you just use the extension and make sure it still works?

src/Extensions/AzureBlobPayloads/Options/LargePayloadStorageOptions.cs
YunchuWang: updated the comment on 1mb, actually 1mb was the limit if not using large payload feature. but good point, we still should have a cap on max limit with large payload, will update

test/Grpc.IntegrationTests/LargePayloadTests.cs
YunchuWang: unfortunately extension is for production payloadstore backed by blob(we dont support e2e test connecting to real blob in github), direct registration of mocked store is easiest way to inject without modifying extension interfaces. but good point, we can add emulator based tests later,

src/Extensions/AzureBlobPayloads/Options/LargePayloadStorageOptions.cs
```diff
@@ -60,7 +60,14 @@ public LargePayloadStorageOptions(Uri accountUri, TokenCredential credential)
...
     /// </summary>
-    public int ExternalizeThresholdBytes { get; set; } = 900_000; // leave headroom below 1MB
+    public int ExternalizeThresholdBytes { get; set; } = 900_000;
+
+    /// <summary>
+    /// Gets or sets the maximum allowed size in bytes for any single externalized payload.
+    /// Defaults to 10MB. Requests exceeding this limit will fail fast
+    /// with a clear error to prevent unbounded payload growth and excessive storage/network usage.
+    /// </summary>
+    public int MaxExternalizedPayloadBytes { get; set; } = 10 * 1024 * 1024;
```
halspang, line 89: Why is this configurable? My comment earlier was that we know a certain payload size is supported without externalization, I think we should make sure that `ExternalizeThresholdBytes` is <= to that. If we let customers set `ExternalizeThresholdBytes` higher than that number, the requests will fail because the input/output is too large.

YunchuWang: discussed offline. we will keep both.

src/Extensions/AzureBlobPayloads/AzureBlobPayloads.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <PackageReference Include="Grpc.Net.Client" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Abstractions\Abstractions.csproj" />
+    <ProjectReference Include="..\..\Client\Core\Client.csproj" />
+    <ProjectReference Include="..\..\Worker\Core\Worker.csproj" />
+    <ProjectReference Include="..\..\Client\Grpc\Client.Grpc.csproj" />
+    <ProjectReference Include="..\..\Worker\Grpc\Worker.Grpc.csproj" />
+    <ProjectReference Include="..\..\Grpc.AzureManagedBackend\Grpc.AzureManagedBackend.csproj" />
```
nytian, line 24: Where is Grpc.AzureManagedBackend.csproj?

PR #471, opened 2025-09-25
Added in-process Test Framework

nytian:


test/InProcessTestFramework/MockDurableTaskClient.cs
```diff
@@ -0,0 +1,122 @@
...
+        object? eventPayload = null, 
+        CancellationToken cancellation = default)
+    {
+        return this.orchestrationService.RaiseEventAsync(instanceId, eventName, eventPayload, cancellation);
+    }
+
+    /// <inheritdoc/>
+    public override Task SuspendInstanceAsync(string instanceId, string? reason = null, CancellationToken cancellation = default)
+    {
+        return this.orchestrationService.SuspendInstanceAsync(instanceId, reason, cancellation);
```
nytian, line 87: I just realized that I shouldn't write suspend/resume like this  as we won't be able to pause the running task. I would just throw not supported exception here for now

nytian:
close as no longer needed

PR #472, opened 2025-09-30
Add version check to activities

halspang:
Activities were previously not checking the versioning status of the worker. This was because it was assumed that the orchestration level check would satisfy the activity. However, when multiple workers of the same version are present, the activities can be sent to workers of the incorrect version. This change adds the orchestration version to the activity to allow for versioning checks.

### DIFF
--- a/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
@@ -154,7 +154,7 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
             {
                 return await this.innerContext.ScheduleTask<T>(
                     name.Name,
-                    name.Version,
+                    this.innerContext.Version,
                     options: ScheduleTaskOptions.CreateBuilder()
                         .WithRetryOptions(policy.ToDurableTaskCoreRetryOptions())
                         .WithTags(tags)
@@ -166,7 +166,7 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
                 return await this.InvokeWithCustomRetryHandler(
                     () => this.innerContext.ScheduleTask<T>(
                         name.Name,
-                        name.Version,
+                        this.innerContext.Version,
                         options: ScheduleTaskOptions.CreateBuilder()
                             .WithTags(tags)
                             .Build(),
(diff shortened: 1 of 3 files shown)


AnatoliB:
This fix will make changes safer in general, but there is a trade-off here: it may hurt performance in scenarios when the user is _not_ making changes to activities (or simply doesn't mind the old activity implementation to be active even for new orchestrations for some time). Currently, old workers can still participate in handling the workload while they are still alive. After the fix, they will be immediately excluded. I still believe the fix is a better default behavior, but let's consider either making it configurable, or (long-term) have separate versioning for activities.

I'm not blocking on this though: if making it configurable is not trivial, I'm still ok with this fix.

halspang:
> This fix will make changes safer in general, but there is a trade-off here: it may hurt performance in scenarios when the user is _not_ making changes to activities (or simply doesn't mind the old activity implementation to be active even for new orchestrations for some time). Currently, old workers can still participate in handling the workload while they are still alive. After the fix, they will be immediately excluded. I still believe the fix is a better default behavior, but let's consider either making it configurable, or (long-term) have separate versioning for activities.
> 
> I'm not blocking on this though: if making it configurable is not trivial, I'm still ok with this fix.

Agreed! Long-term I want to be able to specify activities versions. This change is to target the immediate user pain and then we can work towards that in the future.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -154,7 +154,7 @@ public override async Task<T> CallActivityAsync<T>(
             {
                 return await this.innerContext.ScheduleTask<T>(
                     name.Name,
-                    name.Version,
+                    this.innerContext.Version,
```
cgillum, line 157: Is this a breaking change? Previously the version passed to the activity came from `TaskName.Version`, but now it comes from the orchestration's version.  I highly doubt anyone was using `TaskName.Version` at this point, but it might be weird if this value is suddenly ignored and never used for anything. Would it make sense to make this a bit more flexible, and do something like this?

```suggestion
                    string.IsNullOrEmpty(name.Version) ? this.innerContext.Version : name.Version,
```

But I guess one could argue that this makes the semantics even more confusing because sometimes we use activity versioning and sometimes we use orchestration versioning. Not sure what the right path forward is. More discussion might be required.

halspang: If you look at [TaskName](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskName.cs), you'll see that we never actually gave the ability to _use_ the version. So, this change just removes the deprecated field and uses the Orchestration's version.

I do definitely want to make it configurable in the future, but right now this solves a direct user ask and then gives us time to implement activity versioning how we want to.

PR #474, opened 2025-10-02
Include Exception Properties at FailureDetails

nytian:
This PR supports register custom provider that help include exception properties at `TaskFailureDetails`.  Supports include running with both DurableTaskWorker and Durable Functions (.NET Isolated).

Note: related PR : 

- durabletask.core : https://github.com/Azure/durabletask/pull/1237
- durabletask.protobuf: https://github.com/microsoft/durabletask-protobuf/pull/45

### DIFF
--- a/src/Grpc/versions.txt
+++ b/src/Grpc/versions.txt
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch nytian/failure-details at 2025-10-01 21:51:24 UTC
-https://raw.githubusercontent.com/microsoft/durabletask-protobuf/362f886f1ef7c4dd90cbdfdb2f661f48eeeec4fa/protos/orchestrator_service.proto
+# The following files were downloaded from branch main at 2025-10-13 18:06:55 UTC
+https://raw.githubusercontent.com/microsoft/durabletask-protobuf/97cf9cf6ac44107b883b0f4ab1dd62ee2332cfd9/protos/orchestrator_service.proto
--- a/src/Grpc/orchestrator_service.proto
+++ b/src/Grpc/orchestrator_service.proto
@@ -691,7 +691,6 @@ message SkipGracefulOrchestrationTerminationsResponse {
     // Those instances which could not be terminated because they had locked entities at the time of this termination call,
     // are already in a terminal state (completed, failed, terminated, etc.), are not orchestrations, or do not exist (i.e. have been purged)
     repeated string unterminatedInstanceIds = 1;
-
 }
 
 service TaskHubSidecarService {


test/Grpc.IntegrationTests/GrpcSidecar/Grpc/TaskHubGrpcServer.cs
```diff
@@ -398,22 +398,6 @@ static P.GetInstanceResponse CreateGetInstanceResponse(OrchestrationState state,
         }
     }
 
-    static P.TaskFailureDetails? GetFailureDetails(FailureDetails? failureDetails)
```
nytian, line 401: remove this because we will call the one at ptobufutils instead

src/Worker/Core/IExceptionPropertiesProvider.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
```
nytian, line 1: Question: is this the right place to put this file? Customer need to refer this interface so they need to use this package. I guess I should put it here but I am not 100% sure.

src/Abstractions/TaskFailureDetails.cs
```diff
@@ -160,14 +163,16 @@ internal CoreFailureDetails ToCoreFailureDetails()
...
+                coreEx.FailureDetails?.Properties ?? null);
         }
 
         return new TaskFailureDetails(
             exception.GetType().ToString(),
             exception.Message,
             exception.StackTrace,
-            FromExceptionRecursive(exception.InnerException));
+            FromExceptionRecursive(exception.InnerException),
+            null);// might need to udpate this later
```
nytian: Question for the  `public static TaskFailureDetails FromException(Exception exception)`

I checked all the reference, and it seems all the cases we called this  is when the exception is a DTCore.OrchestrationExcep, which means it should already have properties included. Thus currently we don't need to extract properties here.

But I am not sure if there might be a case when the exception is not from dt.core and this _return new FailureDetails_ block  is reached. If so, then we need to extract properties in this repo,  which means I need to update `OrchestrationContextWrapper` class etc. But I didn't find a case like this from current code path. Let me know if there is any other thought on this.

src/Worker/Grpc/GrpcOrchestrationRunner.cs
```diff
@@ -157,88 +157,97 @@ public static string LoadAndRun(
...
+                {
+                    runtimeState.AddEvent(newEvent);
+                }
+
+                TaskName orchestratorName = new(runtimeState.Name);
+                ParentOrchestrationInstance? parent = runtimeState.ParentInstance is ParentInstance p
+                    ? new(new(p.Name), p.OrchestrationInstance.InstanceId)
+                    : null;
+
+                DurableTaskShimFactory factory = services is null
```
nytian, line 211: I only updated line 211 to 226.. the whole file changed because of the VS outlining thing

src/Abstractions/TaskFailureDetails.cs
```diff
@@ -15,7 +15,8 @@ namespace Microsoft.DurableTask;
 /// <param name="ErrorMessage">A summary description of the failure.</param>
 /// <param name="StackTrace">The stack trace of the failure.</param>
 /// <param name="InnerFailure">The inner cause of the task failure.</param>
-public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? StackTrace, TaskFailureDetails? InnerFailure)
+/// <param name="Properties">Additional properties associated with the exception.</param>
+public record TaskFailureDetails(string ErrorType, string ErrorMessage, string? StackTrace, TaskFailureDetails? InnerFailure, IDictionary<string, object?>? Properties)
```
cgillum, line 19: Even better if this could be `IReadOnlyDictionary`, which is immutable, but if it's too much trouble, then this is fine.

src/Grpc/versions.txt
```diff
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch main at 2025-09-17 01:45:58 UTC
-https://raw.githubusercontent.com/microsoft/durabletask-protobuf/f5745e0d83f608d77871c1894d9260ceaae08967/protos/orchestrator_service.proto
+# The following files were downloaded from branch nytian/failure-details at 2025-10-01 21:51:24 UTC
```
cgillum: Reminder to update this to `main` once the protobuf changes in durabletask-protobuf are merged.

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -1,1172 +1,1256 @@
...
+    /// <returns>The task failure details.</returns>
+    [return: NotNullIfNotNull(nameof(e))]
+    internal static P.TaskFailureDetails? ToTaskFailureDetails(this Exception? e, global::DurableTask.Core.IExceptionPropertiesProvider? exceptionPropertiesProvider = null)
+    {
+        if (e == null)
+        {
+            return null;
+        }
+
+        var properties = exceptionPropertiesProvider?.GetExceptionProperties(e);
```
cgillum: nit: Please avoid using `var` for storing the result of a method since we can't tell from reading what type of variable `var` represents.

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -1,1172 +1,1256 @@
...
+    }
+
+    /// <summary>
+    /// Converts a <see cref="Exception" /> to <see cref="P.TaskFailureDetails" />.
+    /// </summary>
+    /// <param name="e">The exception to convert.</param>
+    /// <param name="exceptionPropertiesProvider">Optional exception properties provider.</param>
+    /// <returns>The task failure details.</returns>
+    [return: NotNullIfNotNull(nameof(e))]
+    internal static P.TaskFailureDetails? ToTaskFailureDetails(this Exception? e, global::DurableTask.Core.IExceptionPropertiesProvider? exceptionPropertiesProvider = null)
```
cgillum:
```suggestion
    internal static P.TaskFailureDetails? ToTaskFailureDetails(this Exception? e, DTCore.IExceptionPropertiesProvider? exceptionPropertiesProvider = null)
```

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -1,1172 +1,1256 @@
...
+        {
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.NullValue:
+                return null;
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.NumberValue:
+                return value.NumberValue;
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.StringValue:
+                string stringValue = value.StringValue;
+
+                // Try to parse as DateTime if it's in ISO format
+                if (DateTime.TryParse(stringValue, null, DateTimeStyles.RoundtripKind, out DateTime dateTime))
```
cgillum: This seems a bit risky. Could we provide some hints in the string value to know if it's actually a datetime or not?

cgillum (commented):
A few initial comments. Haven't reviewed everything yet.

src/Abstractions/TaskFailureDetails.cs
nytian: We can do that. The only thing is that at DT.Core we set this as IDictionary, so we just need to add some helper method here to convert those

cgillum: Probably not worth doing a conversion. This is fine as-is.

src/Shared/Grpc/ProtoUtils.cs
nytian: Add prefix `dt:` or `dto:` when we convert to protobuf

src/Abstractions/TaskFailureDetails.cs
```diff
@@ -160,14 +163,17 @@ internal CoreFailureDetails ToCoreFailureDetails()
                 coreEx.FailureDetails?.ErrorType ?? "(unknown)",
                 coreEx.FailureDetails?.ErrorMessage ?? "(unknown)",
                 coreEx.FailureDetails?.StackTrace,
-                FromCoreFailureDetailsRecursive(coreEx.FailureDetails?.InnerFailure) ?? FromExceptionRecursive(coreEx.InnerException));
+                FromCoreFailureDetailsRecursive(coreEx.FailureDetails?.InnerFailure) ?? FromExceptionRecursive(coreEx.InnerException),
+                coreEx.FailureDetails?.Properties ?? null);
```
philliphoff: Nit: is the `?? null` needed?

test/Grpc.IntegrationTests/GrpcSidecar/Grpc/ProtobufUtils.cs
```diff
@@ -367,14 +372,26 @@ public static string Base64Encode(IMessage message)
...
         };
+
+        // Add properties if they exist
+        if (failureDetails.Properties != null)
+        {
+            var mapField = ConvertDictionaryToMapField(failureDetails.Properties);
+            foreach (var kvp in mapField)
+            {
+                taskFailureDetails.Properties.Add(kvp.Key, kvp.Value);
+            }
```
philliphoff, line 390: Nit: Could this be (to avoid creating the `MapField` instance):

```csharp
foreach (var kvp in failureDetails.Properties)
{
  taskFailureDetails.Properties.Add(kvp.Key, ConvertObjectToValue(kvp.Value));
}
```

src/Worker/Core/Shims/DurableTaskShimFactory.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.DurableTask.Worker.Shims;
 /// </remarks>
 public class DurableTaskShimFactory
 {
+    public readonly IExceptionPropertiesProvider? exceptionPropertiesProvider;
```
cgillum: Was making this public an accident? We should never have public fields.

src/Worker/Core/IExceptionPropertiesProvider.cs
cgillum: This seems okay to me.

src/Worker/Grpc/GrpcDurableTaskWorker.cs
```diff
@@ -37,6 +37,30 @@ public GrpcDurableTaskWorker(
...
+    /// <param name="exceptionPropertiesProvider">The custom exception properties provider that help build failure details.</param>
+    public GrpcDurableTaskWorker(
+        string name,
+        IDurableTaskFactory factory,
+        IOptionsMonitor<GrpcDurableTaskWorkerOptions> grpcOptions,
+        IOptionsMonitor<DurableTaskWorkerOptions> workerOptions,
+        IServiceProvider services,
+        ILoggerFactory loggerFactory,
+        IOrchestrationFilter? orchestrationFilter = null,
+        IExceptionPropertiesProvider? exceptionPropertiesProvider = null)
```
cgillum, line 41: This class is internal, so you can just add this new parameter to the existing constructor if you want.

test/Grpc.IntegrationTests/GrpcSidecar/Grpc/ProtobufUtils.cs
```diff
@@ -438,4 +454,86 @@ internal static Proto.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeR
...
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.NullValue:
+                return null;
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.NumberValue:
+                return value.NumberValue;
+            case Google.Protobuf.WellKnownTypes.Value.KindOneofCase.StringValue:
+                string stringValue = value.StringValue;
+
+                if (stringValue.StartsWith("dt:", StringComparison.Ordinal))
+                {
+                    return DateTime.Parse(stringValue[3..], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
```
cgillum: We should still do `TryParse` here in case the user actually has a string value in some field that starts with `dt:`.

src/Worker/Grpc/GrpcOrchestrationRunner.cs
cgillum: I was able to read it properly if I turned off whitespace in the diff view. This can happen sometimes if VS or the git client changes the line endings.

src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -1,1172 +1,1262 @@
...
-        foreach (OrchestratorAction action in actions)
-        {
-            var protoAction = new P.OrchestratorAction { Id = action.Id };
-
-            P.TraceContext? CreateTraceContext()
-            {
-                if (orchestrationActivity is null)
-                {
-                    return null;
+                    return DateTime.Parse(stringValue[3..], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
```
cgillum: Same comment about TryParse instead of Parse, though I wonder why we have this same logic in two places?

nytian: The grpc integration test will not use the ProtoUtils under folder `src/shared/grpc`  but will use the one at it's own folder which is `test/Grpc.IntegrationTests/GrpcSidecar/Grpc/ProtobufUtils.cs`

src/Worker/Core/Shims/DurableTaskShimFactory.cs
nytian: oh thanks for reminding. I added these at first because I thought we need this when using with durable functions isolated. But the test turns out actually we don't need to propagate provider from worker extensions to here. I just deleted these.

PR #476, opened 2025-10-09
Add API for In-process Testing and Add Class-Syntax Integration Tests

nytian:
This PR Introduce a new api and package  `DurableTaskTestHost` for customer to test their orchestrations in-process

- **DurableTaskTestHost**: A wrapper around the gRPC Fixture that provides a simple API for testing orchestrations and activities in-process without external dependencies
- **DurableTaskTestHost.README.md**: Doc explaining how to use the API
- **InProcessTestHost.Tests** :test for this new added api and also samples.
- **ClassSyntaxIntegrationTests**: Integration tests for class-based orchestrator and activity


cgillum:
Is the goal of this feature to ship it as a public package for other developers to use? If so, we should put this in its own project and make it into a nuget package. But let me know if that was the intention here, or if it's something different.

src/InProcessTestHost/InProcessTestHost.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
cgillum: .NET 6.0 has been deprecated. We should instead target net8.0.

src/InProcessTestHost/InProcessTestHost.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <RootNamespace>Microsoft.DurableTask.Testing</RootNamespace>
+    <AssemblyName>Microsoft.DurableTask.InProcessTestHost</AssemblyName>
+    <PackageId>Microsoft.DurableTask.InProcessTestHost</PackageId>
+    <EnableStyleCop>false</EnableStyleCop>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
cgillum: What is this warning that we're supressing? Can we add a comment here to explain?

test/Grpc.IntegrationTests/ClassSyntaxIntegrationTests.cs
```diff
@@ -0,0 +1,163 @@
...
+namespace Microsoft.DurableTask.Grpc.Tests;
+
+/// <summary>
+/// Integration tests for class-based syntax orchestrators and activities.
+/// These tests demonstrate how to use TaskOrchestrator and TaskActivity base classes
+/// instead of function-based syntax.
+/// </summary>
+public class ClassSyntaxIntegrationTests : IntegrationTestBase
+{
+    public ClassSyntaxIntegrationTests(ITestOutputHelper output, GrpcSidecarFixture sidecarFixture)
```
cgillum, line 18: Is the `GrpcSidecarFixture` using the new test host internally?

src/InProcessTestHost/InProcessTestHost.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <RootNamespace>Microsoft.DurableTask.Testing</RootNamespace>
+    <AssemblyName>Microsoft.DurableTask.InProcessTestHost</AssemblyName>
+    <PackageId>Microsoft.DurableTask.InProcessTestHost</PackageId>
```
cgillum, line 7: I suggest we have a separate versioning scheme for this package, and start it with 0.1.0-preview.1.

This is important for allowing us to make breaking changes while we stabilize it. Once we see that users are successfully adopting it, we can remove the preview suffix.

src/InProcessTestHost/InProcessTestHost.csproj
nytian: oh it's just these moved files have so many format warnings... I tried to fix all but didn't finish. So, I just suppress it for now. But I will fix all of them ...

test/Grpc.IntegrationTests/ClassSyntaxIntegrationTests.cs
nytian: I don't think so,  DurableTaskTestHost is like the combination of `GrpcSidecarFixture` and `IntegrationTestBase`,. So they two do have some overlap

src/InProcessTestHost/InProcessTestHost.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
halspang, line 4: I know some of our packages support `netstandard2.0` and `net6`. Do we need to do that here as well or are we OK with just `net6`?

src/InProcessTestHost/Sidecar/Dispatcher/ITrafficSignal.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.DurableTask.Sidecar.Dispatcher;
 interface ITrafficSignal
 {
     /// <summary>
-    /// Provides a human-friendly reason for why the signal is in the "wait" state.
+    /// Gets provides a human-friendly reason for why the signal is in the "wait" state.
```
halspang: Nit, you don't need `provides`.

src/InProcessTestHost/Sidecar/AsyncManualResetEvent.cs
```diff
@@ -1,13 +1,24 @@
...
 namespace Microsoft.DurableTask.Sidecar;
 
+/// <summary>
+/// An asynchronous manual reset event implementation.
+/// </summary>
+/// <remarks>
+/// This class provides an asynchronous version of ManualResetEvent that can be used
+/// for synchronization in async/await scenarios.
+/// </remarks>
 class AsyncManualResetEvent
```
halspang, line 9: Can I have some more context on what this is? The comment just let's me know it's an async version of a thing that I also don't know what does :)

src/InProcessTestHost/Sidecar/Dispatcher/TaskOrchestrationDispatcher.cs
```diff
@@ -166,12 +232,24 @@ static int GetSortOrderWithinGroup(TaskMessage msg)
...
+        }
+    }
+
     void ApplyOrchestratorActions(
         OrchestratorExecutionResult result,
         ref OrchestrationRuntimeState runtimeState,
-        out IList<TaskMessage> activityMessages,
-        out IList<TaskMessage> orchestratorMessages,
-        out IList<TaskMessage> timerMessages,
+        out List<TaskMessage> activityMessages, // CA1859: Use concrete types for better performance
```
halspang, line 250: Nit, was this warning there before just not suppressed? Should we have a TODO to come back and fix it or does it not really matter since this is a test platform and that's talking about perf?

src/InProcessTestHost/Sidecar/Logs.cs
```diff
@@ -7,6 +7,9 @@
 
 namespace Microsoft.DurableTask.Sidecar
```
halspang: Heads up, you list the root namespace as `Microsoft.DurableTask.Testing` but everywhere under the `Sidecar` folder seems to drop the `Testing`. We should keep that in the namespace.

src/InProcessTestHost/Sidecar/AsyncManualResetEvent.cs
nytian: udpated

src/InProcessTestHost/Sidecar/Logs.cs
nytian: updated, thanks!

src/InProcessTestHost/InProcessTestHost.csproj
nytian: Just tried, we are using Grpc.AspNetCore.Server in this test pkg which doesn't support netstandard2.0. I guess we are fine?  If there are special customer asking for that then I can move back to add the TFM. Now it's just preview pkg

PR #477, opened 2025-10-12
Set low priority for scheduled runs

castrodd:
This PR sets the Priority for scheduled pipeline runs to be low. This is part of an effort to reduce the daily deadlock that we are facing with our various pipelines in 1es-pool-azfunc.

It also reschedules these pipeline runs to 5am UTC.


jviau:
@microsoft-github-policy-service rerun

PR #479, opened 2025-10-15
Add Rewind to .NET isolated

sophiatev:
This PR introduces the rewind API for .NET isolated apps. It also adds logic to correctly handle the new `ExecutionRewoundEvent` in the protos.

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -441,11 +441,13 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <summary>
     /// Rewinds the specified orchestration instance by re-executing any failed Activities, and recursively rewinding
     /// any failed suborchestrations with failed Activities.
+    /// </summary>
+    /// <remarks>
     /// The orchestration's history will be replaced with a new history that excludes the failed Activities and suborchestrations,
     /// and a new execution ID will be generated for the rewound orchestration instance. As the failed Activities and suborchestrations
     /// re-execute, the history will be appended with new TaskScheduled, TaskCompleted, and SubOrchestrationInstanceCompleted events.
     /// Note that only orchestrations in a "Failed" state can be rewound.
-    /// </summary>
+    /// </remarks>
     /// <param name="instanceId">The instance ID of the orchestration to rewind.</param>
     /// <param name="reason">The reason for the rewind.</param>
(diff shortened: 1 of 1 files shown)


src/Client/Core/DurableTaskClient.cs
```diff
@@ -438,7 +438,13 @@ public virtual Task<string> RestartAsync(
         string instanceId,
         bool restartWithNewInstanceId = false,
         CancellationToken cancellation = default)
-        => throw new NotSupportedException($"{this.GetType()} does not support orchestration restart.");
+        => throw new NotSupportedException($"{this.GetType()} does not support orchestration restart.");
+
+    public virtual Task RewindInstanceAsync(
```
cgillum, line 463: Looks like there are some warnings here about missing XML documentation comments.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -431,6 +431,41 @@ public override async Task<string> RestartAsync(
...
+        }
+        catch (RpcException e) when (e.StatusCode == StatusCode.NotFound)
+        {
+            throw new ArgumentException($"An orchestration with the instanceId {instanceId} was not found.", e);
+        }
+        catch (RpcException e) when (e.StatusCode == StatusCode.FailedPrecondition)
+        {
+            throw new InvalidOperationException(
+                $"The orchestration with the instanceId {instanceId} cannot be rewound. " +
+                $"Only orchestrations in the \"Failed\" state can be rewound.",
```
cgillum: Would it make sense to let the server generate this message instead of the client? That way if we change the meaning of `FailedPrecondition` or add other cases that return this error, we're always telling users the right thing, regardless of what version of the SDK they have.

src/Client/Core/DurableTaskClient.cs
```diff
@@ -440,6 +440,25 @@ public virtual Task<string> RestartAsync(
         CancellationToken cancellation = default)
         => throw new NotSupportedException($"{this.GetType()} does not support orchestration restart.");
 
+    /// <summary>
+    /// Rewinds the specified orchestration instance by re-executing any failed Activities, and recursively rewinding
+    /// any failed suborchestrations with failed Activities.
+    /// The orchestration's history will be replaced with a new history that excludes the failed Activities and suborchestrations,
```
cgillum, line 446: nit: The `<summary>...</summary` should just be a once sentence description of the API. The other details around behavior should be in the `<remarks>...</remarks>` section.

cgillum (commented):
One last minor thing.

PR #481, opened 2025-10-20
Export History

YunchuWang:


src/Abstractions/TaskOptions.cs
```diff
@@ -40,6 +41,11 @@ public TaskOptions(TaskRetryOptions? retry = null, IDictionary<string, string>?
     /// </summary>
     public IDictionary<string, string>? Tags { get; init; }
 
+    /// <summary>
+    /// Gets the flag to indicate if the task is exportable.
+    /// </summary>
+    public bool Exportable { get; init; }
```
halspang: Nit: For booleans like this, I prefer `IsExportable`.

PR #482, opened 2025-10-20
Include exception properties in failure details when orchestration throws directly

nytian:
This PR fixes an edge case where, if an orchestration using DurableTaskWorker throws an exception directly, the exception properties were not included in the failure details even when a custom exception provider was configured. Also add two more comprehensive test cases for the exception properties provider.

Fix PR: https://github.com/Azure/durabletask/pull/1251


nytian:
@microsoft-github-policy-service rerun

PR #494, opened 2025-11-05
Export History job

YunchuWang:
This pull request introduces a new sample web application, `ExportHistoryWebApp`, for managing export history jobs via a REST API. The changes add new project files, configuration files, and controller logic, as well as update the solution file to include the new projects and their build configurations.

### New Export History Web Application

* Added new sample project `ExportHistoryWebApp` to the solution, including its project file (`ExportHistoryWebApp.csproj`) and necessary references to Durable Task components and the new `ExportHistory` project. [[1]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eR104-R107) [[2]](diffhunk://#diff-7f813cfd4fa37c11640acd04cdeebf3654b3fab29c40f71bf0af06277ab84463R1-R23)
* Implemented `ExportJobController` to provide REST endpoints for creating, retrieving, listing, and deleting export history jobs. This controller uses the Durable Task ExportHistory client and models.
* Defined the `CreateExportJobRequest` model to represent the payload for creating export jobs, supporting various configuration options such as mode, time ranges, storage container, and filtering.

### Configuration and Environment Setup

* Added launch settings (`launchSettings.json`) and application configuration files (`appsettings.json`, `appsettings.Development.json`) to support local development and environment variable management for connection strings and storage settings. [[1]](diffhunk://#diff-3a2331dfe452afd471048460c2fc1665b80a0b5ac5ec75bab060b6b587b16754R1-R26) [[2]](diffhunk://#diff-a938bb044820395eb265f776a1951513b431d5bb803d0511adf7efe085850cddR1-R10) [[3]](diffhunk://#diff-802174564d7715f6d032a2044e222a1cb451264216f124ff7a7c092c4630adf5R1-R9)
* Added `Program.cs` to configure dependency injection, Durable Task client and worker services, and controller routing for the web application.

### API Usage and Documentation

* Provided an HTTP request collection file (`ExportHistoryWebApp.http`) with example requests for creating, listing, retrieving, and deleting export jobs, including tips and usage notes for common scenarios.

### Solution File Updates

* Updated `Microsoft.DurableTask.sln` to include the new `ExportHistory` and `ExportHistoryWebApp` projects, and adjusted build configuration and project nesting accordingly. [[1]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eR104-R107) [[2]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eL258-L265) [[3]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eR270-R285) [[4]](diffhunk://#diff-96995f9cf755bc5c62e1d6f130695fa92b1502b401f78173d908d96402371a7eL320-R337)

### DIFF
--- a/src/Client/Core/DurableTaskClient.cs
+++ b/src/Client/Core/DurableTaskClient.cs
@@ -370,6 +370,13 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <see cref="PurgeResult.PurgedInstanceCount"/> indicating the number of orchestration instances that were purged,
     /// including the count of sub-orchestrations purged if any.
     /// </returns>
+    /// <exception cref="ArgumentNullException">Thrown if <paramref name="instanceId"/> is null.</exception>
+    /// <exception cref="ArgumentException">Thrown if <paramref name="instanceId"/> is empty or starts with the null character.</exception>
+    /// <exception cref="InvalidOperationException">Thrown if the orchestration is not in a
+    /// <see cref="OrchestrationRuntimeStatus.Completed"/>, <see cref="OrchestrationRuntimeStatus.Failed"/>,
+    /// or <see cref="OrchestrationRuntimeStatus.Terminated"/> state.</exception>
+    /// <exception cref="NotImplementedException">Thrown if the backend storage provider does not support purging instances.</exception>
+    /// <exception cref="OperationCanceledException">Thrown if the operation is canceled via the <paramref name="cancellation"/> token.</exception>
     public virtual Task<PurgeResult> PurgeInstanceAsync(
         string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
     {
@@ -447,7 +454,7 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// The orchestration's history will be replaced with a new history that excludes the failed Activities and suborchestrations,
     /// and a new execution ID will be generated for the rewound orchestration instance. As the failed Activities and suborchestrations
     /// re-execute, the history will be appended with new TaskScheduled, TaskCompleted, and SubOrchestrationInstanceCompleted events.
-    /// Note that only orchestrations in a "Failed" state can be rewound.
+    /// Note that only orchestrations in a <see cref="OrchestrationRuntimeStatus.Failed"/> state can be rewound.
     /// </remarks>
     /// <param name="instanceId">The instance ID of the orchestration to rewind.</param>
     /// <param name="reason">The reason for the rewind.</param>
@@ -460,7 +467,7 @@ public abstract class DurableTaskClient : IOrchestrationSubmitter, IAsyncDisposa
     /// <exception cref="ArgumentException">Thrown if an orchestration with the specified <paramref name="instanceId"/> does not exist,
     /// or if <paramref name="instanceId"/> is the instance ID of an entity.</exception>
     /// <exception cref="InvalidOperationException">Thrown if a precondition of the operation fails, for example if the specified
-    /// orchestration is not in a "Failed" state.</exception>
+    /// orchestration is not in a <see cref="OrchestrationRuntimeStatus.Failed"/> state.</exception>
     /// <exception cref="OperationCanceledException">Thrown if the operation is canceled via the <paramref name="cancellation"/> token.</exception>
     public virtual Task RewindInstanceAsync(
         string instanceId,
--- a/src/Client/Grpc/GrpcDurableTaskClient.cs
+++ b/src/Client/Grpc/GrpcDurableTaskClient.cs
@@ -441,10 +441,16 @@ public sealed class GrpcDurableTaskClient : DurableTaskClient
     public override Task<PurgeResult> PurgeInstanceAsync(
         string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
     {
+        Check.NotNullOrEmpty(instanceId);
         bool recursive = options?.Recursive ?? false;
         this.logger.PurgingInstanceMetadata(instanceId);
 
-        P.PurgeInstancesRequest request = new() { InstanceId = instanceId, Recursive = recursive };
+        P.PurgeInstancesRequest request = new()
+        {
+            InstanceId = instanceId,
+            Recursive = recursive,
+            IsOrchestration = !this.options.EnableEntitySupport || instanceId[0] != '@',
+        };
         return this.PurgeInstancesCoreAsync(request, cancellation);
     }
 
@@ -645,6 +651,14 @@ public sealed class GrpcDurableTaskClient : DurableTaskClient
         {
             throw new OperationCanceledException(
                 $"The {nameof(this.PurgeAllInstancesAsync)} operation was canceled.", e, cancellation);
+        }
+        catch (RpcException e) when (e.StatusCode == StatusCode.FailedPrecondition)
+        {
+            throw new InvalidOperationException(e.Status.Detail);
+        }
+        catch (RpcException e) when (e.StatusCode == StatusCode.Unimplemented)
+        {
+            throw new NotImplementedException(e.Status.Detail);
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v1.19.1
+- Throw an `InvalidOperationException` for purge requests on running orchestrations by sophiatev ([#611](https://github.com/microsoft/durabletask-dotnet/pull/611))
+- Validate c# identifiers in durabletask source generator by Copilot ([#578](https://github.com/microsoft/durabletask-dotnet/pull/578))
+- Document orchestration discovery and method probing behavior in analyzers by Copilot ([#594](https://github.com/microsoft/durabletask-dotnet/pull/594))
+
 ## v1.19.0
 - Extended sessions for entities in .net isolated by sophiatev ([#507](https://github.com/microsoft/durabletask-dotnet/pull/507))
 - Adding the ability to specify tags and a retry policy for suborchestrations by sophiatev ([#603](https://github.com/microsoft/durabletask-dotnet/pull/603))
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -23,7 +23,7 @@
   <!-- Azure.* Packages -->
   <ItemGroup>
     <PackageVersion Include="Azure.Identity" Version="1.17.1" />
-    <PackageVersion Include="Azure.Storage.Blobs" Version="12.26.0" />
+    <PackageVersion Include="Azure.Storage.Blobs" Version="12.27.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker" Version="2.51.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.3.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.7" />
@@ -51,11 +51,11 @@
 
   <!-- Microsoft.CodeAnalysis.* Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.12.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.14.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.11.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.12.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.12.0" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.12.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.14.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing" Version="1.1.2" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing" Version="1.1.2" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing" Version="1.1.2" />
--- a/eng/targets/Release.props
+++ b/eng/targets/Release.props
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.19.0</VersionPrefix>
+    <VersionPrefix>1.19.1</VersionPrefix>
     <VersionSuffix></VersionSuffix>
   </PropertyGroup>
 
--- a/src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
+++ b/src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
@@ -122,6 +122,21 @@ class ShimDurableTaskClient(string name, ShimDurableTaskClientOptions options) :
         string instanceId, PurgeInstanceOptions? options = null, CancellationToken cancellation = default)
     {
         Check.NotNullOrEmpty(instanceId);
+        OrchestrationMetadata? orchestrationState = await this.GetInstanceAsync(instanceId, cancellation);
+
+        // The orchestration doesn't exist, nothing to purge
+        if (orchestrationState == null)
+        {
+            return new PurgeResult(0);
+        }
+
+        bool isEntity = this.options.EnableEntitySupport && instanceId[0] == '@';
+        if (!isEntity && !orchestrationState.IsCompleted)
+        {
+            throw new InvalidOperationException($"Only orchestrations in a terminal state can be purged, " +
+                $"but the orchestration with instance ID {instanceId} has status {orchestrationState.RuntimeStatus}");
+        }
+
         cancellation.ThrowIfCancellationRequested();
 
         // TODO: Support recursive purge of sub-orchestrations
--- a/src/Grpc/orchestrator_service.proto
+++ b/src/Grpc/orchestrator_service.proto
@@ -115,6 +115,7 @@ message SubOrchestrationInstanceCreatedEvent {
     google.protobuf.StringValue version = 3;
     google.protobuf.StringValue input = 4;
     TraceContext parentTraceContext = 5;
+    map<string, string> tags = 6;
 }
 
 message SubOrchestrationInstanceCompletedEvent {
@@ -225,6 +226,11 @@ message ExecutionRewoundEvent {
     google.protobuf.StringValue parentExecutionId = 2; // used only for rewinding suborchestrations, null otherwise
     google.protobuf.StringValue instanceId = 3; // used only for rewinding suborchestrations, null otherwise
     TraceContext parentTraceContext = 4; // used only for rewinding suborchestrations, null otherwise
+    google.protobuf.StringValue name = 5; // used by DTS backend only
+    google.protobuf.StringValue version = 6; // used by DTS backend only
+    google.protobuf.StringValue input = 7;  // used by DTS backend only
+    ParentInstanceInfo parentInstance = 8; // used by DTS backend only
+    map<string, string> tags = 9; // used by DTS backend only
 }
 
 message HistoryEvent {
@@ -496,8 +502,6 @@ message ListInstanceIdsResponse {
     google.protobuf.StringValue lastInstanceKey = 2;
 }
 
-// Removed ListTerminalInstances in favor of using QueryInstances
-
 message PurgeInstancesRequest {
     oneof request {
         string instanceId = 1;
@@ -505,6 +509,8 @@ message PurgeInstancesRequest {
         InstanceBatch instanceBatch = 4;
     }
     bool recursive = 3;
+    // used in the case when an instanceId is specified to determine if the purge request is for an orchestration (as opposed to an entity)
+    bool isOrchestration = 5;
 }
 
 message PurgeInstanceFilter {
@@ -765,7 +771,9 @@ service TaskHubSidecarService {
     // rpc DeleteInstance(DeleteInstanceRequest) returns (DeleteInstanceResponse);
 
     rpc QueryInstances(QueryInstancesRequest) returns (QueryInstancesResponse);
+
     rpc ListInstanceIds(ListInstanceIdsRequest) returns (ListInstanceIdsResponse);
+    
     rpc PurgeInstances(PurgeInstancesRequest) returns (PurgeInstancesResponse);
 
     rpc GetWorkItems(GetWorkItemsRequest) returns (stream WorkItem);
--- a/src/Grpc/versions.txt
+++ b/src/Grpc/versions.txt
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch main at 2025-12-29 22:13:55 UTC
-https://raw.githubusercontent.com/microsoft/durabletask-protobuf/b7e260ad7b84740a2ed5cb4600ce73bef702a979/protos/orchestrator_service.proto
+# The following files were downloaded from branch main at 2026-01-13 00:01:21 UTC
+https://raw.githubusercontent.com/microsoft/durabletask-protobuf/026329c53fe6363985655857b9ca848ec7238bd2/protos/orchestrator_service.proto
--- a/src/InProcessTestHost/DurableTaskTestExtensions.cs
+++ b/src/InProcessTestHost/DurableTaskTestExtensions.cs
@@ -1,0 +1,192 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using DurableTask.Core;
+using Grpc.Net.Client;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Client.Grpc;
+using Microsoft.DurableTask.Testing.Sidecar;
+using Microsoft.DurableTask.Testing.Sidecar.Grpc;
+using Microsoft.DurableTask.Worker;
+using Microsoft.DurableTask.Worker.Grpc;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.Testing;
+
+/// <summary>
+/// Extension methods for integrating in-memory durable task testing with your existing DI container,
+/// such as WebApplicationFactory.
+/// </summary>
+public static class DurableTaskTestExtensions
+{
+    /// <summary>
+    /// These extensions allow you to inject the <see cref="InMemoryOrchestrationService"/> into your
+    /// existing test host so that your orchestrations and activities can resolve services from your DI container.
+    /// </summary>
+    /// <param name="services">The service collection (from your WebApplicationFactory or host).</param>
+    /// <param name="configureTasks">Action to register orchestrators and activities.</param>
+    /// <param name="options">Optional configuration options.</param>
+    /// <returns>The service collection for chaining.</returns>
+    public static IServiceCollection AddInMemoryDurableTask(
+        this IServiceCollection services,
+        Action<DurableTaskRegistry> configureTasks,
+        InMemoryDurableTaskOptions? options = null)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+        ArgumentNullException.ThrowIfNull(configureTasks);
+
+        options ??= new InMemoryDurableTaskOptions();
+
+        // Determine port for the internal gRPC server
+        int port = options.Port ?? Random.Shared.Next(30000, 40000);
+        string address = $"http://localhost:{port}";
+
+        // Register the in-memory orchestration service as a singleton
+        services.AddSingleton<InMemoryOrchestrationService>(sp =>
+        {
+            var loggerFactory = sp.GetService<ILoggerFactory>();
+            return new InMemoryOrchestrationService(loggerFactory);
+        });
+        services.AddSingleton<IOrchestrationService>(sp => sp.GetRequiredService<InMemoryOrchestrationService>());
+        services.AddSingleton<IOrchestrationServiceClient>(sp => sp.GetRequiredService<InMemoryOrchestrationService>());
+
+        // Register the gRPC sidecar server as a hosted service
+        services.AddSingleton<TaskHubGrpcServer>();
+        services.AddHostedService<InMemoryGrpcSidecarHost>(sp =>
+        {
+            return new InMemoryGrpcSidecarHost(
+                address,
+                sp.GetRequiredService<InMemoryOrchestrationService>(),
+                sp.GetService<ILoggerFactory>());
+        });
+
+        // Create a gRPC channel that will connect to our internal sidecar
+        services.AddSingleton<GrpcChannel>(sp => GrpcChannel.ForAddress(address));
+
+        // Register the durable task worker (connects to our internal sidecar)
+        services.AddDurableTaskWorker(builder =>
+        {
+            builder.UseGrpc(address);
+            builder.AddTasks(configureTasks);
+        });
+
+        // Register the durable task client (connects to our internal sidecar)
+        services.AddDurableTaskClient(builder =>
+        {
+            builder.UseGrpc(address);
+            builder.RegisterDirectly();
+        });
+
+        return services;
+    }
+
+    /// <summary>
+    /// Gets the <see cref="InMemoryOrchestrationService"/> from the service provider.
+    /// Useful for advanced scenarios like inspecting orchestration state.
+    /// </summary>
+    /// <param name="services">The service provider.</param>
+    /// <returns>The in-memory orchestration service instance.</returns>
+    public static InMemoryOrchestrationService GetInMemoryOrchestrationService(this IServiceProvider services)
+    {
+        return services.GetRequiredService<InMemoryOrchestrationService>();
+    }
+}
+
+/// <summary>
+/// Options for configuring in-memory durable task support.
+/// </summary>
+public class InMemoryDurableTaskOptions
+{
+    /// <summary>
+    /// Gets or sets the port for the internal gRPC server.
+    /// If not set, a random port between 30000-40000 will be used.
+    /// </summary>
+    public int? Port { get; set; }
+}
+
+/// <summary>
+/// Internal hosted service that runs the gRPC sidecar within the user's host.
+/// </summary>
+sealed class InMemoryGrpcSidecarHost : IHostedService, IAsyncDisposable
+{
+    readonly string address;
+    readonly InMemoryOrchestrationService orchestrationService;
+    readonly ILoggerFactory? loggerFactory;
+    IHost? inMemorySidecarHost;
+
+    public InMemoryGrpcSidecarHost(
+        string address,
+        InMemoryOrchestrationService orchestrationService,
+        ILoggerFactory? loggerFactory)
+    {
+        this.address = address;
+        this.orchestrationService = orchestrationService;
+        this.loggerFactory = loggerFactory;
+    }
+
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        // Build and start the gRPC sidecar
+        this.inMemorySidecarHost = Host.CreateDefaultBuilder()
+            .ConfigureLogging(logging =>
+            {
+                logging.ClearProviders();
+                if (this.loggerFactory != null)
+                {
+                    logging.Services.AddSingleton(this.loggerFactory);
+                }
+            })
+            .ConfigureWebHostDefaults(webBuilder =>
+            {
+                webBuilder.UseUrls(this.address);
+                webBuilder.ConfigureKestrel(kestrelOptions =>
+                {
+                    kestrelOptions.ConfigureEndpointDefaults(listenOptions =>
+                        listenOptions.Protocols = HttpProtocols.Http2);
+                });
+
+                webBuilder.ConfigureServices(services =>
+                {
+                    services.AddGrpc();
+                    // Use the SAME orchestration service instance
+                    services.AddSingleton<IOrchestrationService>(this.orchestrationService);
+                    services.AddSingleton<IOrchestrationServiceClient>(this.orchestrationService);
+                    services.AddSingleton<TaskHubGrpcServer>();
+                });
+
+                webBuilder.Configure(app =>
+                {
+                    app.UseRouting();
+                    app.UseEndpoints(endpoints =>
+                    {
+                        endpoints.MapGrpcService<TaskHubGrpcServer>();
+                    });
+                });
+            })
+            .Build();
+
+        await this.inMemorySidecarHost.StartAsync(cancellationToken);
+    }
+
+    public async Task StopAsync(CancellationToken cancellationToken)
+    {
+        if (this.inMemorySidecarHost != null)
+        {
+            await this.inMemorySidecarHost.StopAsync(cancellationToken);
+        }
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        if (this.inMemorySidecarHost != null)
+        {
+            await this.inMemorySidecarHost.StopAsync();
+            this.inMemorySidecarHost.Dispose();
+        }
+    }
+}
--- a/src/InProcessTestHost/DurableTaskTestHost.cs
+++ b/src/InProcessTestHost/DurableTaskTestHost.cs
@@ -47,6 +47,12 @@ public sealed class DurableTaskTestHost : IAsyncDisposable
     public DurableTaskClient Client { get; }
 
     /// <summary>
+    /// Gets the service provider from the worker host.
+    /// Use this to resolve services registered via <see cref="DurableTaskTestHostOptions.ConfigureServices"/>.
+    /// </summary>
+    public IServiceProvider Services => this.workerHost.Services;
+
+    /// <summary>
     /// Starts a new in-process test host with the specified orchestrators and activities.
     /// </summary>
     /// <param name="registry">Action to configure the task registry by adding orchestrators and activities.</param>
@@ -113,6 +119,10 @@ public sealed class DurableTaskTestHost : IAsyncDisposable
             })
             .ConfigureServices(services =>
             {
+                // Allow user to register their own services FIRST
+                // This ensures their services are available when activities are resolved
+                options.ConfigureServices?.Invoke(services);
+
                 // Register worker that connects to our in-process sidecar
                 services.AddDurableTaskWorker(builder =>
                 {
@@ -170,4 +180,10 @@ public class DurableTaskTestHostOptions
     /// Null by default.
     /// </summary>
     public ILoggerFactory? LoggerFactory { get; set; }
+
+    /// <summary>
+    /// Gets or sets an optional callback to configure additional services in the worker host's DI container.
+    /// Use this to register services that your activities and orchestrators depend on.
+    /// </summary>
+    public Action<IServiceCollection>? ConfigureServices { get; set; }
 }
--- a/src/InProcessTestHost/InProcessTestHost.csproj
+++ b/src/InProcessTestHost/InProcessTestHost.csproj
@@ -5,7 +5,7 @@
     <RootNamespace>Microsoft.DurableTask.Testing</RootNamespace>
     <AssemblyName>Microsoft.DurableTask.InProcessTestHost</AssemblyName>
     <PackageId>Microsoft.DurableTask.InProcessTestHost</PackageId>
-    <Version>0.1.0-preview.1</Version>
+    <Version>0.2.0-preview.1</Version>
     
     <!-- Suppress CA1848: Use LoggerMessage delegates for high-performance logging scenarios -->
     <NoWarn>$(NoWarn);CA1848</NoWarn>
--- a/src/InProcessTestHost/README.md
+++ b/src/InProcessTestHost/README.md
@@ -6,17 +6,17 @@ Supports both **class-based** and **function-based** syntax.
 
 ## Quick Start
 
-1. Configure options
+### 1. Configure options (optional)
+
 ```csharp
 var options = new DurableTaskTestHostOptions
 {
     Port = 31000,                  // Optional: specific port (random by default)
     LoggerFactory = myLoggerFactory // Optional: pass logger factory for logging
 };
-
 ```
 
-2. Register test orchestrations and activities.
+### 2. Register test orchestrations and activities
 
 ```csharp
 await using var testHost = await DurableTaskTestHost.StartAsync(registry =>
@@ -29,15 +29,121 @@ await using var testHost = await DurableTaskTestHost.StartAsync(registry =>
     registry.AddOrchestratorFunc("MyFunc", (ctx, input) => Task.FromResult("done"));
     registry.AddActivityFunc("MyActivity", (ctx, input) => Task.FromResult("result"));
 });
-
 ```
 
-3. Test
+### 3. Test
+
 ```csharp
 string instanceId = await testHost.Client.ScheduleNewOrchestrationInstanceAsync("MyOrchestrator");
 var result = await testHost.Client.WaitForInstanceCompletionAsync(instanceId);
 ```
- .
+
+## Dependency Injection
+
+When your activities depend on services, there are two approaches:
+
+| Approach | When to Use |
+|----------|-------------|
+| **Option 1: ConfigureServices** | Simple tests where you register a few services directly |
+| **Option 2: AddInMemoryDurableTask** | When you have an existing host (e.g., `WebApplicationFactory`) with complex DI setup |
+
+### Option 1: ConfigureServices
+
+Use this when you want the test host to manage everything. Register services directly in the test host options.
+
+```csharp
+await using var host = await DurableTaskTestHost.StartAsync(
+    tasks =>
+    {
+        tasks.AddOrchestrator<MyOrchestrator>();
+        tasks.AddActivity<MyActivity>();
+    },
+    new DurableTaskTestHostOptions
+    {
+        ConfigureServices = services =>
+        {
+            // Register services required by your orchestrator or activity function
+            services.AddSingleton<IMyService, MyService>();
+            services.AddSingleton<IUserRepository, InMemoryUserRepository>();
+            services.AddLogging();
+        }
+    });
+
+var instanceId = await host.Client.ScheduleNewOrchestrationInstanceAsync(nameof(MyOrchestrator), "input");
+var result = await host.Client.WaitForInstanceCompletionAsync(instanceId, getInputsAndOutputs: true);
+```
+
+Access registered services via `host.Services`:
+
+```csharp
+var myService = host.Services.GetRequiredService<IMyService>();
+```
+
+### Option 2: AddInMemoryDurableTask
+
+Use this when you already have a host with complex DI setup (database, auth, external APIs, etc.) and want to add durable task testing to it.
+
+```csharp
+public class MyIntegrationTests : IAsyncLifetime
+{
+    IHost host = null!;
+    DurableTaskClient client = null!;
+
+    public async Task InitializeAsync()
+    {
+        this.host = Host.CreateDefaultBuilder()
+            .ConfigureServices(services =>
+            {
+                // Your existing services (from Program.cs, Startup.cs, etc.)
+                services.AddSingleton<IUserRepository, InMemoryUserRepository>();
+                services.AddScoped<IOrderService, OrderService>();
+                services.AddDbContext<MyDbContext>();
+
+                // Add in-memory durable task support
+                services.AddInMemoryDurableTask(tasks =>
+                {
+                    tasks.AddOrchestrator<MyOrchestrator>();
+                    tasks.AddActivity<MyActivity>();
+                });
+            })
+            .Build();
+
+        await this.host.StartAsync();
+        this.client = this.host.Services.GetRequiredService<DurableTaskClient>();
+    }
+}
+```
+
+Access the in-memory orchestration service:
+
+```csharp
+var orchestrationService = host.Services.GetInMemoryOrchestrationService();
+```
+
+## API Reference
+
+### DurableTaskTestHostOptions
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `Port` | `int?` | Specific port for gRPC sidecar. Random 30000-40000 if not set. |
+| `LoggerFactory` | `ILoggerFactory?` | Logger factory for capturing logs during tests. |
+| `ConfigureServices` | `Action<IServiceCollection>?` | Callback to register services for DI. |
+
+### DurableTaskTestHost
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `Client` | `DurableTaskClient` | Client for scheduling and managing orchestrations. |
+| `Services` | `IServiceProvider` | Service provider with registered services. |
+
+### Extension Methods
+
+| Method | Description |
+|--------|-------------|
+| `services.AddInMemoryDurableTask(configureTasks)` | Adds in-memory durable task support to an existing `IServiceCollection`. |
+| `services.GetInMemoryOrchestrationService()` | Gets the `InMemoryOrchestrationService` from the service provider. |
+
 ## More Samples
 
-See [BasicOrchestrationTests.cs](../../test/InProcessTestHost.Tests/BasicOrchestrationTests.cs) for complete samples showing both class-syntax and function-syntax orchestrations.
+See [BasicOrchestrationTests.cs](../../test/InProcessTestHost.Tests/BasicOrchestrationTests.cs), [DependencyInjectionTests.cs](../../test/InProcessTestHost.Tests/DependencyInjectionTests.cs), and [WebApplicationFactoryIntegrationTests.cs](../../test/InProcessTestHost.Tests/WebApplicationFactoryIntegrationTests.cs) for complete samples.
--- a/test/Client/OrchestrationServiceClientShim.Tests/ShimDurableTaskClientTests.cs
+++ b/test/Client/OrchestrationServiceClientShim.Tests/ShimDurableTaskClientTests.cs
@@ -165,7 +165,9 @@ public class ShimDurableTaskClientTests
     public async Task PurgeInstanceMetadata()
     {
         // arrange
-        string instanceId = Guid.NewGuid().ToString();
+        List<Core.OrchestrationState> states = [CreateState("input", "output")];
+        string instanceId = states.First().OrchestrationInstance.InstanceId;
+        this.orchestrationClient.Setup(m => m.GetOrchestrationStateAsync(instanceId, false)).ReturnsAsync(states);
         this.purgeClient.Setup(m => m.PurgeInstanceStateAsync(instanceId)).ReturnsAsync(new Core.PurgeResult(1));
 
         // act
--- a/test/InProcessTestHost.Tests/BasicOrchestrationTests.cs
+++ b/test/InProcessTestHost.Tests/BasicOrchestrationTests.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
 using Microsoft.DurableTask;
 using Microsoft.DurableTask.Client;
 using Microsoft.DurableTask.Testing;
--- a/test/InProcessTestHost.Tests/DependencyInjectionTests.cs
+++ b/test/InProcessTestHost.Tests/DependencyInjectionTests.cs
@@ -1,0 +1,318 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Testing;
+using Microsoft.DurableTask.Worker;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Xunit;
+
+namespace InProcessTestHost.Tests;
+
+/// <summary>
+/// Tests to verify DurableTaskTestHost works with dependency injection.
+/// </summary>
+public class DependencyInjectionTests
+{
+    /// <summary>
+    /// Verifies an activity can resolve a service registered via ConfigureServices.
+    /// </summary>
+    [Fact]
+    public async Task Activity_CanResolveService_FromDI()
+    {
+        await using var host = await DurableTaskTestHost.StartAsync(
+            tasks =>
+            {
+                tasks.AddOrchestrator<GreetingOrchestrator>();
+                tasks.AddActivity<GreetingActivity>();
+            },
+            new DurableTaskTestHostOptions
+            {
+                ConfigureServices = services =>
+                {
+                    services.AddSingleton<IGreetingService, GreetingService>();
+                }
+            });
+
+        var instanceId = await host.Client.ScheduleNewOrchestrationInstanceAsync(
+            nameof(GreetingOrchestrator),
+            "World");
+
+        var result = await host.Client.WaitForInstanceCompletionAsync(
+            instanceId,
+            getInputsAndOutputs: true);
+
+        Assert.NotNull(result);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, result.RuntimeStatus);
+        var output = result.ReadOutputAs<string>();
+        Assert.Equal("Hello, World! (from DI service)", output);
+    }
+
+    /// <summary>
+    /// Verifies multiple activities can share the same DI-registered service.
+    /// </summary>
+    [Fact]
+    public async Task Activity_CanUseMultipleServices_FromDI()
+    {
+        await using var host = await DurableTaskTestHost.StartAsync(
+            tasks =>
+            {
+                tasks.AddOrchestrator<CalculatorOrchestrator>();
+                tasks.AddActivity<AddActivity>();
+                tasks.AddActivity<MultiplyActivity>();
+            },
+            new DurableTaskTestHostOptions
+            {
+                ConfigureServices = services =>
+                {
+                    services.AddSingleton<ICalculatorService, CalculatorService>();
+                }
+            });
+
+        var instanceId = await host.Client.ScheduleNewOrchestrationInstanceAsync(
+            nameof(CalculatorOrchestrator),
+            new CalculatorInput { A = 5, B = 3 });
+
+        var result = await host.Client.WaitForInstanceCompletionAsync(
+            instanceId,
+            getInputsAndOutputs: true);
+
+        Assert.NotNull(result);
+        Assert.Equal(OrchestrationRuntimeStatus.Completed, result.RuntimeStatus);
+        var output = result.ReadOutputAs<CalculatorOutput>();
+        Assert.NotNull(output);
+        Assert.Equal(8, output.Sum);      // 5 + 3 = 8
+        Assert.Equal(15, output.Product); // 5 * 3 = 15
+    }
+
+    /// <summary>
+    /// Verifies host.Services exposes the DI container for direct service access.
+    /// </summary>
+    [Fact]
+    public async Task Services_Property_AllowsAccessToRegisteredServices()
+    {
+        await using var host = await DurableTaskTestHost.StartAsync(
+            tasks =>
+            {
+                tasks.AddOrchestrator<GreetingOrchestrator>();
+                tasks.AddActivity<GreetingActivity>();
+            },
+            new DurableTaskTestHostOptions
+            {
+                ConfigureServices = services =>
+                {
+                    services.AddSingleton<IGreetingService, GreetingService>();
+                }
+            });
+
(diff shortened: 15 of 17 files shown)


samples/ExportHistoryWebApp/Models/CreateExportJobRequest.cs
```diff
@@ -0,0 +1,60 @@
...
+    public string? Prefix { get; set; }
+
+    /// <summary>
+    /// Gets or sets the export format settings. Optional, defaults to jsonl-gzip.
+    /// </summary>
+    public ExportFormat? Format { get; set; }
+
+    /// <summary>
+    /// Gets or sets the orchestration runtime statuses to filter by. Optional.
+    /// Valid statuses are: Completed, Failed, Terminated, and ContinuedAsNew.
```
halspang: Is `ContinueAsNew` an actual valid status? I thought it just stayed `Running`/`Pending` the whole time?

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+    [HttpGet("list")]
+    public async Task<ActionResult<IEnumerable<ExportJobDescription>>> ListExportJobs(
+        [FromQuery] ExportJobStatus? status = null,
+        [FromQuery] string? jobIdPrefix = null,
+        [FromQuery] DateTimeOffset? createdFrom = null,
+        [FromQuery] DateTimeOffset? createdTo = null,
+        [FromQuery] int? pageSize = null,
+        [FromQuery] string? continuationToken = null)
+    {
+        this.logger.LogInformation("GET list endpoint called with method: {Method}", this.HttpContext.Request.Method);
```
halspang, line 132: Nit: You specify this is a GET, why do we log the HTTP Method?

samples/ExportHistoryWebApp/Models/CreateExportJobRequest.cs
halspang: Is `ContinueAsNew` a valid status? I thought when this happened, the orchestration just stayed in the `Running` state or maybe went `Pending`.

src/Client/Core/DurableTaskClient.cs
```diff
@@ -1,441 +1,464 @@
-﻿// Copyright (c) Microsoft Corporation.
```
halspang, line 1: For my sanity, can you explain what happened here? Why does git think you deleted and remade this file?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -323,6 +325,55 @@ public override async Task<OrchestrationMetadata> WaitForInstanceStartAsync(
...
     }
 
+    // Removed ListTerminalInstances; use GetAllInstancesAsync with OrchestrationQuery instead
+
+    /// <inheritdoc/>
+    public override async Task<Page<string>> ListInstanceIdsAsync(
+        IEnumerable<OrchestrationRuntimeStatus>? runtimeStatus = null,
+        DateTimeOffset? completedTimeFrom = null,
+        DateTimeOffset? completedTimeTo = null,
+        int pageSize = OrchestrationQuery.DefaultPageSize,
```
halspang, line 361: If we're going to let users set this, we should force a max value just so we don't get lists that try and fetch millions of objects at once.

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,322 @@
...
+    /// <summary>
+    /// Gets the file extension for a given export format.
+    /// </summary>
+    /// <param name="format">The export format.</param>
+    /// <returns>The file extension (e.g., "json", "jsonl.gz").</returns>
+    static string GetFileExtension(ExportFormat format)
+    {
+        string formatKind = format.Kind.ToLowerInvariant();
+
+        return formatKind switch
```
halspang: Nit: Any reason we don't just make the format type an enum?

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,322 @@
...
+    async Task UploadToBlobStorageAsync(
+        string containerName,
+        string blobPath,
+        string content,
+        ExportFormat format,
+        string instanceId,
+        CancellationToken cancellationToken)
+    {
+        // Create blob service client from connection string
+        BlobServiceClient serviceClient = new(this.storageOptions.ConnectionString);
```
halspang, line 210: Are any of this clients disposable and should we be using `using` statements?

src/ExportHistory/Activities/ListTerminalInstancesActivity.cs
```diff
@@ -0,0 +1,90 @@
...
+    readonly ILogger<ListTerminalInstancesActivity> logger = Check.NotNull(logger, nameof(logger));
+
+    /// <inheritdoc/>
+    public override async Task<InstancePage> RunAsync(TaskActivityContext context, ListTerminalInstancesRequest input)
+    {
+        Check.NotNull(input, nameof(input));
+
+        try
+        {
+            DurableTaskClient client = this.clientProvider.GetClient();
```
halspang: Nit: Any reason you don't just take the client in the DI instead of the provider?

src/ExportHistory/Client/DefaultExportHistoryClient.cs
```diff
@@ -0,0 +1,190 @@
...
+        ExportJobCreationOptions options,
+        CancellationToken cancellation = default)
+    {
+        Check.NotNull(options, nameof(options));
+        this.logger.ClientCreatingExportJob(options);
+
+        try
+        {
+            // Create export job client instance
+            ExportHistoryJobClient exportHistoryJobClient = new DefaultExportHistoryJobClient(
```
halspang: Nit: Why not use your helper method?

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
```diff
@@ -0,0 +1,216 @@
...
+        try
+        {
+            Check.NotNull(options, nameof(options));
+
+            // Determine default prefix based on mode if not already set
+            string? defaultPrefix = $"{options.Mode.ToString().ToLower(System.Globalization.CultureInfo.CurrentCulture)}/";
+
+            // If destination is not provided, construct it from storage options
+            ExportJobCreationOptions optionsWithDestination = options;
+            if (options.Destination == null)
```
halspang: I feel like this is a better user experience if we don't let them create a job without a destination. constructing a default I think is a recipe for people exporting to locations they didn't mean to.

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
```diff
@@ -0,0 +1,216 @@
...
+            await this.durableTaskClient.WaitForInstanceCompletionAsync(orchestrationInstanceId, cancellation);
+
+            // Purge the orchestration instance after it's terminated
+            await this.durableTaskClient.PurgeInstanceAsync(orchestrationInstanceId, cancellation: cancellation);
+        }
+        catch (RpcException ex) when (ex.StatusCode == StatusCode.NotFound)
+        {
+            // Orchestration instance doesn't exist - this is expected if it was already deleted or never existed
+            this.logger.LogInformation(
+                "Orchestration instance '{OrchestrationInstanceId}' is already purged",
```
halspang: Nit: Do we technically know that this is true? What if they tried to terminate/purge an orchestration that didn't ever exist?

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
```diff
@@ -0,0 +1,216 @@
...
+    public override async Task DeleteAsync(CancellationToken cancellation = default)
+    {
+        try
+        {
+            this.logger.ClientDeletingExportJob(this.JobId);
+
+            string orchestrationInstanceId = ExportHistoryConstants.GetOrchestratorInstanceId(this.JobId);
+
+            // First, delete the entity
+            ExportJobOperationRequest request = new ExportJobOperationRequest(this.entityId, ExportJobOperations.Delete);
```
halspang: Is this how we have to do this? It feels odd to schedule an orchestration with a specific operation. I'd rather we just have different orchestrations for doing different things, I think it's a cleaner interface.

src/ExportHistory/Client/ExportHistoryClient.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Convenience client for managing export jobs via entity signals and reads.
+/// </summary>
+public abstract class ExportHistoryClient
```
halspang, line 9: What's the reasoning for having the `ExportHistoryClient` only do job operations but there is also a `ExportHistoryJobClient` that does other job operations?

src/ExportHistory/Client/ExportHistoryClient.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <param name="filter">The filter for the export jobs.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public abstract AsyncPageable<ExportJobDescription> ListJobsAsync(ExportJobQuery? filter = null);
+
+    /// <summary>
+    /// Gets an export job client.
+    /// </summary>
+    /// <param name="jobId">The ID of the export job.</param>
+    /// <returns>The export job client.</returns>
+    public abstract ExportHistoryJobClient GetJobClient(string jobId);
```
halspang, line 39: It feels a little weird to me to have this be part of the abstract definition.

src/ExportHistory/Client/ExportHistoryJobClient.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Convenience client for managing export jobs via entity signals and reads.
+/// </summary>
+/// <remarks>
+/// Initializes a new instance of the <see cref="ExportHistoryJobClient"/> class.
+/// </remarks>
+public abstract class ExportHistoryJobClient(string jobId)
```
halspang, line 12: Having a dependency on the JobId makes this feel more like a tracker/metadata object than a client.

src/ExportHistory/Models/ExportJobTransitions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <param name="from">The current export job state.</param>
+    /// <param name="targetState">The target state to transition to.</param>
+    /// <returns>True if the transition is valid; otherwise, false.</returns>
+    public static bool IsValidTransition(string operationName, ExportJobStatus from, ExportJobStatus targetState)
+    {
+        return operationName switch
+        {
+            nameof(ExportJob.Create) => from switch
+            {
+                ExportJobStatus.Uninitialized when targetState == ExportJobStatus.Active => true,
```
halspang: Does this mean `Uninitialized` is a defined state rather than having no state at all being `Uninitialized`?

And with `Completed` and `Failed`, is this path basically restarting/recreating the job?

src/ExportHistory/Entity/ExportJobOperations.cs
```diff
@@ -0,0 +1,19 @@
...
+/// <remarks>
+/// Operation names are case-insensitive when matching to entity methods.
+/// These constants match the method names on <see cref="ExportJob"/> for consistency.
+/// </remarks>
+static class ExportJobOperations
+{
+    /// <summary>
+    /// Operation name for deleting the entity.
+    /// </summary>
+    public const string Delete = "Delete";
```
halspang: We can only delete?

src/ExportHistory/Exception/ExportJobClientValidationException.cs
```diff
@@ -0,0 +1,38 @@
...
+        this.JobId = jobId;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ExportJobClientValidationException"/> class.
+    /// </summary>
+    /// <param name="jobId">The ID of the export job that failed validation.</param>
+    /// <param name="message">The validation error message.</param>
+    /// <param name="innerException">The exception that is the cause of the current exception.</param>
+    public ExportJobClientValidationException(string jobId, string message, Exception innerException)
```
halspang: Just for code simplification, it seems like you could just do a single constructor with an optional exception: `ExportJobClientValidationException(string JobId, string message, Exception? innerException = null)`

Unless that freaks out the base compiler, in which case, disregard this. If it does work though, please do it for all your exceptions.

src/ExportHistory/Models/CommitCheckpointRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Gets or sets the checkpoint to commit. If not null, the checkpoint is updated (cursor moves forward).
+    /// If null, the current checkpoint is kept (cursor does not move forward), allowing retry of the same batch.
+    /// </summary>
+    public ExportCheckpoint? Checkpoint { get; set; }
+
+    /// <summary>
+    /// Gets or sets the list of failed instance exports, if any.
+    /// </summary>
+    public List<ExportFailure>? Failures { get; set; }
```
halspang, line 30: Are we expecting to accumulate a lot of errors or will an error cancel the job?

src/ExportHistory/Models/ExportCheckpoint.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Checkpoint information used to resume export.
+/// </summary>
+public sealed record ExportCheckpoint(string? LastInstanceKey = null);
```
halspang, line 9: Do you think we need a class for this or can we just store a string of that name in the bigger object? It's fine if you want this, just thinking about the overall amount of code we're generating here.

src/ExportHistory/Models/ExportFailure.cs
```diff
@@ -0,0 +1,13 @@
...
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Failure of a specific instance export.
+/// </summary>
+/// <param name="InstanceId">The instance ID that failed to export.</param>
+/// <param name="Reason">The reason for the failure.</param>
+/// <param name="AttemptCount">The number of attempts made.</param>
+/// <param name="LastAttempt">The timestamp of the last attempt.</param>
+public sealed record ExportFailure(string InstanceId, string Reason, int AttemptCount, DateTimeOffset LastAttempt);
```
halspang, line 13: Nit: Do we usually capitalize these?

src/ExportHistory/Models/ExportFormat.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Export format settings.
+/// </summary>
+/// <param name="Kind">The kind of export format.</param>
+/// <param name="SchemaVersion">The schema version.</param>
+public partial record ExportFormat(
+    string Kind = "jsonl",
```
halspang: I'd like if this was an enum, personally.

src/ExportHistory/Models/ExportJobConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <param name="Format">The export format settings.</param>
+/// <param name="MaxParallelExports">The maximum number of parallel export operations. Defaults to 32.</param>
+/// <param name="MaxInstancesPerBatch">The maximum number of instances to fetch per batch. Defaults to 100.</param>
+public record ExportJobConfiguration(
+    ExportMode Mode,
+    ExportFilter Filter,
+    ExportDestination Destination,
+    ExportFormat Format,
+    int MaxParallelExports = 32,
+    int MaxInstancesPerBatch = 100);
```
halspang, line 21: This needs a max on it somewhere.

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+        ExportFormat? format = null,
+        List<OrchestrationRuntimeStatus>? runtimeStatus = null,
+        int? maxInstancesPerBatch = null)
+    {
+        // Generate GUID if jobId not provided
+        this.JobId = string.IsNullOrEmpty(jobId) ? Guid.NewGuid().ToString("N") : jobId;
+
+        if (mode == ExportMode.Batch)
+        {
+            if (completedTimeFrom == default)
```
halspang: Is `default` null for a `DateTimeOffset`?

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <param name="completedTimeTo">The end time for the export based on completion time (inclusive). Required for Batch mode, null for Continuous mode.</param>
+    /// <param name="destination">The export destination where exported data will be stored. Required unless default storage is configured.</param>
+    /// <param name="jobId">The unique identifier for the export job. If not provided, a GUID will be generated.</param>
+    /// <param name="format">The export format settings. Optional, defaults to jsonl-gzip.</param>
+    /// <param name="runtimeStatus">The orchestration runtime statuses to filter by. Optional.</param>
+    /// <param name="maxInstancesPerBatch">The maximum number of instances to fetch per batch. Optional, defaults to 100.</param>
+    /// <exception cref="ArgumentException">Thrown when validation fails.</exception>
+    public ExportJobCreationOptions(
+        ExportMode mode,
+        DateTimeOffset completedTimeFrom,
```
halspang: What are your thoughts between a `DateTime` and a `DateTimeOffset` here? Are you trying to enable the case of someone specifying `ago(2d)`? Otherwise, I think a `DateTime` makes more sense because they can just give a date. Makes things more deterministic.

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+        if (mode == ExportMode.Batch)
+        {
+            if (completedTimeFrom == default)
+            {
+                throw new ArgumentException(
+                    "CompletedTimeFrom is required for Batch export mode.",
+                    nameof(completedTimeFrom));
+            }
+
+            if (!completedTimeTo.HasValue)
```
halspang, line 53: Any reason we do a different check here than `== default`? Not that I'm sure what the `default` means there, but it's important to be consistent for readability.

halspang: Also couldn't this be an NPE? It's a nullable object. Or does `HasValue` handle that?

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+            if (completedTimeTo.HasValue && completedTimeTo.Value > DateTimeOffset.UtcNow)
+            {
+                throw new ArgumentException(
+                    $"CompletedTimeTo({completedTimeTo.Value}) cannot be in the future. It must be less than or equal to the current time ({DateTimeOffset.UtcNow}).",
+                    nameof(completedTimeTo));
+            }
+        }
+        else if (mode == ExportMode.Continuous)
+        {
+            if (completedTimeFrom != default)
```
halspang: Don't we need a starting point even for continuous?

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+        else
+        {
+            throw new ArgumentException(
+                "Invalid export mode.",
+                nameof(mode));
+        }
+
+        // Validate maxInstancesPerBatch range if provided (must be 1..999)
+        if (maxInstancesPerBatch.HasValue && (maxInstancesPerBatch.Value <= 0 || maxInstancesPerBatch.Value >= 1001))
```
halspang, line 91: Oh hey, the max value for the page size. Guess we can ignore my other comments about needing it :)

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(maxInstancesPerBatch),
+                maxInstancesPerBatch,
+                "MaxInstancesPerBatch must be between 1 and 1000.");
+        }
+
+        // Validate terminal status-only filter here if provided
+        if (runtimeStatus is { Count: > 0 }
```
halspang, line 100: `is { Count: > 0 }` does what?

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+    public string JobId { get; init; }
+
+    /// <summary>
+    /// Gets the export mode (Batch or Continuous).
+    /// </summary>
+    public ExportMode Mode { get; init; }
+
+    /// <summary>
+    /// Gets the start time for the export based on completion time (inclusive).
+    /// Required for Batch mode. For Continuous mode, this is automatically set to UtcNow when creating the job.
```
halspang, line 139: The code requires it to be null, could be worth calling out here, but I also don't think we have to avoid this. They could want to start a continuous flow but also include the previous x minutes/hours.

src/ExportHistory/Models/ExportJobDescription.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// <summary>
+    /// Gets the time when this export job was created.
+    /// </summary>
+    public DateTimeOffset? CreatedAt { get; init; }
+
+    /// <summary>
+    /// Gets the time when this export job was last modified.
+    /// </summary>
+    public DateTimeOffset? LastModifiedAt { get; init; }
```
halspang, line 29: What modifications are legal for this?

src/ExportHistory/Models/ExportJobState.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Export job state stored in the entity.
+/// </summary>
+public sealed class ExportJobState
```
halspang, line 9: This feels very redundant to `ExportJobDescription`. Do we need both?

src/ExportHistory/Models/ExportJobStatus.cs
```diff
@@ -0,0 +1,30 @@
...
+
+/// <summary>
+/// Represents the current status of an export history job.
+/// </summary>
+public enum ExportJobStatus
+{
+    /// <summary>
+    /// Export history job has not been created.
+    /// </summary>
+    Uninitialized,
```
halspang: It is very odd to me to have a status that means we didn't create it. It's impossible to have a status and not be created, the absence of status is `Uninitialized`.

src/ExportHistory/Models/ExportFormat.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Export format settings.
+/// </summary>
+/// <param name="Kind">The kind of export format.</param>
+/// <param name="SchemaVersion">The schema version.</param>
+public partial record ExportFormat(
```
halspang: Why is this record partial?

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,397 @@
...
+
+                if (currentState == null ||
+                    currentState.Config == null ||
+                    currentState.Status != ExportJobStatus.Active)
+                {
+                    logger.ExportJobOperationWarning(jobId, nameof(ExportJobOrchestrator), "Job is no longer active");
+                    return null;
+                }
+
+                // if (currentState.Checkpoint is not null &&
```
halspang: Commented code.

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,397 @@
...
+                {
+                    if (config.Mode == ExportMode.Continuous)
+                    {
+                        logger.ExportJobOperationInfo(jobId, nameof(ExportJobOrchestrator), "No more instances to export - export complete.");
+                        await context.CreateTimer(ContinuousExportIdleDelay, default);
+                        continue;
+                    }
+                    else if (config.Mode == ExportMode.Batch)
+                    {
+                        logger.ExportJobOperationInfo(jobId, nameof(ExportJobOrchestrator), "No more instances to export currently - will retry later.");
```
halspang, line 114: This makes me wonder if we should force all batch operations to have an end date in the past so they can actually run and complete instead of waiting.

halspang (commented):
Haven't gotten through to the tests yet, but wanted to publish what I have so far.

src/ExportHistory/Models/ExportFailure.cs
halspang: Answering my own question, I've remembered that records like capitalized members.

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+        catch (ArgumentException ex)
+        {
+            this.logger.LogError(ex, "Validation failed while creating export job {JobId}", request.JobId);
+            return this.BadRequest(ex.Message);
+        }
+        catch (Exception ex)
+        {
+            this.logger.LogError(ex, "Error creating export job {JobId}", request.JobId);
+            return this.StatusCode(500, "An error occurred while creating the export job");
+        }
```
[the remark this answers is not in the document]
YunchuWang: no we need to return a generic error

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+            return this.NoContent();
+        }
+        catch (ExportJobNotFoundException)
+        {
+            this.logger.LogWarning("Export job {JobId} not found for deletion", id);
+            return this.NotFound();
+        }
+        catch (Exception ex)
+        {
+            this.logger.LogError(ex, "Error deleting export job {JobId}", id);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+        try
+        {
+            ExportHistoryJobClient jobClient = this.exportHistoryClient.GetJobClient(id);
+            await jobClient.DeleteAsync();
+            this.logger.LogInformation("Successfully deleted export job {JobId}", id);
+            return this.NoContent();
+        }
+        catch (ExportJobNotFoundException)
+        {
+            this.logger.LogWarning("Export job {JobId} not found for deletion", id);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <returns>No content if successful.</returns>
+    [HttpDelete("{id}")]
+    public async Task<IActionResult> DeleteExportJob(string id)
+    {
+        this.logger.LogInformation("DELETE endpoint called for job ID: {JobId}", id);
+        try
+        {
+            ExportHistoryJobClient jobClient = this.exportHistoryClient.GetJobClient(id);
+            await jobClient.DeleteAsync();
+            this.logger.LogInformation("Successfully deleted export job {JobId}", id);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// <summary>
+    /// Deletes an export job by its ID.
+    /// </summary>
+    /// <param name="id">The ID of the export job to delete.</param>
+    /// <returns>No content if successful.</returns>
+    [HttpDelete("{id}")]
+    public async Task<IActionResult> DeleteExportJob(string id)
+    {
+        this.logger.LogInformation("DELETE endpoint called for job ID: {JobId}", id);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+            ExportJobDescription? job = await this.exportHistoryClient.GetJobAsync(id);
+            return this.Ok(job);
+        }
+        catch (ExportJobNotFoundException)
+        {
+            return this.NotFound();
+        }
+        catch (Exception ex)
+        {
+            this.logger.LogError(ex, "Error retrieving export job {JobId}", id);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+            return this.CreatedAtAction(nameof(GetExportJob), new { id = description.JobId }, description);
+        }
+        catch (ArgumentException ex)
+        {
+            this.logger.LogError(ex, "Validation failed while creating export job {JobId}", request.JobId);
+            return this.BadRequest(ex.Message);
+        }
+        catch (Exception ex)
+        {
+            this.logger.LogError(ex, "Error creating export job {JobId}", request.JobId);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+            ExportHistoryJobClient jobClient = await this.exportHistoryClient.CreateJobAsync(creationOptions);
+            ExportJobDescription description = await jobClient.DescribeAsync();
+
+            this.logger.LogInformation("Created new export job with ID: {JobId}", description.JobId);
+
+            return this.CreatedAtAction(nameof(GetExportJob), new { id = description.JobId }, description);
+        }
+        catch (ArgumentException ex)
+        {
+            this.logger.LogError(ex, "Validation failed while creating export job {JobId}", request.JobId);
```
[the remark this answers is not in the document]
YunchuWang: this is intended

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: this is equivalent to if (runtimeStatus != null && runtimeStatus.Count > 0)

test/ExportHistory.Tests/Options/ExportHistoryStorageOptionsTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        var options = new ExportHistoryStorageOptions
+        {
+            Prefix = "test-prefix",
+        };
+
+        // Act
+        options.Prefix = null;
+
+        // Assert
+        options.Prefix.Should().BeNull();
```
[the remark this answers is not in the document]
YunchuWang: this is intended

samples/ExportHistoryWebApp/Models/CreateExportJobRequest.cs
YunchuWang: removed

src/Client/Grpc/GrpcDurableTaskClient.cs
YunchuWang: its already validated in createscheduleoptions

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
YunchuWang: / Note: Azure.Storage.Blobs clients (BlobServiceClient, BlobContainerClient, BlobClient) are lightweight
        // wrappers that don't implement IDisposable/IAsyncDisposable. They use HttpClient internally which is
        // managed by the framework and doesn't require explicit disposal. The clients are designed to be
        // stateless and safe for reuse or short-lived usage without disposal.

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
YunchuWang: will update

src/ExportHistory/Client/DefaultExportHistoryClient.cs
```diff
@@ -0,0 +1,192 @@
...
+            throw;
+        }
+        catch (OperationCanceledException) when (cancellation.IsCancellationRequested)
+        {
+            // the operation was cancelled as requested. No need to log this.
+            throw;
+        }
+        catch (Exception ex)
+        {
+            this.logger.ClientError(nameof(this.GetJobAsync), jobId, ex);
```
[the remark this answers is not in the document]
YunchuWang: intended

src/ExportHistory/Client/DefaultExportHistoryClient.cs
```diff
@@ -0,0 +1,192 @@
...
+            return exportHistoryJobClient;
+        }
+        catch (OperationCanceledException) when (cancellation.IsCancellationRequested)
+        {
+            // the operation was cancelled as requested. No need to log this.
+            throw;
+        }
+        catch (Exception ex)
+        {
+            this.logger.ClientError(nameof(this.CreateJobAsync), options.JobId, ex);
```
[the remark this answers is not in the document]
YunchuWang: intended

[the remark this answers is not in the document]
YunchuWang: intended

src/ExportHistory/Models/ExportJobState.cs
YunchuWang: yes, state is for whole metadata, description is for customer view.

src/ExportHistory/Models/ExportJobConfiguration.cs
YunchuWang: as in other comment, already in place

src/Client/Core/DurableTaskClient.cs
YunchuWang: unsure seems git is messed up. i reset it, will update in next push

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,397 @@
...
+                    new TaskOptions(ExportActivityRetryPolicy)));
+
+            // Limit parallel export activities
+            if (exportTasks.Count >= config.MaxParallelExports)
+            {
+                ExportResult[] batchResults = await Task.WhenAll(exportTasks);
+                results.AddRange(batchResults);
+                exportTasks.Clear();
+            }
+        }
```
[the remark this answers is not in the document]
YunchuWang: no its more readable this way

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,397 @@
...
+                            "; ",
+                            batchResult.Failures
+                                .Take(10)
+                                .Select(f =>
+                                    $"InstanceId: {f.InstanceId}, Reason: {f.Reason}"));
+                    }
+                    else
+                    {
+                        failureDetails = "No failure details available";
+                    }
```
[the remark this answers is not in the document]
YunchuWang: no its more readable this way

src/ExportHistory/Client/ExportHistoryClient.cs
YunchuWang: its same design style as scheduled task. it offers separation of concerns: collection operations vs. instance operations
Scoped operations: once you have a job client, you don't pass jobId repeatedly
Type safety: jobId is validated once in the constructor
API clarity: collection-level vs. instance-level operations

src/ExportHistory/Client/ExportHistoryClient.cs
YunchuWang: same as above comment, also azure blob sdk. it is easier for customers to operate on the jobclient for job level operations only.

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
YunchuWang: customers need to set the destination explicitly when configuring it, they wont export to some locations unintended. and this way they can reuse container name or prefix instead of specifying it every time when creating the export job. if they want to specify it fully, they can create job with destination. we provide both options.  builder.Services.AddDurableTaskClient(clientBuilder =>
{
    clientBuilder.UseDurableTaskScheduler(connectionString);
    clientBuilder.UseExportHistory(options =>
    {
        options.ConnectionString = storageConnectionString;
        options.ContainerName = containerName;
        options.Prefix = builder.Configuration.GetValue<string>("EXPORT_HISTORY_PREFIX");
    });
});

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
YunchuWang: yes and no. their flows (as we discussed in last meeting) are very similar (except exit condition). The flows are nearly identical except for the exit condition. Splitting them into separate orchestrators would duplicate logic and increase maintenance overhead, i want to keep a single parameterized orchestrator keeps the logic simple for now and if we have future needs to branch off, we can refactor them into two.

src/ExportHistory/Client/ExportHistoryJobClient.cs
YunchuWang: this is consistent with scheduled task client pattern as discussed. This pattern is standard across Azure SDKs — a client bound to a specific resource ID. It avoids repeating the identifier on every call and keeps the API surface clean

src/ExportHistory/Entity/ExportJobOperations.cs
YunchuWang: will remove

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
YunchuWang: yes we already have that in creationjoboptions validation for batch mode, for continous mode though customers cant set end time, and we want it  to run forever

src/ExportHistory/Models/ExportJobDescription.cs
YunchuWang: it will be 
Create() (line 54) - Sets LastModifiedAt when the job is created
CommitCheckpoint() (line 137) - Updates LastModifiedAt when a checkpoint is committed
MarkAsCompleted() (line 167) - Updates LastModifiedAt when the job completes
MarkAsFailed() (line 207) - Updates LastModifiedAt when the job fails
StartExportOrchestration() (line 273) - Updates LastModifiedAt when the export orchestration starts

src/ExportHistory/Models/ExportFormat.cs
YunchuWang: will update

src/ExportHistory/Models/ExportFormat.cs
YunchuWang: updated to enum

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: i think continuous is for now to future? and batch handles the past. this makes their scopes not overlap and easy to manage.

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: will update

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: will remove default, it is confusing. we will do explicit null check

src/ExportHistory/Models/CommitCheckpointRequest.cs
YunchuWang: yes it will mark it failed with failures from current batch. Failures property is only for reporting the failures from the batch that caused the job to fail - it's not a mechanism for accumulating errors across multiple batches.

src/ExportHistory/Models/ExportCheckpoint.cs
YunchuWang: I prefer keeping ExportCheckpoint as a dedicated type.
Checkpoints tend to evolve (paging, timestamps, etags, schema versions),
and having a structured type keeps the export schema stable and extensible
without breaking existing clients.

src/ExportHistory/Models/ExportJobTransitions.cs
YunchuWang: Yes, Uninitialized (value 0) is a defined state. It represents the default state of a new entity that has been allocated but not yet logically created or configured. This follows the .NET best practice of having an explicit "None" or "Uninitialized" value as the default (0) for enums to prevent unconfigured objects from appearing valid.
Regarding Completed and Failed: Yes, the transitions allowing these states to go back to Active via Create imply that we support restarting or recreating a job on an existing entity.

YunchuWang: actually the progress should be reset when restarting, will fix

src/ExportHistory/Models/ExportJobStatus.cs
YunchuWang: A state machine requires a true initial state. Uninitialized serves as that “Start” state before any transitions occur.
When ExportJob.Create runs, it validates transitions via CanTransitionTo.
Without Uninitialized, the default state would be Active, making creation look like an invalid Active → Active transition.
By introducing Uninitialized, new jobs correctly follow Uninitialized → Active, making the lifecycle explicit and keeping the logic clear: object allocated in memory ≠ job logically created.

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: will update to default to utcnow for continuous mode if not set.

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: i would favor datetimeoffset. 
These filters operate on exact completion timestamps. The underlying gRPC ListInstanceIds API requires precise UTC instants via .ToTimestamp(). DateTime without an explicit UTC kind introduces ambiguity (if someone run same code in different zones want to export same utc time range of orchestrations, they need to convert utc time to local datetime then pass it)

It matches the rest of the codebase. All timestamp-related APIs—StartOrchestrationOptions.StartAt, OrchestrationQuery.CreatedFrom/To, and ListInstanceIdsAsync—already use DateTimeOffset, and all gRPC conversions rely on DateTimeOffset.ToTimestamp().

Clients still get flexibility. Even though the API doesn’t parse expressions like ago(2d), clients can compute DateTimeOffset.UtcNow.AddDays(-2) themselves and pass a precise, offset-aware timestamp.

samples/ExportHistoryWebApp/Models/CreateExportJobRequest.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.ExportHistory;
+
+namespace ExportHistoryWebApp.Models;
+
+/// <summary>
+/// Represents a request to create a new export job.
+/// </summary>
+public class CreateExportJobRequest
```
halspang, line 12: Is there a reason this is in the sample? Shouldn't we provide a standard request for export jobs in the SDK?

halspang: I now see that it's just for the controller, but I wonder if it's better to have the controller take the options? Just wondering what we want to encourage with users.

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <param name="id">The ID of the export job to delete.</param>
+    /// <returns>No content if successful.</returns>
+    [HttpDelete("{id}")]
+    public async Task<IActionResult> DeleteExportJob(string id)
+    {
+        this.logger.LogInformation("DELETE endpoint called for job ID: {JobId}", id);
+        try
+        {
+            ExportHistoryJobClient jobClient = this.exportHistoryClient.GetJobClient(id);
+            await jobClient.DeleteAsync();
```
halspang, line 186: From a user perspective, is this deleting an old job, or is this how you stop an ongoing job? If it's for stopping, we should update the API name. If it's strictly like a purge, we can keep it, but maybe use the same terminology from the existing APIs.

src/Client/Core/DurableTaskClient.cs
```diff
@@ -464,15 +465,61 @@ public virtual Task RewindInstanceAsync(
...
+    /// Lists orchestration instance IDs filtered by completed time.
+    /// </summary>
+    /// <param name="runtimeStatus">The runtime statuses to filter by.</param>
+    /// <param name="completedTimeFrom">The start time for completed time filter (inclusive).</param>
+    /// <param name="completedTimeTo">The end time for completed time filter (inclusive).</param>
+    /// <param name="pageSize">The page size for pagination.</param>
+    /// <param name="lastInstanceKey">The last fetched instance key.</param>
+    /// <param name="cancellation">The cancellation token.</param>
+    /// <returns>A page of instance IDs with continuation token.</returns>
+    public virtual Task<Page<string>> ListInstanceIdsAsync(
```
halspang, line 521: Is this related to the PR? Also, if we're going to state in the comment that this is for listing _with_ a filter, we should name it less generically than `ListInstancesById`.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -468,6 +517,64 @@ public override async Task RewindInstanceAsync(
...
+
+        using AsyncServerStreamingCall<P.HistoryChunk> streamResponse =
+            this.sidecarClient.StreamInstanceHistory(request, cancellationToken: cancellation);
+
+        IAsyncStreamReader<P.HistoryChunk> responseStream = streamResponse.ResponseStream;
+
+        // Create conversion state to maintain orchestration instance across events
+        // This is required for entity-related events (EntityOperationCalled, EntityLockRequested, etc.)
+        // which need the parent orchestration instance information from ExecutionStartedEvent
+        Microsoft.DurableTask.ProtoUtils.EntityConversionState conversionState = new(insertMissingEntityUnlocks: false);
```
halspang: Are entities exported in the same way as an orchestration's history?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -468,6 +517,64 @@ public override async Task RewindInstanceAsync(
...
+            }
+            catch (RpcException e) when (e.StatusCode == StatusCode.Cancelled)
+            {
+                throw new OperationCanceledException(
+                    $"The {nameof(this.StreamInstanceHistoryAsync)} operation was canceled.", e, cancellation);
+            }
+            catch (RpcException e) when (e.StatusCode == StatusCode.NotFound)
+            {
+                throw new ArgumentException($"An orchestration with the instanceId {instanceId} was not found.", e);
+            }
```
halspang: Do we generally throw the generic exception out? Or should we be wrapping it in something?

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,321 @@
...
+    /// <summary>
+    /// Gets the file extension for a given export format.
+    /// </summary>
+    /// <param name="format">The export format.</param>
+    /// <returns>The file extension (e.g., "json", "jsonl.gz").</returns>
+    static string GetFileExtension(ExportFormat format)
+    {
+        return format.Kind switch
+        {
+            ExportFormatKind.Jsonl => "jsonl.gz",  // JSONL format is compressed
```
halspang, line 154: Is there any negative to using the compressed format? If not, I wonder if we need to provide the other format at all?

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,321 @@
...
+        // Create blob service client from connection string
+        // Note: Azure.Storage.Blobs clients (BlobServiceClient, BlobContainerClient, BlobClient) are lightweight
+        // wrappers that don't implement IDisposable/IAsyncDisposable. They use HttpClient internally which is
+        // managed by the framework and doesn't require explicit disposal. The clients are designed to be
+        // stateless and safe for reuse or short-lived usage without disposal.
+        BlobServiceClient serviceClient = new(this.storageOptions.ConnectionString);
+        BlobContainerClient containerClient = serviceClient.GetBlobContainerClient(containerName);
+
+        // Ensure container exists
+        await containerClient.CreateIfNotExistsAsync(
```
halspang, line 214: Is there anyway we can do this in an earlier step? Checking the container exists every upload is a lot of extra operations.

src/ExportHistory/Activities/ListTerminalInstancesActivity.cs
```diff
@@ -0,0 +1,88 @@
...
+    readonly ILogger<ListTerminalInstancesActivity> logger = Check.NotNull(logger, nameof(logger));
+
+    /// <inheritdoc/>
+    public override async Task<InstancePage> RunAsync(TaskActivityContext context, ListTerminalInstancesRequest input)
+    {
+        Check.NotNull(input, nameof(input));
+
+        try
+        {
+            // Try to use ListInstanceIds endpoint first (available in gRPC client)
```
halspang: What do we use next if not this?

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
```diff
@@ -0,0 +1,203 @@
...
+
+            throw;
+        }
+    }
+
+    // TODO: there is no atomicity guarantee of deleting entity and purging the orchestrator
+    // Add sweeping process to clean up orphaned orchestrations failed to be purged
+
+    /// <inheritdoc/>
+    public override async Task DeleteAsync(CancellationToken cancellation = default)
```
halspang, line 90: Do we want to standardize this to a Purge operation to match the API's existing naming scheme? Or should we update it to `Stop` to more directly indicate what this is doing?

samples/ExportHistoryWebApp/ExportJobController.cs
```diff
@@ -0,0 +1,202 @@
...
+        this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    /// <summary>
+    /// Creates a new export job based on the provided configuration.
+    /// </summary>
+    /// <param name="request">The export job creation request.</param>
+    /// <returns>The created export job description.</returns>
+    [HttpPost]
+    public async Task<ActionResult<ExportJobDescription>> CreateExportJob([FromBody] CreateExportJobRequest request)
```
halspang, line 42: Instead of having users create their own request, should we just show them using the options here as the domain object? I want to make sure we set the correct precedent for them.

samples/ExportHistoryWebApp/ExportJobController.cs
halspang, line 186: Do we need the pattern of getting a dedicated client for this? Or do you think we could wrap this into the `ExportHistoryClient`? Just wondering if we think it's worth it for the extra layer of abstraction?

samples/ExportHistoryWebApp/ExportJobController.cs
halspang, line 42: Do you think we should have users define this or do you think we should have users use the `ExportJobCreationOptions` in the controller? I'm worried we're asking them to define something that we should otherwise provide to the user?

samples/ExportHistoryWebApp/ExportJobController.cs
halspang, line 186: Do you think we really need both the `ExportHistoryJobClient` and the `ExportHistoryClient`? It feels like we could wrap the behavior into just one of these and reduce our complexity to users.

src/Client/Core/DurableTaskClient.cs
halspang, line 521: Instead of this generic name, if this is designed to always filter by completed time (as per the comment), we should update the method name to reflect that.

Or, if the comment is wrong, update that to have this be more generic. It feels more like a query than a list with all the filters.

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,321 @@
...
+
+        return $"{hash}.{extension}";
+    }
+
+    /// <summary>
+    /// Gets the file extension for a given export format.
+    /// </summary>
+    /// <param name="format">The export format.</param>
+    /// <returns>The file extension (e.g., "json", "jsonl.gz").</returns>
+    static string GetFileExtension(ExportFormat format)
```
halspang, line 150: Do we get a benefit from supporting two formats from the start? I would assume, to save costs, we'd default to the compressed format. Or, does that introduce an issue with things like Kusto ingestion?

Alternatively, are you trying to provide a lower compute by higher storage option?

src/Client/Core/DurableTaskClient.cs
halspang: Also, how does this differ from `GetAllInstancesAsync`? Is it just the query/filter ability?

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// Initializes a new instance of the <see cref="ExportJobCreationOptions"/> class.
+    /// </summary>
+    public ExportJobCreationOptions()
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ExportJobCreationOptions"/> class.
+    /// </summary>
+    /// <param name="mode">The export mode (Batch or Continuous).</param>
```
halspang, line 22: I thought originally we were going to say a continuous one was just an export with no end time? Did we decide to have specific modes just for increased specificity?

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,376 @@
...
+
+                // Call activity to list terminal instances with only necessary information
+                ListTerminalInstancesRequest listRequest = new ListTerminalInstancesRequest(
+                    CompletedTimeFrom: currentState.Config.Filter.CompletedTimeFrom,
+                    CompletedTimeTo: currentState.Config.Filter.CompletedTimeTo,
+                    RuntimeStatus: currentState.Config.Filter.RuntimeStatus,
+                    LastInstanceKey: currentState.Checkpoint?.LastInstanceKey,
+                    MaxInstancesPerBatch: currentState.Config.MaxInstancesPerBatch);
+
+                InstancePage pageResult = await context.CallActivityAsync<InstancePage>(
```
halspang, line 97: Late to the party, but is this why we do `ListInstancesById` instead of `GetAllInstancesAsync`? Because we want to do manual paging? If so, you can disregard my previous comment asking why we do it :)

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,376 @@
...
+                        break;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("Invalid export mode.");
+                    }
+                }
+
+                // Process batch with retry logic
+                BatchExportResult batchResult = await this.ProcessBatchWithRetryAsync(
```
halspang, line 124: Should stuff like this be an activity for additional checkpointing? Same with the commit.

YunchuWang:
@copilot resolve the conflicts

samples/ExportHistoryWebApp/Models/CreateExportJobRequest.cs
YunchuWang: i think DTO pattern in this case is better, it decouples from options evolution and avoid overexposure

YunchuWang: this shows people can plug in options by different ways also

samples/ExportHistoryWebApp/ExportJobController.cs
YunchuWang: its for deleting the export job. it is consistent with src\ScheduledTasks\Client\ScheduleClient.cs delete schedule function. I think delete is more user friendly here

src/Client/Core/DurableTaskClient.cs
YunchuWang: will clarify in comment

src/Client/Grpc/GrpcDurableTaskClient.cs
YunchuWang: no entities not supported, they wont have history events to export

src/Client/Grpc/GrpcDurableTaskClient.cs
YunchuWang: not applicable anymore

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
YunchuWang: i see a few
Debuggability: jsonl.gz can’t be easily inspected in the Azure Portal or with basic tools without an extra decompress step, which slows down ad-hoc troubleshooting.

Partial reads: Gzip is poorly suited for range reads or random access. Fetching or processing only part of a blob via HTTP ranges becomes impractical.

CPU overhead: Usually small, but for very small exports it’s pure overhead, and at high scale repeated compress/decompress steps can add up in downstream pipelines.

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
YunchuWang: I think export is not latency-sensitive, and the additional Blob call is negligible compared to the overall cost of history retrieval, serialization, compression, and upload. Performing the container existence check here also keeps the export path self-contained and resilient, avoiding hidden dependencies on external provisioning order.

This is particularly reasonable given that the container name may be user-provided, exports can be invoked ad-hoc, and we cannot always assume the destination container has been pre-provisioned.

src/ExportHistory/Activities/ListTerminalInstancesActivity.cs
YunchuWang: will remove this comment

src/ExportHistory/Client/DefaultExportHistoryJobClient.cs
YunchuWang: ditto, i think keeping as delete is fine and consistent with schedules

samples/ExportHistoryWebApp/ExportJobController.cs
YunchuWang: i think this is more intuitive and keeps separation of concern, consistent with schedules too

YunchuWang: ditto, i would prefer this for consistency with schedule.

samples/ExportHistoryWebApp/ExportJobController.cs
YunchuWang: ditto

YunchuWang: ditto

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
YunchuWang: ditto

src/ExportHistory/Models/ExportJobCreationOptions.cs
YunchuWang: yes we said to do two modes for clarity though their behaviors are similar.

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
YunchuWang: its fanning out to activities per instance export inside this function, this function itself(wrapping fan out code) does not need to be in an activity because the fanned out activities already do the checkpointing, even it rerun the wrapping code, it will skip the exported activities. the wrapping code itself is trivial to be checkpointed and we cant fan out activities inside a parent activity

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
YunchuWang: thats true you get me

src/Client/Core/DurableTaskClient.cs
YunchuWang: as below

src/ExportHistory/Activities/ExportInstanceHistoryActivity.cs
```diff
@@ -0,0 +1,314 @@
...
+            }
+
+            // Stream all history events
+            this.logger.LogInformation("Streaming history events for instance {InstanceId}", instanceId);
+            List<HistoryEvent> historyEvents = this.client.GetOrchestrationHistoryAsync(instanceId, CancellationToken.None);
+
+            this.logger.LogInformation(
+                "Retrieved {EventCount} history events for instance {InstanceId}",
+                historyEvents.Count,
+                instanceId);
```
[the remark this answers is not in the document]
YunchuWang: no count is in icollection

src/ExportHistory/Models/ExportJobCreationOptions.cs
```diff
@@ -0,0 +1,169 @@
...
+        if (runtimeStatus is { Count: > 0 }
+            && runtimeStatus.Any(
+                s => s is not (OrchestrationRuntimeStatus.Completed
+                               or OrchestrationRuntimeStatus.Failed
+                               or OrchestrationRuntimeStatus.Terminated)))
+        {
+            throw new ArgumentException(
+                "Export supports terminal orchestration statuses only. Valid statuses are: Completed, Failed, and Terminated.",
+                nameof(runtimeStatus));
+        }
```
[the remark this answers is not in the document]
YunchuWang: no longer true

src/ExportHistory/Client/ExportHistoryClient.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public abstract Task<ExportHistoryJobClient> CreateJobAsync(ExportJobCreationOptions options, CancellationToken cancellation = default);
+
+    /// <summary>
+    /// Gets an export job.
+    /// </summary>
+    /// <param name="jobId">The ID of the export job.</param>
+    /// <param name="cancellation">The cancellation token.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    public abstract Task<ExportJobDescription> GetJobAsync(string jobId, CancellationToken cancellation = default);
```
halspang, line 25: Is there a real difference between this and the `ExportHistoryJobClient.DescribeJob`? I'm still not convinced we need the two different clients.

src/ExportHistory/Models/ExportJobStatus.cs
halspang: I still don't like this status. In order to have a status, you have to have a job, so `Uninitialized` indicates that the job hasn't been created but it has to exist as an object to be in this state.

If we want to keep a status like this, could we use `Pending`? That matches our existing status schema and also gives a good idea that the job has been created but is not yet active.

src/ExportHistory/Client/ExportHistoryClient.cs
halspang, line 25: How does this differ from the `ExportHistoryJobClient.DescribeJob`? I'm still not sold on the idea of two separate clients.

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,376 @@
...
+            // Get the export job state and configuration from the entity
+            ExportJobState? jobState = await context.Entities.CallEntityAsync<ExportJobState?>(
+                input.JobEntityId,
+                nameof(ExportJob.Get),
+                null);
+
+            if (jobState == null || jobState.Config == null)
+            {
+                throw new InvalidOperationException($"Export job '{jobId}' not found or has no configuration.");
+            }
```
halspang, line 53: To confirm, we do this because we expect a job to have already been created elsewhere, right?

src/ExportHistory/ExportHistory.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
halspang: Are we still on `net6`?

samples/ExportHistoryWebApp/Program.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
```
halspang, line 1: Can we get a README for this sample talking about how to interact with the export stuff?

halspang (commented):
Coming together overall, just a few questions.

src/ExportHistory/Client/ExportHistoryClient.cs
YunchuWang: Great question. These two clients follow a common SDK design pattern but serve **distinct roles**.

## ExportHistoryClient — collection-level client
This is the root client for managing export jobs as a set. It supports operations that act across jobs, including:

- `CreateJobAsync()` — create a new export job  
- `GetJobAsync(jobId)` — retrieve a job by ID  
- `ListJobsAsync()` — list or query jobs with filtering  
- `GetJobClient(jobId)` — obtain a client scoped to a specific job  

## ExportHistoryJobClient — instance-level client
This client is scoped to a single export job and operates on that specific instance:

- `CreateAsync()` — create the job (invoked internally by the collection client)  
- `DescribeAsync()` — retrieve detailed status and metadata  
- `DeleteAsync()` — delete the job  

## Why split them?

- **Clear separation of concerns**  
  Collection-wide operations (create, list, query) are separated from instance-specific operations (describe, delete).

- **Established SDK pattern**  
  This mirrors Azure SDK hierarchies such as `BlobServiceClient → BlobContainerClient → BlobClient`.

- **Stateful job context**  
  `ExportHistoryJobClient` retains the `jobId`, avoiding repeated parameter passing:

  ```csharp
  // Instead of:
  await client.DescribeJobAsync(jobId);
  await client.DeleteJobAsync(jobId);

  // You do:
  ExportHistoryJobClient jobClient = client.GetJobClient(jobId);
  await jobClient.DescribeAsync();
  await jobClient.DeleteAsync();

- **Better encapsulation**  
The job client internally manages identifiers (e.g., EntityInstanceId) and orchestration coordination, hiding those details from callers.

Because ExportHistoryJobClient already carries job-specific context and state, DescribeAsync() is naturally an instance operation rather than a stateless query on the collection client. This makes the API more expressive, safer, and easier to use.

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
YunchuWang: Yes, exactly. The export job entity should have been created and configured before the orchestrator runs. This check ensures we fail fast with a clear error message if the job doesn't exist or wasn't properly configured, rather than encountering a NullReferenceException later in the orchestration.

samples/ExportHistoryWebApp/Program.cs
YunchuWang: updated

src/ExportHistory/ExportHistory.csproj
YunchuWang: updated

src/ExportHistory/Client/ExportHistoryClient.cs
YunchuWang: explained in comment above

src/ExportHistory/Models/ExportJobStatus.cs
YunchuWang: make sense, i updated to pending

src/ExportHistory/Orchestrations/ExportJobOrchestrator.cs
```diff
@@ -0,0 +1,376 @@
...
+        EntityInstanceId jobEntityId,
+        string? errorMessage)
+    {
+        await context.Entities.CallEntityAsync(
+            jobEntityId,
+            nameof(ExportJob.MarkAsFailed),
+            errorMessage);
+    }
+
+    sealed class BatchExportResult
```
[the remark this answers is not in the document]
YunchuWang: its already sealed, false alarm

src/ExportHistory/Entity/ExportJob.cs
```diff
@@ -0,0 +1,300 @@
...
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DurableTask.ExportHistory;
+
+/// <summary>
+/// Durable entity that manages a history export job: lifecycle, configuration, and progress.
+/// </summary>
+/// <param name="logger">The logger instance.</param>
+class ExportJob(ILogger<ExportJob> logger) : TaskEntity<ExportJobState>
```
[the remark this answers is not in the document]
YunchuWang: noise

halspang (approved):
I think we should follow-up with an example of what to do with your exported data, but we can do that later, potentially just in the docs.

PR #496, opened 2025-11-07
Initial attempt to fix carryover events issue on continue-as-new

### LINKED ISSUE #387: External events may be lost when using ContinueAsNew
### Repro steps

Consider the following orchestrator code:

```csharp
async Task<int> OrchestratorFunc(TaskOrchestrationContext ctx, int counter)
{
    await ctx.WaitForExternalEvent<string>("Event");
    counter++;

    if (injectTimers)
    {
        await ctx.CreateTimer(TimeSpan.Zero, CancellationToken.None);
    }

    if (counter < EventCount)
    {
        ctx.ContinueAsNew(counter, preserveUnprocessedEvents: true);
    }

    return counter;
}
```

If you send `EventCount` events quickly, the orchestration will sometimes fail to complete. If `injectTimers` is true, it seems to fail more frequently.

cgillum:
This is an in-progress attempt to fix #387. These changes are really old and have been stashed away on my dev box for many months. I've merged it with the latest from main and fixed the merge conflicts to the best of my ability.

The bug can be reproduced using the `ContinueAsNewEventsArePreserved` test in test/Grpc.IntegrationTests/OrchestrationPatterns.cs.

As far as I remember, I haven't been able to get the test to pass. I'm creating this draft PR as a reference for someone who can hopefully carry this work to the finish line.

### DIFF
--- a/src/Shared/Grpc/ProtoUtils.cs
+++ b/src/Shared/Grpc/ProtoUtils.cs
@@ -394,7 +394,7 @@ static class ProtoUtils
                         && sendEventAction.EventName is not null
                         && sendEventAction.EventData is not null)
                     {
-                        P.SendEntityMessageAction sendAction = new();
+                        P.SendEntityMessageAction sendAction = new P.SendEntityMessageAction();
                         protoAction.SendEntityMessage = sendAction;
 
                         EntityConversions.DecodeEntityMessageAction(


src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -461,11 +461,8 @@ internal static P.OrchestratorResponse ConstructOrchestratorResponse(
 
                     var completeAction = (OrchestrationCompleteOrchestratorAction)action;
                     protoAction.CompleteOrchestration = new P.CompleteOrchestrationAction
-                    {
-                        CarryoverEvents =
-                        {
-                            // TODO
-                        },
+                    {
+                        CarryoverEvents = { completeAction.CarryoverEvents.Select(ToProtobuf) },
```
cgillum, line 465: Should we be using the existing conversion code for this instead of creating a new `ToProtobuf` method?
```suggestion
                        CarryoverEvents = { completeAction.CarryoverEvents.Select(ConvertHistoryEvent) },
```

sophiatev: That's from proto to `HistoryEvent`, right? Whereas we need this to go from `HistoryEvent` to the proto type

cgillum: Ah, I see.

cgillum:
@sophiatev your changes look good to me. We'll need another person to sign off on this since I'm the original creator of this PR.

nytian (approved):

PR #503, opened 2025-11-13
Introduce custom copilot review instructions

halspang:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,75 @@
+# GitHub Copilot Instructions
+
+This repository contains C# code.
+
+The purpose of the code is to provide a standalone SDK for interacting with Durable Functions.
```
cgillum: Not just Durable Functions! Durable Functions is an Azure Functions feature, and this SDK is used for both Azure Functions and non-Functions via DTS (and theoretically for other backends, like Dapr Workflow - we're not super opinionated about this since we just communicate with a gRPC contract).

.github/copilot-instructions.md
```diff
@@ -0,0 +1,75 @@
...
+- Environment variables should use upper snake_case naming convention.
+- Secrets should not be hardcoded in the code or committed to the repository.
+- The code should be well-documented with comments explaining the purpose of each step.
+- The code should be simple and to the point, avoiding unnecessary complexity.
+- Prefer inline literals over constants for values that are not reused. For example, use `new ChatClientAgent(chatClient, instructions: "You are a helpful assistant.")` instead of defining a constant for "instructions
+- Use the Async suffix on the name of all async methods.
+- Prefer defining variables using types rather than var, to help users understand the types involved.
+- Follow the patterns in the samples in the same directories where new samples are being added.
+- The structure of the sample should be as follows:
+  - The top of the Program.cs should have a copyright notice:
```
cgillum: nit: this is redundant with the above.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,75 @@
...
+  - Finally, add any helper methods or classes at the bottom of the file.
+
+### C# Unit Test Guidelines
+
+Unit tests are located in the `test` root directory in projects with a `.Tests.csproj` suffix.
+
+Unit tests should follow these guidelines:
+
+- Add Arrange, Act and Assert comments for each test
+- Use the Async suffix on the name of all async methods.
```
cgillum: I think this is the third time I've seen this bullet point. Can we just have it in one place?

.github/copilot-instructions.md
halspang: Updated to be more broad.

cgillum (approved):

.github/copilot-instructions.md
halspang: Removed the redundant line.

.github/copilot-instructions.md
halspang: We were very serious about our async definition :) It's only in the general C# area now.

PR #506, opened 2025-11-17
Fix FunctionContext Check and Polymorphic Type Conversions in Activity Analyzer

### LINKED ISSUE #453: Analyzer Rule DURABLE2001 False Positives
The latest changes to the analyzers with `1.12.0` have introduced some false positives with rule `DURABLE2001` including:

- Flagging missing input on activities that bind to `FunctionContext` for their `ActivityTrigger` when used in Azure Functions:
```csharp
[Function(nameof(GetNumber))]
public async Task<int> GetNumber([ActivityTrigger] FunctionContext context)
{
    // Something ...
}

[Function(nameof(DoMath))]
public async Task DoMath([OrchestrationTrigger] TaskOrchestrationContext context)
{
    // error DURABLE2001: CallActivityAsync is passing the incorrect type 'none' instead of 'FunctionContext' to the activity function 'GetNumber'
    int num = await context.CallActivityAsync<int>(nameof(GetNumber));
    // ... More stuff ...
}
```
- Polymorphism does not seem to be considered properly on input. E.g. a `List<int>` argument is flagged as invalid when passed to an activity that accepts `IReadOnlyList<int>`. I know this can be a bit hairy given the extension's own JSON representation though.
    - Or perhaps all types represented by the same JSON type are acceptable alternatives? E.g. Any .NET BCL collection is a JSON array, so they can be interchanged? Although, that may rely too much on implementation details

```csharp
[Function(nameof(Sum))]
public int Sum([ActivityTrigger] IReadOnlyList<int> numbers)
    => numbers.Sum();

[Function(nameof(DoMath))]
public async Task DoMath([OrchestrationTrigger] TaskOrchestrationContext context)
{
    // error DURABLE2001: CallActivityAsync is passing the incorrect type 'List<int>' instead of 'IReadOnlyList<int>' to the activity function 'Sum'
    List<int> numbers = [1, 2, 3, 4, 5];
    int sum = await context.CallActivityAsync<int>(nameof(Sum), numbers);
    // ... More stuff ...
}
```

nytian:
Fix #453

### DIFF
--- a/.github/workflows/validate-build.yml
+++ b/.github/workflows/validate-build.yml
@@ -28,6 +28,16 @@ jobs:
       with:
         dotnet-version: '6.0.x'
 
+    - name: Setup .NET 8.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '8.0.x'
+
+    - name: Setup .NET 10.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '10.0.x'
+
     - name: Setup .NET from global.json
       uses: actions/setup-dotnet@v3
       with:
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -38,7 +38,7 @@
         "type": "func",
         "dependsOn": "build (AzureFunctionsApp)",
         "options": {
-          "cwd": "${workspaceFolder}/out/samples/bin/Debug/AzureFunctionsApp/net6.0"
+          "cwd": "${workspaceFolder}/out/samples/bin/Debug/AzureFunctionsApp/net8.0"
         },
         "command": "host start",
         "isBackground": true,
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,23 +9,22 @@
 
   <!-- Microsoft.Extensions.* Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="6.0.3" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="6.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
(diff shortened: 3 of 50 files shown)


src/Analyzers/Activities/MatchingInputOutputTypeActivityAnalyzer.cs
bachuv, line 1: Can you give more examples in the comments for each if statement check of what would make the if statement true? The if statements with comments like the one below were helpful when reading the code.

> // Special handling for collection types since ClassifyConversion doesn't always recognize
        // generic interface implementations (e.g., List<T> to IReadOnlyList<T>)

src/Analyzers/Activities/MatchingInputOutputTypeActivityAnalyzer.cs
```diff
@@ -330,6 +341,133 @@ public override void Initialize(AnalysisContext context)
...
+    static bool AreTypesCompatible(Compilation compilation, ITypeSymbol? sourceType, ITypeSymbol? targetType)
+    {
+        // Both null = compatible
+        if (sourceType == null && targetType == null)
+        {
+            return true;
+        }
+
+        // One is null, the other isn't = not compatible
+        if (sourceType == null || targetType == null)
```
andystaples: Just checking - does this logic support passing null as input to a nullable function parameter? That was also an issue with this analyzer previously

src/Analyzers/Activities/MatchingInputOutputTypeActivityAnalyzer.cs
```diff
@@ -330,6 +345,139 @@ public override void Initialize(AnalysisContext context)
...
+            if (sourceType == null && targetType != null)
+            {
+                // Check if target is a nullable reference type (NullableAnnotation.Annotated)
+                // or a nullable value type (Nullable<T>)
+                if (targetType.NullableAnnotation == NullableAnnotation.Annotated ||
+                    targetType.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T)
+                {
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
nytian: turns out this is not right

nytian:
E2E test 
-- with latest v1.11.1
image
--with fix
image

PR #507, opened 2025-11-20
Extended sessions for entities in .NET isolated

sophiatev:
This PR introduces the extended sessions feature for entities in .NET isolated. It essentially copies what was done to enable extended sessions for orchestrations in the `GrpcOrchestrationRunner` into the `GrpcEntityRunner`, and introduces a new class for shared logic between the two.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -31,7 +31,7 @@
 
   <!-- DurableTask Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.Azure.DurableTask.Core" Version="3.6.0" />
+    <PackageVersion Include="Microsoft.Azure.DurableTask.Core" Version="3.7.0" />
     <PackageVersion Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.12.1" />
   </ItemGroup>
 
--- a/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
+++ b/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
@@ -226,7 +226,8 @@ sealed partial class TaskOrchestrationContextWrapper : TaskOrchestrationContext
                     version,
                     instanceId,
                     policy.ToDurableTaskCoreRetryOptions(),
-                    input);
+                    input,
+                    options.Tags);
             }
             else if (options?.Retry?.Handler is AsyncRetryHandler handler)
             {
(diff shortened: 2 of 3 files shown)


src/Worker/Grpc/GrpcEntityRunner.cs
```diff
@@ -51,24 +52,99 @@ public static class GrpcEntityRunner
...
-        EntityBatchResult result = await entity.ExecuteOperationBatchAsync(batch);
-
+            : ActivatorUtilities.GetServiceOrCreateInstance<DurableTaskShimFactory>(services);
+
+        TaskEntity entity = factory.CreateEntity(entityName, implementation, id);
+        EntityBatchResult result = await entity.ExecuteOperationBatchAsync(batch);
+
+        if (addToExtendedSessions)
+        {
+            extendedSessions!.Set(
```
cgillum, line 139: I suggest adding a comment briefly explaining why it's safe to assume that `extendedSessions` is never null here.

src/Worker/Grpc/GrpcEntityRunner.cs
```diff
@@ -51,24 +52,99 @@ public static class GrpcEntityRunner
...
+                request.InstanceId,
+                result.EntityState,
+                new MemoryCacheEntryOptions { SlidingExpiration = TimeSpan.FromSeconds(extendedSessionIdleTimeoutInSeconds) });
+        }
+        else
+        {
+            extendedSessions?.Remove(request.InstanceId);
```
cgillum, line 146: In what situations would we successfully remove something from this cache here?

src/Worker/Grpc/GrpcOrchestrationRunner.cs
```diff
@@ -223,7 +207,7 @@ public static string LoadAndRun(
 
                 if (addToExtendedSessions && !executor.IsCompleted)
                 {
-                    extendedSessions.Set<ExtendedSessionState>(
+                    extendedSessions!.Set<ExtendedSessionState>(
```
cgillum, line 211: Same feedback about adding a comment. This is probably a good thing to always do when using the `!` operator, because it's rarely something we should be using and could therefore use an explanation.

cgillum (approved):
No blockers - just a few comment suggestions.

src/Worker/Grpc/GrpcEntityRunner.cs
sophiatev: This is to handle the situation where the caller of this method has specifically indicated that the extended session has ended (so the extended sessions cache is non-null, but `isExtendedSession` is false. We want to honor the termination of the extended session and remove the entity state from the cache here). The other path that could get us here is if extended sessions are simply not enabled, in which case the cache will be null, hence the need for the `?`

Issue #508, opened 2025-11-20
Breaking change in WaitForExternalEvent causes external events being dropped


davkrall-futureordering:
**Description**

1. WaitForExternalEvent is called with a given event name and a cancellation token
2. A cancellation request is sent using the cancellation token
3. WaitForExternalEvent is called again with the same event name as the first call
4. An event with that name is raised.

**Expected behavior**
The latest, non-cancelled WaitForExternalEvent call will consume the event

**Actual behavior**
The event will be routed to the first waiter, whose task was cancelled.

**Findings**
In the previous version of the Durable Task SDK, the task completion sources were stored in a stack (see [DurableOrchestrationContext.cs line 897](https://github.com/Azure/azure-functions-durable-extension/blob/453dd019925d3f9af9013d51947d8bbc7cb2ee00/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs#L897)), so following the above description, the new WaitForExternalEvent call's waiter would consume the event in a LIFO pattern (see [DurableOrchestrationContext.cs line 951](https://github.com/Azure/azure-functions-durable-extension/blob/453dd019925d3f9af9013d51947d8bbc7cb2ee00/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs#L951).

In the new SDK, the waiters/task completion sources are stored in a queue ([see TaskOrchestrationContextWrapper.cs line 303](http://github.com/microsoft/durabletask-dotnet/blob/b971693a11c848ed62277057cd7db635435d1d4f/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs#L303)) and the calls are processed in a FIFO pattern. Even though the task is cancelled in step 2 ([line 309](https://github.com/microsoft/durabletask-dotnet/blob/b971693a11c848ed62277057cd7db635435d1d4f/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs#L309)), the task completion source is not removed from the queue. Since the incoming events are processed in a FIFO pattern, the event is assigned to the first task completion source, which has not been removed from the queue, and is effectively lost.

This same issue has also been described in https://github.com/Azure/azure-functions-durable-extension/issues/3255.

This seems like an unintended breaking change, since WaitForExternalEvent now doesn't behave like the documentation suggests, when awaiting multiple events with the same name. I would greatly appreciate a fix for this issue, as it is blocker for migrating to the isolated worker model for my use case.

PR #509, opened 2025-11-21
Align waitforexternalevent waiter picking order to LIFO

YunchuWang:
Fix https://github.com/microsoft/durabletask-dotnet/issues/508

This pull request refactors the handling of external events in the orchestration worker to use a stack (LIFO) approach instead of a queue (FIFO). This change improves the delivery of events to the most recent waiter, making it easier for users to abandon events they no longer care about and avoids issues with cancelled waiters. The update is thoroughly validated by new and updated integration tests that confirm correct LIFO event delivery and proper handling of cancellation and buffering scenarios.

**Core event handling changes:**

* Refactored the `externalEventSources` data structure in `TaskOrchestrationContextWrapper` from a queue to a stack using a single-linked list, so the most recent waiter receives events first.
* Updated the logic in `WaitForExternalEvent` and `CompleteExternalEvent` to add new waiters to the top of the stack and deliver events in LIFO order, ensuring correct handling when waiters are cancelled or abandoned [[1]](diffhunk://#diff-a3e22a5f2875eb86f485dc42e661d7495fd717e88dbee835f76cce886f9c7df1L283-R312) [[2]](diffhunk://#diff-a3e22a5f2875eb86f485dc42e661d7495fd717e88dbee835f76cce886f9c7df1L419-R442) [[3]](diffhunk://#diff-a3e22a5f2875eb86f485dc42e661d7495fd717e88dbee835f76cce886f9c7df1L435-L440).
* Extended the `IEventSource` interface and its implementation to support stack-based linking via a new `Next` property [[1]](diffhunk://#diff-a2d337e764b7a49ab1b7a67ac9021b3f52384bade7fa242bad735b8448d2f4aaR24-R28) [[2]](diffhunk://#diff-a2d337e764b7a49ab1b7a67ac9021b3f52384bade7fa242bad735b8448d2f4aaR41-R43).

**Testing and validation:**

* Added a comprehensive test suite (`ExternalEventStackTests.cs`) that covers LIFO event delivery, cancellation scenarios, multiple waiters, and event buffering to ensure correctness of the new stack-based behavior.
* Updated existing tests (e.g., `OrchestrationPatterns.cs`) to expect LIFO delivery order, reflecting the new semantics.

### DIFF
--- a/.github/workflows/validate-build.yml
+++ b/.github/workflows/validate-build.yml
@@ -28,6 +28,16 @@ jobs:
       with:
         dotnet-version: '6.0.x'
 
+    - name: Setup .NET 8.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '8.0.x'
+
+    - name: Setup .NET 10.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '10.0.x'
+
     - name: Setup .NET from global.json
       uses: actions/setup-dotnet@v3
       with:
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -38,7 +38,7 @@
         "type": "func",
         "dependsOn": "build (AzureFunctionsApp)",
         "options": {
-          "cwd": "${workspaceFolder}/out/samples/bin/Debug/AzureFunctionsApp/net6.0"
+          "cwd": "${workspaceFolder}/out/samples/bin/Debug/AzureFunctionsApp/net8.0"
         },
         "command": "host start",
         "isBackground": true,
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -9,23 +9,22 @@
 
   <!-- Microsoft.Extensions.* Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="6.0.3" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Abstractions" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="6.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="6.0.1" Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net48'" />
-    <PackageVersion Include="Microsoft.Extensions.Hosting" Version="8.0.1" Condition="'$(TargetFramework)' == 'net8.0'"/>
-    <PackageVersion Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.2" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="6.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Options.DataAnnotations" Version="6.0.1" />
+    <PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="8.0.1" />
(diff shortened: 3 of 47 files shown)


src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -280,29 +283,33 @@ public override async Task CreateTimer(DateTime fireAt, CancellationToken cancel
...
+            // Add new waiter to the top of the stack
+            eventSource.Next = existing;
         }
-        else
+
+        // New waiter becomes the top of the stack
+        this.externalEventSources[eventName] = eventSource;
+
+        // Check the buffer to see if any events came in before the orchestrator was listening
+        if (this.externalEventBuffer.TryTake(eventName, out string? bufferedEvent))
```
philliphoff, line 309: In the original, if an event had already come in, it'd be returned without enqueuing anything.  In the new version, it's added to the top of the stack and then immediately removed from the top of the stack.  Any specific reason why?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -280,29 +283,33 @@ public override async Task CreateTimer(DateTime fireAt, CancellationToken cancel
...
         {
-            Queue<IEventSource> eventSourceQueue = new();
-            eventSourceQueue.Enqueue(eventSource);
-            this.externalEventSources.Add(eventName, eventSourceQueue);
+            // We can complete the event right away, since we already have an event's input
```
halspang, line 311: Is the `bufferedEvent` the input? How do we know that this works?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
```diff
@@ -280,29 +283,33 @@ public override async Task CreateTimer(DateTime fireAt, CancellationToken cancel
...
             }
 
-            existing.Enqueue(eventSource);
+            // Add new waiter to the top of the stack
+            eventSource.Next = existing;
         }
-        else
+
+        // New waiter becomes the top of the stack
+        this.externalEventSources[eventName] = eventSource;
```
halspang, line 306: It took me a second to figure out how this actually made a stack. Do we want to use a built-in type for this or are we getting a benefit by doing it this way?

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
YunchuWang: this aligns with in proc model. functionality wise, they are almost same, this is for making code path consistent and stack as only source of truth for waiters.  
https://github.com/Azure/azure-functions-durable-extension/blob/453dd019925d3f9af9013d51947d8bbc7cb2ee00/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs#L892-L903

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
YunchuWang: this is most lightweight way to do stack and aligns with in proc model https://github.com/Azure/azure-functions-durable-extension/blob/453dd019925d3f9af9013d51947d8bbc7cb2ee00/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs#L892-L903

benefit is maintaining consistent waiter selection order as in proc model. if waiters wait sequentially, the new waiter cant wait until old waiter receives the event or timed out, in either case, when raising the event, the new waiter should be the one who receives the event first.

src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs
YunchuWang: yes, this testcase validates it EventBuffering_EventArrivesBeforeWaiter_FirstWaiterReceivesBufferedEvent in durabletask-dotnet\test\Grpc.IntegrationTests\ExternalEventStackTests.cs

PR #510, opened 2025-11-21
Update project to target .NET 8.0 and .NET 10 and upgrade dependencies

torosent:
This pull request upgrades the project to support .NET 8.0 and .NET 10.0, modernizes dependencies across the codebase, and refactors the source generator for improved performance and maintainability. The most important changes are grouped below.

**.NET Version Upgrades and Multi-Targeting**

* Updated build and workflow configuration to use .NET 8.0 in `.github/workflows/validate-build.yml` and set the default SDK to .NET 10.0 in `global.json`.
* Changed project files for all samples and main libraries to target multiple frameworks, including net6.0, net8.0, and net10.0 where appropriate. This ensures compatibility and future-proofs the codebase.

**Dependency Updates**

* Upgraded package versions for Microsoft.Extensions.*, Azure.*, Grpc.*, Microsoft.CodeAnalysis.*, and other libraries in `Directory.Packages.props` to their latest releases compatible with .NET 8.0/10.0. This improves security, performance, and feature support.
* Updated testing and analysis dependencies (e.g., BenchmarkDotNet, xunit, FluentAssertions) to latest versions for improved reliability and compatibility.

**Build and Development Environment**

* Updated VS Code task configuration to use the net8.0 output directory for Azure Functions, aligning with the new target frameworks.

**Source Generator Refactor**

* Refactored `DurableTaskSourceGenerator` to use the incremental generator API (`IIncrementalGenerator`), improving performance and reliability of code generation. The new implementation uses Roslyn incremental providers and removes the legacy syntax receiver pattern.

**General Code Modernization**

* Added `System.Collections.Immutable` import in the generator for modern collection handling.

### DIFF
--- a/.github/workflows/validate-build.yml
+++ b/.github/workflows/validate-build.yml
@@ -23,10 +23,20 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - name: Setup .NET 6.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '6.0.x'
+
     - name: Setup .NET 8.0
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: '8.0.x'
+
+    - name: Setup .NET 10.0
+      uses: actions/setup-dotnet@v3
+      with:
+        dotnet-version: '10.0.x'
 
     - name: Setup .NET from global.json
       uses: actions/setup-dotnet@v3


test/TestHelpers/TestHelpers.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <TargetFrameworks>net6.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.0</TargetFrameworks>
```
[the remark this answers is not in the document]
torosent: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/durabletask-dotnet/pull/510#discussion_r2550976652)

halspang:
@jviau - I tried this change a while back and you flagged it as a breaking change. Do you know if we've updated/what we need to update to be able to do this?

jviau:
Dropping net6.0 TFM will introduce a breaking change to net6.0 customers. And a potentially obscure / difficult to identify one at that. Since these projects all multi-target `netstandard2.0` as well, net6.0 customers won't outright fail restore. They will be silently shifted to `netstandard2.0` and their entire transitive dependency graph as well as a result. This can be very harmful to those customers. Especially the change in Grpc dependencies: these customers will be moved from `Grpc.Net.Client` to `Grpc.Core`, which is unsupported and a much worse option that `Grpc.Net.Client`.

There are a couple options to supporting net8.0:

1. Do not remove net6.0. Just add net8.0 in addition to net6.0 (might as well add net10.0 while you are at it)
    - this is by far the easiest and least disruptive option.
    - net6.0 can then be removed in a major version rev.
2. Remove net6.0 but add targets to explicit break the customers build, telling them net6.0 is out of support and cannot be used with these packages anymore. 
    - This is fairly disruptive. Will also need to block net7.0 as well. Dotnet BCL libraries started in net10 (or is it net9?) include targets that warn when you use them in out of support TFMs. I would go with an error though here as it is more impactful than the dotnet BCL case, they are just warning it isn't tested. In our case it is an outright massive behavior change / break, but one that can manifest only at runtime. Failing at build time would be much preferred.
    
Regardless of the choice made, I recommend also adding `net10.0` as it is LTS. I would then condition the packages in `Directory.Packages.props` to be `8.x` or `10.x` based on the TFM. If you do keep net6.0, I think it is acceptable to still bump all net6.0 dependencies to 8.x. So you would have: netstandard2.0, net6.0, net8.0 --> 8.x dependencies. net10.0 --> 10.x dependencies.

torosent:
> Dropping net6.0 TFM will introduce a breaking change to net6.0 customers. And a potentially obscure / difficult to identify one at that. Since these projects all multi-target `netstandard2.0` as well, net6.0 customers won't outright fail restore. They will be silently shifted to `netstandard2.0` and their entire transitive dependency graph as well as a result. This can be very harmful to those customers. Especially the change in Grpc dependencies: these customers will be moved from `Grpc.Net.Client` to `Grpc.Core`, which is unsupported and a much worse option that `Grpc.Net.Client`.
> 
> There are a couple options to supporting net8.0:
> 
> 1. Do not remove net6.0. Just add net8.0 in addition to net6.0 (might as well add net10.0 while you are at it)
>    
>    * this is by far the easiest and least disruptive option.
>    * net6.0 can then be removed in a major version rev.
> 2. Remove net6.0 but add targets to explicit break the customers build, telling them net6.0 is out of support and cannot be used with these packages anymore.
>    
>    * This is fairly disruptive. Will also need to block net7.0 as well. Dotnet BCL libraries started in net10 (or is it net9?) include targets that warn when you use them in out of support TFMs. I would go with an error though here as it is more impactful than the dotnet BCL case, they are just warning it isn't tested. In our case it is an outright massive behavior change / break, but one that can manifest only at runtime. Failing at build time would be much preferred.
> 
> Regardless of the choice made, I recommend also adding `net10.0` as it is LTS. I would then condition the packages in `Directory.Packages.props` to be `8.x` or `10.x` based on the TFM. If you do keep net6.0, I think it is acceptable to still bump all net6.0 dependencies to 8.x. So you would have: netstandard2.0, net6.0, net8.0 --> 8.x dependencies. net10.0 --> 10.x dependencies.

Thanks Jacob. I updated to support 6.0, 8.0 and 10.0

samples/AzureFunctionsApp/AzureFunctionsApp.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
halspang, line 4: Should this also have net10?

samples/AzureFunctionsUnitTests/AzureFunctionsApp.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
halspang, line 4: Same question.

src/Generators/DurableTaskSourceGenerator.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.DurableTask.Generators
     /// Generator for DurableTask.
     /// </summary>
     [Generator]
-    public class DurableTaskSourceGenerator : ISourceGenerator
+    public class DurableTaskSourceGenerator : IIncrementalGenerator
```
halspang, line 19: Nit: Unless this is required for the `net6` -> `net8/10` upgrade, I would prefer it in another PR to keep things focused :)

halspang (commented):
Minor comments/questions. Overall looks good!

src/Generators/DurableTaskSourceGenerator.cs
torosent: it's required

samples/AzureFunctionsUnitTests/AzureFunctionsApp.Tests.csproj
torosent: The Microsoft.Azure.Functions.Worker.Extensions.DurableTask needs to be updated to net10 first

samples/AzureFunctionsApp/AzureFunctionsApp.csproj
torosent: The Microsoft.Azure.Functions.Worker.Extensions.DurableTask needs to be updated to net10 first

PR #542, opened 2025-12-08
Support dedup status when starting orchestration

YunchuWang:
Fix https://github.com/microsoft/durabletask-dotnet/issues/183
This pull request introduces support for idempotent orchestration creation by allowing orchestration instances to be deduplicated based on their runtime status. The main changes add a new `DedupeStatuses` option for orchestration creation, implement type-safe configuration via extension methods, propagate this option through the client and server layers, and add comprehensive integration tests to verify the new behavior.

### Idempotent orchestration creation

* Added a new `DedupeStatuses` property to `StartOrchestrationOptions`, enabling orchestration creation to throw an exception if an instance with the same ID exists in a specified status. This supports idempotent orchestration creation.
* Introduced `StartOrchestrationOptionsExtensions` with a type-safe `WithDedupeStatuses` method and a list of valid deduplication statuses, improving API usability and safety.

### Client and server propagation

* Updated `GrpcDurableTaskClient`, `ShimDurableTaskClient`, and `TaskHubGrpcServer` to honor the new deduplication logic, mapping dedupe statuses to protocol fields and throwing an `OrchestrationAlreadyExistsException` when appropriate. [[1]](diffhunk://#diff-1cbc56f55c712147e9731872d3513aebfdd938d74874d59ab6ddea60e75586b4R126-R166) [[2]](diffhunk://#diff-e924b510c336b66023f6252b8db94dcfbf70fb400c4954bd82d8bdac04257120L195-R214) [[3]](diffhunk://#diff-f134f8a3a86fd8307ffc0857dce7c7b7630c63468f9fe98d4c2295d92d03d458R206-R233) [[4]](diffhunk://#diff-f134f8a3a86fd8307ffc0857dce7c7b7630c63468f9fe98d4c2295d92d03d458L219-R255)
* Improved error handling by converting deduplication-related exceptions to gRPC errors with status code `AlreadyExists`.

### Integration tests

* Added multiple integration tests to verify the new deduplication behavior, including scenarios for successful replacement, exception throwing when dedupe status matches, and handling multiple dedupe statuses.

### DIFF
--- a/src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
+++ b/src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License.
 
 using System.IO.Compression;
+using System.Net;
 using System.Text;
+using Azure;
 using Azure.Core;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
@@ -117,20 +119,30 @@ public sealed class BlobPayloadStore : PayloadStore
 
         BlobClient blob = this.containerClient.GetBlobClient(name);
 
-        using BlobDownloadStreamingResult result = await blob.DownloadStreamingAsync(cancellationToken: cancellationToken);
-        Stream contentStream = result.Content;
-        bool isGzip = string.Equals(
-            result.Details.ContentEncoding, ContentEncodingGzip, StringComparison.OrdinalIgnoreCase);
-
-        if (isGzip)
+        try
         {
-            using GZipStream decompressed = new(contentStream, CompressionMode.Decompress);
-            using StreamReader reader = new(decompressed, Encoding.UTF8);
-            return await ReadToEndAsync(reader, cancellationToken);
-        }
+            using BlobDownloadStreamingResult result = await blob.DownloadStreamingAsync(cancellationToken: cancellationToken);
+            Stream contentStream = result.Content;
+            bool isGzip = string.Equals(
+                result.Details.ContentEncoding, ContentEncodingGzip, StringComparison.OrdinalIgnoreCase);
 
-        using StreamReader uncompressedReader = new(contentStream, Encoding.UTF8);
-        return await ReadToEndAsync(uncompressedReader, cancellationToken);
+            if (isGzip)
+            {
+                using GZipStream decompressed = new(contentStream, CompressionMode.Decompress);
+                using StreamReader reader = new(decompressed, Encoding.UTF8);
+                return await ReadToEndAsync(reader, cancellationToken);
+            }
+
+            using StreamReader uncompressedReader = new(contentStream, Encoding.UTF8);
+            return await ReadToEndAsync(uncompressedReader, cancellationToken);
+        }
+        catch (RequestFailedException ex) when (ex.Status == (int)HttpStatusCode.NotFound)
+        {
+            throw new InvalidOperationException(
+                $"The blob '{name}' was not found in container '{container}'. " +
+                "The payload may have been deleted or the container was never created.",
+                ex);
+        }
     }
 
     /// <inheritdoc/>


src/Client/Grpc/ProtoUtils.cs
```diff
@@ -1,15 +1,93 @@
...
+    /// Converts an OrchestrationIdReusePolicy with replaceable statuses to dedupe statuses
+    /// (statuses that should NOT be replaced).
+    /// </summary>
+    /// <param name="policy">The OrchestrationIdReusePolicy containing replaceable statuses.</param>
+    /// <returns>An array of orchestration statuses that should NOT be replaced, or null if all terminal statuses are replaceable.</returns>
+    /// <remarks>
+    /// The policy uses "replaceableStatus" - these are statuses that CAN be replaced.
+    /// dedupeStatuses are statuses that should NOT be replaced (should throw exception).
+    /// So dedupeStatuses = all terminal statuses MINUS replaceableStatus.
+    /// </remarks>
```
halspang, line 70: This comment block doesn't make a ton of sense to me. I would intuit that `DedupeStatuses` are things that can be replaced because they can be deduped. Looking at the code in the backend, it appears to be the case. 

I guess the `Dedupe` is throwing me off. Is there a reason we decided to add that layer of abstraction here instead of just giving customers the ability to set the policy/name it the same in our domain?

src/Client/Grpc/ProtoUtils.cs
```diff
@@ -1,15 +1,93 @@
...
+    /// Converts dedupe statuses (statuses that should NOT be replaced) to an OrchestrationIdReusePolicy
+    /// with replaceable statuses (statuses that CAN be replaced).
+    /// </summary>
+    /// <param name="dedupeStatuses">The orchestration statuses that should NOT be replaced. These are statuses for which an exception should be thrown if an orchestration already exists.</param>
+    /// <returns>An OrchestrationIdReusePolicy with replaceable statuses set, or null if all terminal statuses are dedupe statuses.</returns>
+    /// <remarks>
+    /// The policy uses "replaceableStatus" - these are statuses that CAN be replaced.
+    /// dedupeStatuses are statuses that should NOT be replaced.
+    /// So replaceableStatus = all terminal statuses MINUS dedupeStatuses.
+    /// </remarks>
```
halspang, line 41: Similar comment, I get `Replaceable` but the `Dedupe` is throwing me off. The default behavior is what you have for `Dedupe` which is really what I would consider a collision. To me `Dedupe` is an option I would set if I _didn't_ want it to throw. 

So, same question here, why add the new terminology? Can we map this more directly?

src/InProcessTestHost/Sidecar/Grpc/TaskHubGrpcServer.cs
```diff
@@ -216,7 +231,14 @@ await this.client.CreateTaskOrchestrationAsync(
                             : null
                     },
                     OrchestrationInstance = instance,
-                });
+                },
+                dedupeStatuses);
+        }
+        catch (OrchestrationAlreadyExistsException e)
```
halspang, line 237: This isn't exposed anywhere, right? It's just for testing? Otherwise, changing the exception signature could be a breaking change.

YunchuWang:
> I feel like this is backwards to what I would intuitively have thought. I think, looking at the actual reuse policy makes more sense from a naming perspective.
> 
> I have an orchestration with an Id `1` and a status `Completed`. I want to be able to reuse that Id, so I have a policy with the status `Completed` in it.
> 
> In the case of this PR, it seems like that's actually the opposite of what we're doing. If I have an orchestration who's Id I want to reuse that is completed, I provide the values `Failed`, `Running`, and `Terminated`? Or, I could provide an empty array?
> 
> Is this following a paradigm from another portion of Durable Functions? If not, I would strongly consider switching the naming around. Or, if you feel really strongly about it, we can have that conversation and see how others feel about it since I could be the only one who feels this way :)

https://github.com/microsoft/durabletask-dotnet/issues/183 this pr addresses this issue to support dedupstatus overloaded creation as defined in durabletask. without it, it is breaking customers as mentioned in the github issue. yes agree it feels confusing and inverse of replaceablestatuses but it mainly is needed to ensure backward compatibility from replaceablestatus to dedupstatus.

src/Client/Grpc/ProtoUtils.cs
YunchuWang: this aligns with azuremanaged sdk behavior in protoutils.

src/Client/Grpc/ProtoUtils.cs
YunchuWang: ditto.

src/InProcessTestHost/Sidecar/Grpc/TaskHubGrpcServer.cs
YunchuWang: this is for testing only. all tests pass on this.

PR #613, opened 2026-01-13
Add dependency injection support to DurableTaskTestHost

nytian:
Adds two approaches for dependency injection in `DurableTaskTestHost`:

1. **ConfigureServices** - Register services directly in test host options
2. **AddInMemoryDurableTask()** - Integrate with existing hosts (e.g., `WebApplicationFactory`)

### DIFF
--- a/src/InProcessTestHost/DurableTaskTestExtensions.cs
+++ b/src/InProcessTestHost/DurableTaskTestExtensions.cs
@@ -112,12 +112,12 @@ public class InMemoryDurableTaskOptions
 /// <summary>
 /// Internal hosted service that runs the gRPC sidecar within the user's host.
 /// </summary>
-internal sealed class InMemoryGrpcSidecarHost : IHostedService, IAsyncDisposable
+sealed class InMemoryGrpcSidecarHost : IHostedService, IAsyncDisposable
 {
-    private readonly string address;
-    private readonly InMemoryOrchestrationService orchestrationService;
-    private readonly ILoggerFactory? loggerFactory;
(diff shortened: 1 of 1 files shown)


src/InProcessTestHost/DurableTaskTestHost.cs
```diff
@@ -170,4 +180,38 @@ public class DurableTaskTestHostOptions
...
+    /// any services your orchestrations and activities need. Activities can then receive
+    /// these services via constructor injection.
+    /// </para>
+    /// <example>
+    /// <code>
+    /// var options = new DurableTaskTestHostOptions
+    /// {
+    ///     ConfigureServices = services =>
+    ///     {
+    ///         services.AddSingleton&lt;IMyService, MyService&gt;();
```
halspang: Nit: Is there a reason this got url encoded? Note the `&lt;` and `&gt;`. If it's what's required for an example doc like this, that's fine, just looks weird in a comment.

src/InProcessTestHost/README.md
```diff
@@ -1,22 +1,20 @@
 # DurableTaskTestHost - Testing Durable Orchestrations In-Process
 
-`DurableTaskTestHost` is a simple API for testing your durable task orchestrations and activities **in-process** without requiring any external backend.
-
-Supports both **class-based** and **function-based** syntax.
```
halspang, line 5: Removing this and replacing with the simpler comment implies we no longer support class based and function based. That's not the case is it? If not, why did we remove this?

nytian: added back.

PR #616, opened 2026-01-15
Introduce WorkItemFilters into worker flow

halspang:
# Summary
## What changed?
This change adds WorkItemFilters into the grpc worker. This includes builder methods to specify them and the connection into the GetWorkItems flow inside the worker processor.

## Why is this change needed?
Every orchestration, activity, and entity would go to any worker regardless of their actual ability. This change allows workers to specify what they can work on and receive those changes.

## Issues / work items
- Resolves #
- Related #

---

# Project checklist
- [ ] Release notes are not required for the next release
- [ ] Breaking change?
  - [ ] If yes:
    - Impact:
    - Migration guidance:
---

- [ ] No
- [x] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)

If AI was used:
- Tool(s):
- AI-assisted areas/files: Tests

- [x] I verified referenced APIs/types exist and are correct
- [x] I reviewed edge cases/failure paths (timeouts, retries, cancellation, exceptions)
- [x] I reviewed concurrency/async behavior
- [x] I checked for unintended breaking or behavior changes

---

# Testing
## Automated tests
- Result: Passed

- Environment (OS, .NET version, components): Windows/Aspire, .NET, Custom test bed
- Steps + observed results:
  1. Hookup a variety of workers with different filters specified (including auto-gen).
  2. Run tests on them and ensure the appropriate work lands on the appropriate worker.

- Evidence (optional):
I didn't take screenshots but I can go back if required.
---

# Notes for reviewers
- N/A

### DIFF
--- a/src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
+++ b/src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
@@ -35,18 +35,25 @@ public class DurableTaskWorkerWorkItemFilters
     /// <returns>A new instance of <see cref="DurableTaskWorkerWorkItemFilters"/> constructed from the provided registry.</returns>
     internal static DurableTaskWorkerWorkItemFilters FromDurableTaskRegistry(DurableTaskRegistry registry, DurableTaskWorkerOptions? workerOptions)
     {
-        // TODO: Support multiple versions per orchestration/activity. For now, grab the worker version from the options.
+        // TODO: Support multiple versions per orchestration/activity.
+        // For now, fetch the version based on the versioning match strategy if defined. If undefined, default to null (all versions match).
+        IReadOnlyList<string> versions = [];
+        if (workerOptions?.Versioning?.MatchStrategy == DurableTaskWorkerOptions.VersionMatchStrategy.Strict)
+        {
+            versions = [workerOptions.Versioning.Version];
+        }
+
         return new DurableTaskWorkerWorkItemFilters
         {
             Orchestrations = registry.Orchestrators.Select(orchestration => new OrchestrationFilter
             {
                 Name = orchestration.Key,
-                Versions = workerOptions?.Versioning != null ? [workerOptions.Versioning.Version] : [],
+                Versions = versions,
             }).ToList(),
             Activities = registry.Activities.Select(activity => new ActivityFilter
             {
                 Name = activity.Key,
-                Versions = workerOptions?.Versioning != null ? [workerOptions.Versioning.Version] : [],
+                Versions = versions,
             }).ToList(),
             Entities = registry.Entities.Select(entity => new EntityFilter
             {
--- a/test/Worker/Core.Tests/DependencyInjection/UseWorkItemFiltersTests.cs
+++ b/test/Worker/Core.Tests/DependencyInjection/UseWorkItemFiltersTests.cs
@@ -111,7 +111,7 @@ public class UseWorkItemFiltersTests
     }
 
     [Fact]
-    public void WorkItemFilters_DefaultWithVersioning_WhenNoExplicitFiltersConfigured()
+    public void WorkItemFilters_DefaultNullWithVersioningCurrentOrOlder_WhenNoExplicitFiltersConfigured()
     {
         // Arrange
         ServiceCollection services = new();
@@ -126,7 +126,74 @@ public class UseWorkItemFiltersTests
             {
                 options.Versioning = new DurableTaskWorkerOptions.VersioningOptions
                 {
-                    Version = "1.0"
+                    Version = "1.0",
+                    MatchStrategy = DurableTaskWorkerOptions.VersionMatchStrategy.CurrentOrOlder,
+                };
+            });
+        });
+
+        // Act
+        ServiceProvider provider = services.BuildServiceProvider();
+        IOptionsMonitor<DurableTaskWorkerWorkItemFilters> filtersMonitor =
+            provider.GetRequiredService<IOptionsMonitor<DurableTaskWorkerWorkItemFilters>>();
+        DurableTaskWorkerWorkItemFilters actual = filtersMonitor.Get("test");
+
+        // Assert
+        actual.Orchestrations.Should().ContainSingle(o => o.Name == nameof(TestOrchestrator) && o.Versions.Count == 0);
+        actual.Activities.Should().ContainSingle(a => a.Name == nameof(TestActivity) && a.Versions.Count == 0);
+    }
+
+    [Fact]
+    public void WorkItemFilters_DefaultNullWithVersioningNone_WhenNoExplicitFiltersConfigured()
+    {
+        // Arrange
+        ServiceCollection services = new();
+        services.AddDurableTaskWorker("test", builder =>
+        {
+            builder.AddTasks(registry =>
+            {
+                registry.AddOrchestrator<TestOrchestrator>();
+                registry.AddActivity<TestActivity>();
+            });
+            builder.Configure(options =>
+            {
+                options.Versioning = new DurableTaskWorkerOptions.VersioningOptions
+                {
+                    Version = "1.0",
+                    MatchStrategy = DurableTaskWorkerOptions.VersionMatchStrategy.None,
+                };
+            });
+        });
+
+        // Act
+        ServiceProvider provider = services.BuildServiceProvider();
+        IOptionsMonitor<DurableTaskWorkerWorkItemFilters> filtersMonitor =
+            provider.GetRequiredService<IOptionsMonitor<DurableTaskWorkerWorkItemFilters>>();
+        DurableTaskWorkerWorkItemFilters actual = filtersMonitor.Get("test");
+
+        // Assert
+        actual.Orchestrations.Should().ContainSingle(o => o.Name == nameof(TestOrchestrator) && o.Versions.Count == 0);
+        actual.Activities.Should().ContainSingle(a => a.Name == nameof(TestActivity) && a.Versions.Count == 0);
+    }
+
+    [Fact]
+    public void WorkItemFilters_DefaultVersionWithVersioningStrict_WhenNoExplicitFiltersConfigured()
+    {
+        // Arrange
+        ServiceCollection services = new();
+        services.AddDurableTaskWorker("test", builder =>
+        {
+            builder.AddTasks(registry =>
+            {
+                registry.AddOrchestrator<TestOrchestrator>();
+                registry.AddActivity<TestActivity>();
+            });
+            builder.Configure(options =>
+            {
+                options.Versioning = new DurableTaskWorkerOptions.VersioningOptions
+                {
+                    Version = "1.0",
+                    MatchStrategy = DurableTaskWorkerOptions.VersionMatchStrategy.Strict,
                 };
             });
         });
@@ -291,7 +358,7 @@ public class UseWorkItemFiltersTests
         worker1Filters.Should().NotBeSameAs(worker2Filters);
     }
 
-    class TestOrchestrator : TaskOrchestrator<object, object>
+    sealed class TestOrchestrator : TaskOrchestrator<object, object>
     {
         public override Task<object> RunAsync(TaskOrchestrationContext context, object input)
         {
@@ -299,7 +366,7 @@ public class UseWorkItemFiltersTests
         }
     }
 
-    class TestActivity : TaskActivity<object, object>
+    sealed class TestActivity : TaskActivity<object, object>
     {
         public override Task<object> RunAsync(TaskActivityContext context, object input)
         {
@@ -307,7 +374,7 @@ public class UseWorkItemFiltersTests
         }
     }
 
-    class TestEntity : TaskEntity<object>
+    sealed class TestEntity : TaskEntity<object>
     {
     }
 }
--- a/test/Worker/Grpc.Tests/DurableTaskWorkerWorkItemFiltersExtensionTests.cs
+++ b/test/Worker/Grpc.Tests/DurableTaskWorkerWorkItemFiltersExtensionTests.cs
@@ -5,7 +5,6 @@ using Microsoft.DurableTask.Worker.Grpc.Internal;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
 using P = Microsoft.DurableTask.Protobuf;
 
 namespace Microsoft.DurableTask.Worker.Grpc.Tests;
@@ -407,7 +406,7 @@ public class DurableTaskWorkerWorkItemFiltersExtensionTests
         filters.Entities.Should().BeEmpty();
     }
 
-    class TestOrchestrator : TaskOrchestrator<object, object>
+    sealed class TestOrchestrator : TaskOrchestrator<object, object>
     {
         public override Task<object> RunAsync(TaskOrchestrationContext context, object input)
         {
@@ -415,7 +414,7 @@ public class DurableTaskWorkerWorkItemFiltersExtensionTests
         }
     }
 
-    class TestActivity : TaskActivity<object, object>
+    sealed class TestActivity : TaskActivity<object, object>
     {
         public override Task<object> RunAsync(TaskActivityContext context, object input)
         {


src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
```diff
@@ -137,4 +137,32 @@ public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTask
...
+            builder.Services.AddSingleton(workItemFilters);
+        }
+        else
+        {
+            // Auto-generate the filters from registered orchestrations, activities, and entitites.
+            builder.Services.AddSingleton(provider =>
+            {
+                DurableTaskRegistry registry = provider.GetRequiredService<IOptionsMonitor<DurableTaskRegistry>>().Get(builder.Name);
+                DurableTaskWorkerOptions? options = provider.GetOptions<DurableTaskWorkerOptions>(builder.Name);
+                return new DurableTaskWorkerWorkItemFilters(registry, options);
```
cgillum: I'm curious what the difference is between the following two options:
 
1. I build a worker without configuring filtering (basically, what people are doing today)
2. I build a worker with `.UseWorkItemFilters()` (no arguments)

Intuitively I wouldn't expect there to be a scenario where I'd use filters without any configuration, so I need a bit of help understanding the use case for option 2.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        this.Activities = activityActions;
+        List<EntityFilter> entityActions = new();
+        foreach (var entity in registry.Entities)
+        {
+            entityActions.Add(new EntityFilter
+            {
+                // Entity names are normalized to lowercase in the backend.
+                Name = entity.Key.ToString().ToLowerInvariant(),
```
cgillum: If the backend is already normalizing these, do we need to lower-case them here? I worry a bit about operations like this since I've frequently seen them lead to subtle bugs. My personal preference is that we instead rely on case-insensitive string comparisons everywhere.

halspang: I'll actually make sure I move this logic to the backend only. Right now, the backend is actually what forces the normalization so we can at least keep it all in one place/use case-insensitive comparisons where appropriate.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
halspang: That auto-generates the filters for you based on what is registered in the `DurableTaskRegistry`. Would you prefer I had a different method like `UseAutoGeneratedWorkItemFitlers()`? I combined them into one because I didn't want someone to be able to specify filters and auto-gen at the same time.

src/Worker/Grpc/Internal/DurableTaskWorkerWorkItemFiltersExtension.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        foreach (var entityFilter in workItemFilter.Entities)
+        {
+            var grpcEntityAction = new P.EntityFilter
+            {
+                Name = entityFilter.Name,
+            };
+            grpcWorkItemFilters.Entities.Add(grpcEntityAction);
+        }
```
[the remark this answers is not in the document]
halspang: I'm going to keep it this way for consistency with Orchestrations and Activities. Entities may also be versioned in the future, so it matches better.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
```diff
@@ -137,4 +137,32 @@ public static IDurableTaskWorkerBuilder UseOrchestrationFilter(this IDurableTask
...
+        }
+        else
+        {
+            // Auto-generate the filters from registered orchestrations, activities, and entities.
+            builder.Services.AddSingleton(provider =>
+            {
+                DurableTaskRegistry registry = provider.GetRequiredService<IOptionsMonitor<DurableTaskRegistry>>().Get(builder.Name);
+                DurableTaskWorkerOptions? options = provider.GetOptions<DurableTaskWorkerOptions>(builder.Name);
+                return DurableTaskWorkerWorkItemFilters.FromDurableTaskRegistry(registry, options);
+            });
```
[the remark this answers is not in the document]
halspang: Updated the logic to handle named builders and added a test to verify.

src/Worker/Grpc/Internal/DurableTaskWorkerWorkItemFiltersExtension.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        foreach (var entityFilter in workItemFilter.Entities)
+        {
+            var grpcEntityFilter = new P.EntityFilter
+            {
+                Name = entityFilter.Name,
+            };
+            grpcWorkItemFilters.Entities.Add(grpcEntityFilter);
+        }
```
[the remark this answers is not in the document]
halspang: This is on purpose to match the lines above this and because entities may later have more fields.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
cgillum: I see, so this is to prevent the worker from fetching work items that it doesn't know how to process. Is there any reason why someone *wouldn't* want this behavior?

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
cgillum: Friendly reminder on this. I *think* we agreed that this use of `ToLowerInvariant()` can be removed from the SDK.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        /// <summary>
+        /// Gets or initializes the name of the orchestration to filter.
+        /// </summary>
+        public string Name { get; init; }
+
+        /// <summary>
+        /// Gets or initializes the versions of the orchestration to filter.
+        /// </summary>
+        public IReadOnlyList<string> Versions { get; init; }
```
cgillum: Let's either make this property nullable or set a default (empty) value:

```suggestion
        public IReadOnlyList<string> Versions { get; init; } = [];
```

Same comment for `ActivityFilter.Versions`.

src/Worker/Grpc/Internal/DurableTaskWorkerWorkItemFiltersExtension.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        Check.NotNull(workItemFilter);
+        var grpcWorkItemFilters = new P.WorkItemFilters();
+        foreach (var orchestrationFilter in workItemFilter.Orchestrations)
+        {
+            var grpcOrchestrationFilter = new P.OrchestrationFilter
+            {
+                Name = orchestrationFilter.Name,
+            };
+            grpcOrchestrationFilter.Versions.AddRange(orchestrationFilter.Versions);
```
cgillum, line 28: Related to my other comment, we need to ensure that `orchestrationFilter.Versions` is either never null or we need to do a null check here.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <returns>A new instance of <see cref="DurableTaskWorkerWorkItemFilters"/> constructed from the provided registry.</returns>
+    internal static DurableTaskWorkerWorkItemFilters FromDurableTaskRegistry(DurableTaskRegistry registry, DurableTaskWorkerOptions? workerOptions)
+    {
+        // TODO: Support multiple versions per orchestration/activity. For now, grab the worker version from the options.
+        return new DurableTaskWorkerWorkItemFilters
+        {
+            Orchestrations = registry.Orchestrators.Select(orchestration => new OrchestrationFilter
+            {
+                Name = orchestration.Key,
+                Versions = workerOptions?.Versioning != null ? [workerOptions.Versioning.DefaultVersion] : [],
```
cgillum: `DefaultVersion` is intended to be used for the creation of new orchestrations, but here it looks like we're now also using it for filtering. I worry about potential unintended consequences of this dual meaning for `DefaultVersion`. Wouldn't it be safer to simply leave this as null/empty if no explicit filters are configured? Otherwise, how does in-place upgrade work?

Same comment for activity registrations.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
halspang: The only real reason someone might not want this is if they have a wide spread of versions and use `context.Version` in their orchestrations so any worker can work on any version. Or, if you wanted to make sure any unknown version would get to the worker and be set to `Failed`. However, both of these things are possible with the filter as it has turned out since you can now specify multiple versions if you do it by hand and the unrecognized versions still go to the default queue.

When I first wrote this, the behavior was slightly different in that we didn't have the baseline queue where "unregistered" work went. Now that we have it, it's a lot safer to have this be the default behavior. Do you think we should go default?

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
halspang: We did agree to that, and I thought I had made the change but it appears it didn't make it into the commit itself. Updated!

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
halspang: Updated to have a default value which also included adding in a constructor because apparently it necessitates that.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
halspang: You are correct that I used `DefaultVersion` wrong here. I meant to use `Versioning.Version` which is actually what the worker uses. 

In regards to in-place upgrades, the old worker gets deprovisioned which removes the old queue. The new worker with the new version gets provisioned and registers the new queue. The old versions end up on the default queue as they should eventually stop getting created. In the case of multiple versions, you'd want to manually specify the filters to have both versions.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
cgillum: I'm thinking of the case where someone does a rolling upgrade of their fleet, migrating from one v1 to v2. In this example, the new v2 workers should be able to handle both v1 and v2 orchestrations (a case that I expect will be common). Per my understanding, a "default filter" behavior will work because when the v1 workers go away, any in-flight work will go to the "default" queue and get picked up by v2 workers. The downside, however, is that during the upgrade, the worker capacity for v1 orchestrations decreases as v1 workers go away since v2 workers will never process v1 orchestrations as long as ANY v1 worker exists (based on my understanding of the backend changes). This is different compared to the "no filters" behavior where both v1 and v2 workers can process v1 orchestrations, so there's no capacity issue in the case where there's a large backlog of v1 orchestrations to process during the upgrade.

Am I thinking about this behavior correctly? If so, does that impact your thinking on whether this is a good/safe default behavior?

halspang: Yes, that is what would happen! I would still think that using the version in the default filter is the correct behavior though. Strictly because if the customer has opted-in to the versioning feature, I would expect that it affects all areas of that feature. If I had already specified it in 1 place, having to double specify the filter is kind of clunky in my opinion. 

I would also argue that that drain pattern is correct because if you are deploying v2 workers, you are also most likely starting v2 orchestrations. And, once you deprovision all the workers, the v1 orchestrations can also start flowing to the v2 workers through the default queue. 

The just named filter that you are describing could also be easily manually implemented by the user by defining their own filter. If we're tending towards removing user friction, this is the easier "manual" filter to define as opposed to the feature one.

Thoughts?

philliphoff:
Separate from the discussion about version filters, I'd advocate for the SDK to generally enable filtering by default as it allows tooling to immediately identify two classes of common DT mistakes:

 - Scheduling orchestrations not implemented by any connected worker: Azure Functions will immediately complete such orchestrations with a (very Functions specific) failure but raw SDK use will simply leave the orchestration in the pending state; filters will allow tools to correlate an orchestration and worker and directly indicate why it failed and/or remains pending
 - Having multiple workers connected with disparate orchestration implementations: a common issue is forgetting to disconnect one worker before starting another (e.g. trying various DT samples and using a common scheduler connection string); filters will allow likely allow applications to proceed as work items are routed to a matching worker

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,104 @@
...
+            }).ToList(),
+            Activities = registry.Activities.Select(activity => new ActivityFilter
+            {
+                Name = activity.Key,
+                Versions = workerOptions?.Versioning != null ? [workerOptions.Versioning.Version] : [],
+            }).ToList(),
+            Entities = registry.Entities.Select(entity => new EntityFilter
+            {
+                // Entity names are normalized to lowercase in the backend.
+                Name = entity.Key.ToString(),
```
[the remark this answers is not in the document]
halspang: We don't do the invariant here _because_ it is done in the backend.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
cgillum, line 133: Are these "obsolete" experimental APIs still relevant?

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// <returns>A new instance of <see cref="DurableTaskWorkerWorkItemFilters"/> constructed from the provided registry.</returns>
+    internal static DurableTaskWorkerWorkItemFilters FromDurableTaskRegistry(DurableTaskRegistry registry, DurableTaskWorkerOptions? workerOptions)
+    {
+        // TODO: Support multiple versions per orchestration/activity. For now, grab the worker version from the options.
+        return new DurableTaskWorkerWorkItemFilters
+        {
+            Orchestrations = registry.Orchestrators.Select(orchestration => new OrchestrationFilter
+            {
+                Name = orchestration.Key,
+                Versions = workerOptions?.Versioning != null ? [workerOptions.Versioning.Version] : [],
```
cgillum: I'm still a bit confused regarding the use of `workerOptions.Versioning.Version` here. My understanding is that we should be registering the name/version information of all registered orchestrations and activities, but this code doesn't really consider the registered version numbers at all. Rather, it's considering the potentially unrelated worker version. I need help understanding if this is correct semantic behavior. I'm also wondering what will happen when a user rolls out two versions of an orchestration or activity. Will that scenario simply not work?

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        /// <summary>
+        /// Gets or initializes the name of the orchestration to filter.
+        /// </summary>
+        public string Name { get; init; } = name;
+
+        /// <summary>
+        /// Gets or initializes the versions of the orchestration to filter.
+        /// </summary>
+        public IReadOnlyList<string> Versions { get; init; } = versions ?? [];
```
cgillum, line 89: I think it's still technically possible for `Versions` to be null. Since this is a struct, it has an implicit constructor with no parameters. If that version of a filter gets created, then both `Name` and `Versions` will be null. You can fix this by adding a default constructor explicitly and setting these properties to non-null values. It might be worth doing this so we don't run into unexpected NREs (like in DurableTaskWorkerWorkItemFiltersExtension.ToGrpcWorkItemFilters).

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
halspang: They are an entirely different feature with an unfortunately similar name. I can sync with the folks who wanted it to see if they are still using it.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
halspang: We register what the worker can work on. The version that the worker can work on is `workerOptions.Versioning.Version`. So currently, the registry does not have any concept of a version. When we do support multiple versions in the registry, we will need to update this code. There is a `TODO` at the top of this block stating that.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
halspang: Ah, I didn't realize this worked that way. I thought the `??` was always evaluated. Updated to use the constructor approach you recommended.

src/Worker/Core/DurableTaskWorkerWorkItemFilters.cs
cgillum: I still have concerns here. Let's sync offline.

src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
cgillum: > I would also argue that that drain pattern is correct because if you are deploying v2 workers, you are also most likely starting v2 orchestrations. And, once you deprovision all the workers, the v1 orchestrations can also start flowing to the v2 workers through the default queue.

I don't think this is necessarily true. Architectures where clients are decoupled from workers are likely going to deploy the V2 workers first before updating clients to V2. In that case, the flow of incoming V1 orchestrations remains the same, but the capacity to handle them decreases. I can see this being a big usability problem, making it impractical for those customers to do rolling upgrades in some cases.

I think the fundamental issue is that there's no equivalent to the `VersionMatchStrategy.CurrentOrOlder` in the work item filtering config - thus there's an opportunity for mismatched customer expectations. If filters supported `CurrentOrOlder` matching, then I think the remaining issues in my mind would be solved.

Note that this is related to our other unresolved comment thread.

halspang: After chatting, decided to go with a combined approach here to hit the most user scenarios. Default version is now null _unless_ the version match strategy is set to strict, then it just uses that value.

cgillum (approved):

PR #622, opened 2026-01-21
Changing the default dedupe statuses behavior

sophiatev:
# Summary
Currently, only terminal statuses are reusable when creating a new orchestration with a given instance ID. This is enforced when creating the [`OrchestrationIdReusePolicy`](https://github.com/microsoft/durabletask-dotnet/blob/e944e3c6f8d0e4dd2cb47d1efd7a310bb284e017/src/Client/Grpc/GrpcDurableTaskClient.cs#L146), which ensures that only terminal statuses are considered "reusable" (minus whatever statuses the user includes in the dedupe statuses passed to the creation call).  This PR allows all statuses to be reusable, with the assumption that the server-side implementation will terminate an existing orchestration if it is running, pending, or suspended (in a non-terminal status) before creating a new one. This is a breaking change for which all server-side implementations will need to be updated accordingly:

1. This [PR](https://github.com/Azure/azure-functions-durable-extension/pull/3316) addresses the WebJobs implementation for the DF use-case. 
2. DTS will need to be updated for the standalone SDK use-case to add the termination logic. However, the current backend implementation already does not allow creating a new orchestration if one exists in a non-terminal status, so changing the client-side implementation will not "break" anything until then.
3. Dapr sidecar, which forks this repo so it should be fine? 

---

# Project checklist
- [x] Release notes are not required for the next release
- [ ] Breaking change?
  - [x] If yes:
    - Impact: See summary
    - Migration guidance:
---

- [x] No
- [ ] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)

If AI was used:
- Tool(s):
- AI-assisted areas/files:

- [ ] I understand the code and can explain it
- [ ] I verified referenced APIs/types exist and are correct
- [ ] I reviewed edge cases/failure paths (timeouts, retries, cancellation, exceptions)
- [ ] I reviewed concurrency/async behavior
- [ ] I checked for unintended breaking or behavior changes

# Notes for reviewers
- N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+## 1.21.0
+- Introduce WorkItemFilters into worker flow by halspang ([#616](https://github.com/microsoft/durabletask-dotnet/pull/616))
+- Fix Analyzers treating passed in variable argument name as null by wangbill ([#640](https://github.com/microsoft/durabletask-dotnet/pull/640))
+- Move DURABLE0009/0010 from Unshipped to Shipped for v0.3.0 by cgillum ([#641](https://github.com/microsoft/durabletask-dotnet/pull/641))
+
 ## 1.20.1
 - Fix GrpcChannel handle leak in AzureManaged backendby nytian ([#629](https://github.com/microsoft/durabletask-dotnet/pull/629))
 
--- a/eng/targets/Release.props
+++ b/eng/targets/Release.props
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <VersionPrefix>1.20.1</VersionPrefix>
+    <VersionPrefix>1.21.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
   </PropertyGroup>
 
--- a/src/Grpc/orchestrator_service.proto
+++ b/src/Grpc/orchestrator_service.proto
@@ -822,6 +822,7 @@ message GetWorkItemsRequest {
     int32 maxConcurrentEntityWorkItems = 3;
 
     repeated WorkerCapability capabilities = 10;
+    WorkItemFilters workItemFilters = 11;
 }
 
 enum WorkerCapability {
@@ -842,6 +843,26 @@ enum WorkerCapability {
     // Work items may contain URI references instead of inline data, and the worker must fetch them.
     // This avoids message size limits and reduces network overhead.
     WORKER_CAPABILITY_LARGE_PAYLOADS = 3;
+}
+
+message WorkItemFilters {
+    repeated OrchestrationFilter orchestrations = 1;
+    repeated ActivityFilter activities = 2;
+    repeated EntityFilter entities = 3;
+}
+
+message OrchestrationFilter {
+    string name = 1;
+    repeated string versions = 2;
+}
+
+message ActivityFilter {
+    string name = 1;
+    repeated string versions = 2;
+}
+
+message EntityFilter {
+    string name = 1;
 }
 
 message WorkItem {
--- a/src/Grpc/versions.txt
+++ b/src/Grpc/versions.txt
@@ -1,2 +1,2 @@
-# The following files were downloaded from branch main at 2026-01-13 00:01:21 UTC
-https://raw.githubusercontent.com/microsoft/durabletask-protobuf/026329c53fe6363985655857b9ca848ec7238bd2/protos/orchestrator_service.proto
+# The following files were downloaded from branch main at 2026-02-24 00:01:28 UTC
+https://raw.githubusercontent.com/microsoft/durabletask-protobuf/1caadbd7ecfdf5f2309acbeac28a3e36d16aa156/protos/orchestrator_service.proto
--- a/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/DurableTaskWorkerBuilderExtensions.cs
@@ -137,4 +137,39 @@ public static class DurableTaskWorkerBuilderExtensions
         builder.Services.AddSingleton(filter);
         return builder;
     }
+
+    /// <summary>
+    /// Adds <see cref="DurableTaskWorkerWorkItemFilters"/> to the specified <see cref="IDurableTaskWorkerBuilder"/>.
+    /// </summary>
+    /// <param name="builder">The builder to set the builder target for.</param>
+    /// <param name="workItemFilters">The instance of a <see cref="DurableTaskWorkerWorkItemFilters"/> to use.
+    /// If <c>null</c>, the auto-generated default filters will be cleared.</param>
+    /// <returns>The same <see cref="IDurableTaskWorkerBuilder"/> instance, allowing for method chaining.</returns>
+    /// <remarks>Work item filters are auto-generated from the registry by default.
+    /// Use this method with explicit filters to override the defaults, or with <c>null</c> to opt out of filtering entirely.</remarks>
+    public static IDurableTaskWorkerBuilder UseWorkItemFilters(this IDurableTaskWorkerBuilder builder, DurableTaskWorkerWorkItemFilters? workItemFilters)
+    {
+        Check.NotNull(builder);
+
+        // Use PostConfigure to ensure provided filters override the auto-generated defaults.
+        // When null is passed, the filters are cleared to opt out of filtering entirely.
+        builder.Services.AddOptions<DurableTaskWorkerWorkItemFilters>(builder.Name)
+            .PostConfigure(opts =>
+            {
+                if (workItemFilters is null)
+                {
+                    opts.Orchestrations = [];
+                    opts.Activities = [];
+                    opts.Entities = [];
+                }
+                else
+                {
+                    opts.Orchestrations = workItemFilters.Orchestrations;
+                    opts.Activities = workItemFilters.Activities;
+                    opts.Entities = workItemFilters.Entities;
+                }
+            });
+
+        return builder;
+    }
 }
--- a/src/Worker/Core/DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Worker/Core/DependencyInjection/ServiceCollectionExtensions.cs
@@ -86,6 +86,21 @@ public static class ServiceCollectionExtensions
                 }
             });
 
+        // Auto-generate work item filters from the registry by default.
+        // Users can override these by calling UseWorkItemFilters(customFilters) on the builder.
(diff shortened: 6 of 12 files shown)


src/Client/Core/DurableTaskClient.cs
```diff
@@ -73,10 +73,11 @@ public virtual Task<string> ScheduleNewOrchestrationInstanceAsync(
     /// <remarks>
     /// <para>All orchestrations must have a unique instance ID. You can provide an instance ID using the
     /// <paramref name="options"/> parameter or you can omit this and a random instance ID will be
-    /// generated for you automatically. If an orchestration with the specified instance ID already exists and is in a
-    /// non-terminal state (Pending, Running, etc.), then this operation may fail silently. However, if an orchestration
-    /// instance with this ID already exists in a terminal state (Completed, Terminated, Failed, etc.) then the instance
-    /// may be recreated automatically, depending on the configuration of the backend instance store.
+    /// generated for you automatically. If an orchestration with the specified instance ID already exists and its status
+    /// is not in the <see cref="StartOrchestrationOptions.DedupeStatuses"/> field of <paramref name="options"/>, then
+    /// a new orchestration may be recreated automatically, depending on the configuration of the backend instance store.
```
sophiatev, line 79: What does "depending on the configuration of the backend instance store." in the existing method comment mean? Do not all backends support dedupe statuses?

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -139,17 +136,11 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
...
+                })];
 
-            if (policy != null)
-            {
-                request.OrchestrationIdReusePolicy = policy;
-            }
-        }
+        // Convert dedupe statuses to protobuf statuses and create reuse policy
+        IEnumerable<P.OrchestrationStatus> dedupeStatusesProto = dedupeStatuses.Select(s => s.ToGrpcStatus());
+        request.OrchestrationIdReusePolicy = ProtoUtils.ConvertDedupeStatusesToReusePolicy(dedupeStatusesProto);
```
sophiatev: Do we want to add logic to wrap the "AlreadyExists" `RpcException` thrown if an orchestration already exists with a status in dedupe statuses?

src/Client/Grpc/ProtoUtils.cs
```diff
@@ -13,18 +13,22 @@ namespace Microsoft.DurableTask.Client.Grpc;
...
 {
     /// <summary>
-    /// Gets the terminal orchestration statuses that are commonly used for deduplication.
+    /// Gets an array of all orchestration statuses.
     /// These are the statuses that can be used in OrchestrationIdReusePolicy.
     /// </summary>
-    /// <returns>An immutable array of terminal orchestration statuses.</returns>
-    public static ImmutableArray<P.OrchestrationStatus> GetTerminalStatuses()
+    /// <returns>An immutable array of all orchestration statuses.</returns>
+    public static ImmutableArray<P.OrchestrationStatus> GetAllStatuses()
```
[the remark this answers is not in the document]
sophiatev: This is a good point. Is this something we need to worry about?

src/Client/Grpc/ProtoUtils.cs
```diff
@@ -13,18 +13,21 @@ namespace Microsoft.DurableTask.Client.Grpc;
...
     {
 #pragma warning disable CS0618 // Type or member is obsolete - Canceled is intentionally included for compatibility
```
sophiatev, line 22: Compatibility with what?

src/Client/Core/DurableTaskClient.cs
cgillum: I need to be more specific when I write these comments because I often forget, but yeah, I think this meant that some backends don't support automatically re-creating a terminal instance. I don't recall what backends that would be though.

src/Client/Grpc/ProtoUtils.cs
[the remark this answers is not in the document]
cgillum: I don't know why anyone besides us would take a dependency on these APIs, but is there any particular reason we need to change them? I haven't yet looked closely to understand why they were changed.

src/Client/Grpc/GrpcDurableTaskClient.cs
```diff
@@ -124,9 +124,7 @@ public override async Task<string> ScheduleNewOrchestrationInstanceAsync(
         }
 
         // Set orchestration ID reuse policy for deduplication support
-        // Note: This requires the protobuf to support OrchestrationIdReusePolicy field
-        // If the protobuf doesn't support it yet, this will need to be updated when the protobuf is updated
-        if (options?.DedupeStatuses != null && options.DedupeStatuses.Count > 0)
+        if (options?.DedupeStatuses != null)
```
cgillum, line 129: What's the reason for removing the `options.DedupeStatuses.Count` check? Is that a behavior change?

src/Client/Grpc/ProtoUtils.cs
cgillum: Durable Task Framework defines a `Canceled` status in its public API (though it's never been implemented).

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -385,4 +387,53 @@ Task SendInstanceMessageAsync(string instanceId, HistoryEvent @event, Cancellati
 
         return this.Client.SendTaskOrchestrationMessageAsync(message);
     }
+
+    async Task TerminateTaskOrchestrationWithReusableRunningStatusAndWaitAsync(
+            string instanceId,
+            OrchestrationStatus[]? dedupeStatuses,
+            CancellationToken cancellation)
```
cgillum: nit: please clean up the whitespace in this method.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -385,4 +387,53 @@ Task SendInstanceMessageAsync(string instanceId, HistoryEvent @event, Cancellati
...
+                        $"{(dedupeStatuses == null ? "[]" : string.Join(", ", dedupeStatuses))}, do not contain the orchestration's status, " +
+                        $"the orchestration has been terminated and a new instance with the same instance ID will be created.";
+
+                    await this.TerminateInstanceAsync(instanceId, terminationReason, cancellation);
+
+                    while (metadata != null && !metadata.IsCompleted)
+                    {
+                        cancellation.ThrowIfCancellationRequested();
+                        await Task.Delay(TimeSpan.FromSeconds(1), cancellation);
+                        metadata = await this.GetInstancesAsync(instanceId, getInputsAndOutputs: false, cancellation);
```
cgillum: Rather than polling the status every second, shouldn't we be using the WaitForCompleted API?

src/Client/Grpc/ProtoUtils.cs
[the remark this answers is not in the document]
sophiatev: This was being relied on as a list of all statuses that can be reusable. Since we're changing the behavior to now allow for running statuses to be reusable, I changed the name

src/Client/Grpc/GrpcDurableTaskClient.cs
sophiatev: This is indeed a behavior change. My thinking is:

1. If the user didn't specify dedupe statuses at all, we should default to whatever the backend implementation does
2. If the user specifically set the dedupe statuses to an empty array, we should take that to mean all statuses are reusable

Previously, I think both situations would default to whatever the backend does

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -169,6 +170,8 @@ public override Task RaiseEventAsync(
     }
 
     /// <inheritdoc/>
+    /// This implementation treats a null <see cref="StartOrchestrationOptions.DedupeStatuses"/> field as all statuses
+    /// being reusable.
```
cgillum: The use of `///` is invalid here since this isn't really part of any XML tags (`<inheritdoc/>` above is closed). Can you just do `//` instead or are you trying to embed this into the official documentation that gets generated from the XML comments?

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -287,6 +291,9 @@ public override async Task<OrchestrationMetadata> WaitForInstanceStartAsync(
     }
 
     /// <inheritdoc/>
+    /// This implementation will terminate an existing non-terminal instance if <paramref name="restartWithNewInstanceId"/>
+    /// is <c>false</c> and wait for the existing instance to enter a terminal state before restarting it, until
+    /// <paramref name="cancellation"/> is cancelled.
```
cgillum: Same comment about misplaced documentation comments.

src/Client/Core/StartOrchestrationOptionsExtensions.cs
```diff
@@ -9,14 +9,20 @@ namespace Microsoft.DurableTask.Client;
 /// Extension methods for <see cref="StartOrchestrationOptions"/> to provide type-safe deduplication status configuration.
 /// </summary>
 public static class StartOrchestrationOptionsExtensions
-{
+{
+#pragma warning disable CS0618 // Type or member is obsolete - Cancelled is intentionally included for compatibility with the
+                               // Durable Task Framework
     public static readonly OrchestrationRuntimeStatus[] ValidDedupeStatuses = new[]
```
cgillum: Hmm...I'm surprised this is public and mutable. Users could definitely abuse this. It would be a breaking change, but I think we should either make this `internal` or we should make it into an immutable collection (like `IReadOnlyList<OrchestrationRuntimeStatus>`.

src/Client/OrchestrationServiceClientShim/ShimDurableTaskClient.cs
```diff
@@ -385,4 +387,48 @@ Task SendInstanceMessageAsync(string instanceId, HistoryEvent @event, Cancellati
...
+                }
+
+                if (runningStatuses.Contains(orchestrationStatus))
+                {
+                    // Check for cancellation before attempting to terminate the orchestration
+                    cancellation.ThrowIfCancellationRequested();
+
+                    string terminationReason = $"A new instance creation request has been issued for instance {instanceId} which " +
+                        $"currently has status {metadata.RuntimeStatus}. Since the dedupe statuses of the creation request, " +
+                        $"{(dedupeStatuses == null ? "[]" : string.Join(", ", dedupeStatuses))}, do not contain the orchestration's status, " +
```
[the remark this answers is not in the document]
cgillum: This seems valid to me.

PR #654, opened 2026-03-01
Propagate trace context for entity operations

### LINKED ISSUE #653: Entity operations don't propagate trace context from parent orchestration
## Summary

When an orchestration calls an entity via `CallEntityAsync()`, the DTS SDK does **not** propagate the orchestration's trace context (`ParentTraceContext`) to the entity operation. This causes entity spans (and any child spans created within the entity) to appear as disconnected root traces instead of being parented under the calling orchestration's trace tree.

## Root Cause

In [`ProtoUtils.ConstructOrchestratorResponse`](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Shared/Grpc/ProtoUtils.cs), `ParentTraceContext = CreateTraceContext()` is set for:

- ✅ `ScheduleTaskAction` (activities)
- ✅ `CreateSubOrchestrationAction` (sub-orchestrations)
- ❌ `SendEntityMessageAction` — **no trace context is set**

This means the W3C `traceparent` / `tracestate` that enables distributed trace stitching is never included in entity messages dispatched from orchestrations.

## Expected Behavior

Entity operations dispatched from an orchestration should inherit the orchestration's trace context, just like activities and sub-orchestrations do. The `invoke_agent` spans (or any other work done inside the entity) should appear as children of the orchestration span in a trace viewer.

## Actual Behavior

Entity operations start a new, independent trace. In a trace viewer (e.g., Aspire Dashboard, Jaeger), entity spans appear as separate root traces with no parent relationship to the orchestration that dispatched them.

## Impact

This is particularly impactful for scenarios that use Durable Entities as the execution model for agent invocations (e.g., `DurableAIAgent` in the Microsoft Agent Framework). Each `invoke_agent` span appears disconnected from the orchestration trace, making it impossible to view the full execution tree in a single trace.

## Suggested Fix

In `ProtoUtils.cs`, add `ParentTraceContext = CreateTraceContext()` to the `SendEntityMessage` action construction (similar to how it's done for `ScheduleTask` and `CreateSubOrchestration`). The entity work item processing path would also need to parse and restore this context when executing entity operations.

cgillum:
## Summary

Entity operations triggered from orchestrations don't propagate distributed trace context because `SendEntityMessageAction` lacks a `parentTraceContext` field, and entity operation events (`EntityOperationSignaledEvent`, `EntityOperationCalledEvent`) also lack this field. This causes entity operations to appear as disconnected root spans rather than being properly nested under their parent orchestration in distributed tracing tools.

### Proto updates (see https://github.com/microsoft/durabletask-protobuf/pull/64)
- Add `parentTraceContext` field to `SendEntityMessageAction` (field 5)
- Add `parentTraceContext` field to `EntityOperationSignaledEvent` (field 6) and `EntityOperationCalledEvent` (field 8)

### Trace context propagation (ProtoUtils.cs)
- **Sending side**: Set `ParentTraceContext` on `SendEntityMessageAction` in `ConstructOrchestratorResponse`, matching the existing pattern for tasks and sub-orchestrations. Uses `CreateTraceContext()` which generates a random `clientSpanId` for Client→Server span pairing.
- **Receiving side**: Extract `parentTraceContext` from `EntityOperationSignaledEvent` / `EntityOperationCalledEvent` in `ToEntityBatchRequest`, populating `OperationRequest.TraceContext`.

### Entity Server spans (TraceHelper.cs, Processor.cs)
- Add `StartTraceActivityForEntityOperation()` — creates a **Server** span during entity execution, parented under the Client span via the propagated trace context.
- Wire up in `OnRunEntityBatchAsync` using the first operation's trace context.

### Entity Client spans (TraceHelper.cs, Processor.cs)
- Add `EmitTraceActivityForEntityOperationCompleted()` and `EmitTraceActivityForEntityOperationFailed()` — emit **Client** spans during orchestration replay when entity operation results arrive.
- Add `StartTraceActivityForSchedulingEntityOperation()` — creates the Client span and uses `SetSpanId()` to match the `clientSpanId` from `ParentTraceContext`, creating the Client→Server nesting.
- Handle `EntityOperationCompleted` and `EntityOperationFailed` events in the orchestrator `newEvents` processing loop.
- Add `GetEntityOperationCalledEvent()` helper for correlating completion events with their original called events via `requestId`.

### Schema and constants
- Add `EntityOperation = "entity_operation"` to `TraceActivityConstants.cs`
- Add `EntityOperationName = "durabletask.entity.operation"` to `Schema.cs`

### Tests
- 5 unit tests for trace context propagation on the sending side
- 3 unit tests for trace context extraction on the receiving side

## How it works

Entity operations now follow the same Client→Server span pair pattern as activities and sub-orchestrations:

1. **Orchestration sends entity message** → `CreateTraceContext()` generates a random `clientSpanId` → stored in `SendEntityMessageAction.ParentTraceContext`
2. **Backend dispatches entity work** → copies `ParentTraceContext` to both the entity message event and the orchestrator history event
3. **Entity worker executes** → `OnRunEntityBatchAsync` creates a **Server** span using the propagated trace context (clientSpanId becomes the parent)
4. **Orchestration replays with result** → looks up `EntityOperationCalledEvent` from history → `StartTraceActivityForSchedulingEntityOperation` creates a **Client** span with `SetSpanId()` matching the `clientSpanId`
5. **Result**: Client span (orchestration) → Server span (entity) — properly nested in trace viewers

## Screenshots

For this fix, we used a Durable Agent scenario to validate. The source code is [here](https://github.com/microsoft/agent-framework/blob/23644ac6a78167d07a1959c017750438511d7896/dotnet/samples/04-hosting/DurableAgents/ConsoleApps/04_AgentOrchestration_Conditionals/Program.cs#L57-L99). TL/DR, it's an orchestration that calls two entities and then an activity function. The activity function call was needed so that we could compare the spans but also to ensure we didn't regress activity behavior (an earlier version of this change did regress activities a bit):

### Before

Here's what the scenario looked like prior to the fix.

image

Notice the entity traces don't appear at all. They can be seen as top-level unparented traces (not shown).

### After

Here's what the scenario looked like after the fix.

image

🎉

## Merge note

> **Do not merge** until the protobuf change (https://github.com/microsoft/durabletask-protobuf/pull/64) has been properly merged and vendored.

## Related PRs

- Proto: https://github.com/microsoft/durabletask-protobuf/pull/64
- DTS backend: Server-side changes to propagate `ParentTraceContext` on entity history events

Fixes #653


src/Shared/Grpc/ProtoUtils.cs
```diff
@@ -1063,7 +1065,7 @@ internal static T Base64Decode<T>(this MessageParser parser, string encodedMessa
                     }
                 }
 
-                // If the value starts with the 'dto:' prefix, it may represent a DateTime value � attempt to parse it.
+                // If the value starts with the 'dto:' prefix, it may represent a DateTime value � attempt to parse it.
```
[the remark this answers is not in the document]
cgillum: Not related to my change but fixed it anyways.

src/Grpc/orchestrator_service.proto
```diff
@@ -318,6 +320,8 @@ message SendEntityMessageAction {
         EntityLockRequestedEvent entityLockRequested = 3;
         EntityUnlockSentEvent entityUnlockSent = 4;
     }
+
+    TraceContext parentTraceContext = 5;
```
cgillum, line 325: **Note to self**: don't merge changes to this file until properly vendored (requires https://github.com/microsoft/durabletask-protobuf/pull/64 to be merged).

src/Shared/Grpc/ProtoUtils.cs
sophiatev, line 1: Can you fix the line endings on this diff?

cgillum: Sure. In the meantime, you can disable whitespace diffs to see what the actual changes were.

cgillum: Looks like this file previously had a mix of LF and CRLF and my editor normalized everything to CRLF. I can revert that to make the diff smaller but this might happen again the next time someone changes this file.

sophiatev: Oh I didn't realize this superpower existed. No worries then

cgillum: I reverted the line ending change.

sophiatev (dismissed):
Looks good, can you just add a screenshot of an example trace for posterity?

andystaples (dismissed):

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -865,6 +900,15 @@ async Task OnRunEntityBatchAsync(
...
             EntityId entityId = new(coreEntityId.Name, coreEntityId.Key);
 
+            // Start a trace span for entity operation execution using the first operation's trace context.
+            OperationRequest? firstOp = batchRequest.Operations?.FirstOrDefault();
+            using Activity? traceActivity = TraceHelper.StartTraceActivityForEntityOperation(
+                entityId.Name,
+                firstOp?.Operation,
+                batchRequest.InstanceId!,
+                firstOp?.TraceContext?.TraceParent,
+                firstOp?.TraceContext?.TraceState);
```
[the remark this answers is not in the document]
cgillum: The feedback is valid, but the suggested fix (setting the parent to null) is not. I'll create a proper fix in the next update.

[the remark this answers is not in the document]
cgillum: This should be addressed now.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -865,6 +900,29 @@ async Task OnRunEntityBatchAsync(
...
+                        batchRequest.InstanceId!,
+                        op.TraceContext?.TraceParent,
+                        op.TraceContext?.TraceState);
+
+                    if (activity != null)
+                    {
+                        traceActivities ??= [];
+                        traceActivities.Add(activity);
+                    }
+                }
```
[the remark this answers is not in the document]
cgillum: Done. Replaced the foreach loop with a LINQ \.Select(...).OfType<Activity>().ToList()\ expression.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -913,6 +971,12 @@ async Task OnRunEntityBatchAsync(
                 {
                     FailureDetails = new FailureDetails(frameworkException),
                 };
+
+                traceActivities?.ForEach(a => a.SetStatus(ActivityStatusCode.Error, frameworkException.Message));
+            }
+            finally
+            {
+                traceActivities?.ForEach(a => a.Dispose());
```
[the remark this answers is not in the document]
cgillum: A `using` statement can't be applied here because `traceActivities` is a `List<Activity>` (not a single `IDisposable`), and we need to conditionally set error status on each activity in the `catch` block and entity-not-found path before disposal. The `try/finally` pattern is the correct approach for managing a collection of disposable objects with conditional error marking.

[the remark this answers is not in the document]
cgillum: Good catch. Added error status marking for trace activities in the entity-not-found path, consistent with how `OnRunActivityAsync` handles the `ActivityTaskNotFound` case.

andystaples:
@cgillum What's the story with entity lock events wrt. distributed tracing? Do we currently emit spans/traces for locking? Is it worth adding here if not?

cgillum:
> What's the story with entity lock events wrt. distributed tracing? Do we currently emit spans/traces for locking? Is it worth adding here if not?

@andystaples I considered Lock requests as out-of-scope for this PR since it's not directly applicable to entity execution traces. However, I think it's a good future work item for us to consider since it would be helpful for users to be able to see when locks are added/removed.

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -409,6 +506,74 @@ static string CreateSpanName(string spanDescription, string? taskName, string? t
...
+    {
+        if (calledEvent == null)
+        {
+            return null;
+        }
+
+        string targetInstanceId = historyEvent?.EntityOperationCalled?.TargetInstanceId ?? string.Empty;
+
+        // Extract entity name from instance ID (format: "@name@key")
+        string entityName = targetInstanceId;
```
sophiatev, line 532: I believe we have APIs to do this in `EntityInstanceId` if I'm not mistaken?

src/Shared/Grpc/Tracing/TraceHelper.cs
```diff
@@ -409,6 +506,74 @@ static string CreateSpanName(string spanDescription, string? taskName, string? t
...
+            return null;
+        }
+
+        if (calledEvent.ParentTraceContext != null
+            && ActivityContext.TryParse(
+                calledEvent.ParentTraceContext.TraceParent,
+                calledEvent.ParentTraceContext?.TraceState,
+                out ActivityContext parentContext))
+        {
+            newActivity.SetSpanId(parentContext.SpanId.ToString());
```
sophiatev, line 562: Do you happen to know why we use `Activity.Current.Context` as the parent trace context but manually set the span ID here? Why aren't we setting the parent trace context to just be the `parentContext` extracted here?

I see this is done for the other methods here as well but I forget (or perhaps never understood) why

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -913,6 +965,12 @@ async Task OnRunEntityBatchAsync(
                 {
                     FailureDetails = new FailureDetails(frameworkException),
                 };
+
+                traceActivities.ForEach(a => a.SetStatus(ActivityStatusCode.Error, frameworkException.Message));
```
sophiatev, line 1163: Again just a question for my own understanding - in this case and the case above, we will just end up with Activities with just server spans and no corresponding client span?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -565,6 +573,33 @@ async Task OnRunOrchestratorAsync(
                                 newEvent.Timestamp.ToDateTime(),
                                 newEvent.TimerFired);
                             break;
+
+                        case P.HistoryEvent.EventTypeOneofCase.EntityOperationCompleted:
```
sophiatev, line 713: Where is the logic for emitting client spans for an entity signal request? Maybe I'm missing something but looks like this is just for entity calls

test/Worker/Grpc.Tests/ProtoUtilsTraceContextTests.cs
sophiatev, line 1: I just remembered that @philliphoff added an extensive `TracingIntegrationTests` class under `Grpc.Integration.Tests`. It would be good to give copilot the task of adding tests for the entities cases as well

PR #710, opened 2026-04-22
PROBE: validate dispatcher fix resolves retry-handler flake (DO NOT MERGE)

torosent:
Probe branch on top of #709's dispatcher fix to validate whether the strict `Assert.Equal(expectedNumberOfAttempts, retryHandlerCalls)` assertions are now stable on Windows CI.

If green across multiple runs → the dispatcher race was the real culprit and the SDK retry-handler design fix is hardening, not a release blocker.
If still flaky → the SDK retry path has its own bug to address.

**Do not merge.** Will be closed once we have signal.


torosent:
Probe complete — confirmed retry-handler over-invocation reproduces with strict assertions (failed on Linux at OrchestrationErrorHandling.cs:426). Resuming the SDK retry-handler design fix as a real (not Windows-specific) determinism bug.

PR #736, opened 2026-06-08
Ondemand sandbox

YunchuWang:
## The scenario

A .NET app using Durable Task Scheduler should be able to keep its normal orchestration worker local, but run selected expensive or isolated activities in a DTS-started sandbox worker container.

The intended flow is:

1. The local app registers orchestrators and local activities with the existing DTS client/worker APIs.
2. The local app declares a sandbox worker profile: activity names, container image, managed identities, CPU/memory, environment variables, and max concurrency.
3. DTS stores that declaration as configuration. It means: when this activity is needed, this is the sandbox worker profile that can run it.
4. DTS starts a sandbox worker container from the declared image.
5. The sandbox worker calls `UseSandboxWorker()`, connects back to DTS, reports the activities it registered, and heartbeats active activity count.
6. DTS only advertises live capacity after the worker registration matches the declaration, then routes the declared activity work to that sandbox worker.

In the sample, the local app runs the orchestration and `LocalHello`; the sandbox worker image runs `RemoteHello`. Both processes connect to the same task hub, but they must not receive the same work.

## What's missing

`durabletask-dotnet` did not have a .NET SDK surface for this pattern.

There was no client package for declaring sandbox worker profiles, no worker package for code running inside the sandbox, and no shared .NET transport for the sandbox activity protobuf service.

A few correctness details also fell out during validation:

- The local sample worker must call `UseWorkItemFilters()`. Without it, DTS can dispatch `RemoteHello` to the local worker, which does not implement that activity, so the orchestration can get stuck retrying the wrong worker.
- Worker registration metadata must be trustworthy. Sending a whitespace task hub or empty `DTS_SANDBOX_ID` makes server-side routing and telemetry hard to reason about.
- The sandbox worker needs activity start/completed notifications to report active activity count, but failures in that notification hook must not break normal activity dispatch.

## The change

The SDK surface is split into two opt-in preview packages:

- `Microsoft.DurableTask.Client.AzureManaged.Sandboxes`
- `Microsoft.DurableTask.Worker.AzureManaged.Sandboxes`

A good review path is:

1. Start with `src/Grpc/sandbox_service.proto` and `src/Shared/AzureManaged.Sandboxes/`. This is the shared contract and transport for declaring sandbox activities, removing declarations, and connecting a live sandbox worker stream.
2. Read `src/Client/AzureManaged.Sandboxes/`. This is the local declarer-app side: `[SandboxWorkerProfile]`, `SandboxWorkerProfileOptions`, declaration discovery, validation, and `SandboxActivitiesClient`.
3. Read `src/Worker/AzureManaged.Sandboxes/`. This is the sandbox-worker side: `UseSandboxWorker()`, activity-only filters, DTS-injected runtime settings, start/heartbeat messages, and active activity tracking.
4. Read the small gRPC worker hook changes under `src/Worker/Grpc/`. These add guarded activity execution notifications used by the sandbox worker heartbeat path.
5. Read `samples/on-demand-sandbox/` last. It shows the intended end-to-end shape: local declarer app, remote worker image, and shared activity names.

The new sandbox packages target `net10.0` only because this is new preview functionality. NuGet publish jobs are split per package so an already-published AzureManaged package version does not block publishing the new Sandboxes packages.


samples/on-demand-sandbox/main-app/appsettings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "OnDemandSandboxSample": {
+    "EndpointAddress": "https://<scheduler-endpoint>",
+    "TaskHubName": "OnDemandSandboxPocHub"
```
YunchuWang: use placeholder

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net8.0;net10.0</TargetFrameworks>
+    <PackageDescription>Azure Managed on-demand sandbox activity management extensions for the Durable Task Framework client.</PackageDescription>
+    <AssemblyName>Microsoft.DurableTask.Client.AzureManaged.OnDemandSandbox</AssemblyName>
```
cgillum: `OnDemandSandbox` is an awkward name. I suggest we simply uses the plural `Sandboxes` suffix. The "on-demand" part is mostly for marketing purposes.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <RootNamespace>Microsoft.DurableTask</RootNamespace>
```
cgillum: I suggest we remove this line and have the root namespace match the assembly name (the default behavior)

README.md
```diff
@@ -198,6 +198,8 @@ This SDK can also be used with the Durable Task Scheduler directly, without any
 
 For runnable DTS emulator examples that demonstrate versioning, see the [WorkerVersioningSample](samples/WorkerVersioningSample/README.md) (deployment-based versioning), the [EternalOrchestrationVersionMigrationSample](samples/EternalOrchestrationVersionMigrationSample/README.md) (multi-version routing with `[DurableTask(Version = "...")]`), the [ActivityVersioningSample](samples/ActivityVersioningSample/README.md) (activity versioning with inherited defaults and explicit override support), and the [EntityWithVersionedOrchestrationSample](samples/EntityWithVersionedOrchestrationSample/README.md) (a single instance migrating v1→v2 via `ContinueAsNew(NewVersion)` while preserving entity-held state).
 
+The [on-demand sandbox activities sample](samples/on-demand-sandbox/README.md) shows how to declare selected activities for DTS-managed on-demand sandbox execution and build the remote worker container image separately from the declarer app.
```
cgillum: nit: we haven't defined "DTS" in this document. I suggest either spelling it out or introducing the acronym at the beginning of this section to ensure that readers understand what it stands for.

CHANGELOG.md
```diff
@@ -1,465 +1,468 @@
...
-
-## v0.4.1-beta
-
-Initial public release
-
-
-
-
-
+# Changelog
```
cgillum: nit: did we change the newlines on this file? GitHub shows that every line was changed, making it harder to see the actual changes.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
```diff
@@ -0,0 +1,29 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net6.0;net8.0;net10.0</TargetFrameworks>
```
cgillum: I don't recommend targeting out-of-support versions of .NET. We should just target `net10.0` for this new feature unless we have specific customer asks for more versions.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
YunchuWang: Agreed. I will switch the .NET surface to use `Sandboxes` naming and remove `OnDemandSandbox` from the code/package-facing names where possible. The on-demand wording can stay in docs/description where it explains the preview scenario, but the code names should be the simpler sandbox terminology.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
             }
             else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.ActivityRequest)
             {
+                this.internalOptions.NotifyActivity?.Invoke(ActivityNotificationPhase.Started);
                 this.RunBackgroundTask(
                     workItem,
```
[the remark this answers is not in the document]
YunchuWang: Fixed in `fbc550c`. `NotifyActivity` now runs through a guarded helper that logs callback failures and keeps activity dispatch/completion flowing; added a regression test for this path.

samples/on-demand-sandbox/main-app/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.DurableTask.Worker.AzureManaged;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+const string Input = "on-demand-sandbox-sample";
+
+HostApplicationBuilder builder = Host.CreateApplicationBuilder(args);
+string endpoint = builder.Configuration["OnDemandSandboxSample:EndpointAddress"]!;
+string taskHub = builder.Configuration["OnDemandSandboxSample:TaskHubName"]!;
```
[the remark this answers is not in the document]
YunchuWang: Fixed in `fbc550c`. The sample now validates required configuration at startup with an explicit setting name and trims the returned value.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
YunchuWang: Fixed in `fbc550c`. Renamed the new package/code-facing SDK surface to `Sandboxes`/`Sandbox...` names, while leaving the protobuf wire contract names unchanged.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
YunchuWang: Fixed in `fbc550c`. Removed the explicit `RootNamespace` from both new package projects so the default root namespace follows the assembly name.

README.md
YunchuWang: Fixed in `fbc550c`. The README now spells out Durable Task Scheduler (DTS) in the sample sentence.

CHANGELOG.md
YunchuWang: Fixed in `fbc550c`. Restored the changelog line endings so the diff is back to the actual changelog entry changes instead of the whole file.

src/Client/AzureManaged.OnDemandSandbox/Client.AzureManaged.OnDemandSandbox.csproj
YunchuWang: Fixed in `fbc550c`. The new sandbox packages now target `net10.0` only.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -449,6 +460,18 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
...
+        void NotifyActivity(ActivityNotificationPhase phase)
+        {
+            try
+            {
+                this.internalOptions.NotifyActivity?.Invoke(phase);
+            }
+            catch (Exception ex)
+            {
+                this.Logger.ActivityNotificationFailed(phase, ex);
+            }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in `6864544`. `NotifyActivity` now uses a filtered catch so ordinary callback failures are logged/swallowed, while fatal CLR exceptions are allowed to propagate. Verified with the focused `GrpcDurableTaskWorkerTests` run.

src/Worker/AzureManaged.Sandboxes/DurableTaskSchedulerSandboxWorkerExtensions.cs
```diff
@@ -0,0 +1,209 @@
...
+        builder.UseDurableTaskScheduler(options =>
+        {
+            options.EndpointAddress = endpoint;
+            options.TaskHubName = taskHub;
+            options.AllowInsecureCredentials = true;
+            if (UsesManagedIdentityAuthentication(Environment.GetEnvironmentVariable("DTS_AUTHENTICATION")))
+            {
+                options.Credential = CreateManagedIdentityCredential();
+                options.AllowInsecureCredentials = false;
+            }
```
torosent: If the default is `AllowInsecureCredentials = true`, it means that missing/misspelled env var silently configures unauthenticated scheduler access.

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
```diff
@@ -0,0 +1,405 @@
...
+                await this.WriteSessionMessageAsync(registrationSession, startMessage, cancellationToken).ConfigureAwait(false);
+                Logs.SandboxActivityWorkerRegistered(
+                    this.logger,
+                    startMessage.Start.TaskHub,
+                    activityCount,
+                    startMessage.Start.SandboxProvider,
+                    startMessage.Start.DtsSandboxIdentifier);
+
+                retryDelay = this.GetInitialRetryDelay();
+                await this.RunRegistrationSessionAsync(registrationSession, cancellationToken).ConfigureAwait(false);
```
torosent: Copilot: Backoff resets immediately after the start write succeeds. If WaitForCompletionAsync() fails immediately every time, retries remain at initial delay.
Suggested fix: 
Reset retry delay only after a session survives a minimum healthy duration or after first successful heartbeat / server acknowledgement. Add a regression test where ResponseAsync fails immediately after start.

src/Client/AzureManaged.Sandboxes/SandboxActivityDeclarationProvider.cs
```diff
@@ -0,0 +1,150 @@
...
+            foreach (Type type in types)
+            {
+                yield return type;
+            }
+        }
+    }
+
+    static void ValidateActivityOwnership(IEnumerable<SandboxWorkerProfileOptions> declarations)
+    {
+        Dictionary<string, string> activityOwners = new(StringComparer.Ordinal);
```
torosent: Change to `StringComparer.OrdinalIgnoreCase` to avoid duplicate ownership

src/Grpc/orchestrator_service.proto
```diff
@@ -370,12 +370,14 @@ message OrchestratorResponse {
...
+    /* Chunking logic has since been deprecated and fields related to it are marked as such */
+
     // True if this is a partial (chunked) completion. The backend must keep the work item open until the final chunk (isPartial=false).
-    bool isPartial = 8;
+    bool isPartial = 8 [deprecated=true];
 
     // Zero-based position of the current chunk within a chunked completion sequence.
     // This field is omitted for non-chunked completions.
-    google.protobuf.Int32Value chunkIndex = 9;
+    google.protobuf.Int32Value chunkIndex = 9 [deprecated=true];
```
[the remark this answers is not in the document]
YunchuWang: this aligns with proto

samples/on-demand-sandbox/main-app/Orchestrators.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.DurableTask.Samples.OnDemandSandbox.MainApp;
+
+[DurableTask(nameof(HelloOrchestrator))]
+internal sealed class HelloOrchestrator : TaskOrchestrator<string, string>
+{
+    public override async Task<string> RunAsync(TaskOrchestrationContext context, string input)
+    {
+        string localResult = await context.CallActivityAsync<string>(OnDemandSandboxTaskNames.LocalHello, input);
+        string remoteResult = await context.CallActivityAsync<string>(ActivityNames.RemoteHello, input);
```
halspang: Nit: Instead of defining the constant classes to share the names, could we just use the generated context methods to call the activities? Or, are those not supported with sandbox work?

samples/on-demand-sandbox/README.md
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+```
+
+Then run the main app:
+
+```powershell
+Push-Location .\samples\on-demand-sandbox\main-app
+dotnet run
+Pop-Location
```
halspang: Nit: `Push-Location` and `Pop-Location` aren't super common in these instructions. I'd rather we just include the path in `dotnet run ...` as I think that's a little cleaner.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+            return decimal.TryParse(
+                value[..^2],
+                NumberStyles.Number,
+                CultureInfo.InvariantCulture,
+                out decimal mib)
+                ? (long)mib
+                : null;
+        }
+
+        return decimal.TryParse(
```
halspang, line 208: After checking for a labeled `MiB`, we then also default to it? Should we also check for `KiB` or `B`? I'm not actually sure if those are options but this feels limited in what we check.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+                "On-demand sandbox activity memory resources must be a positive Kubernetes-style memory quantity. " +
+                "Use formats like '256Mi', '1Gi', or '2048'.");
+        }
+
+        return normalized;
+    }
+
+    static long? TryParseCpuMillicores(string value)
+    {
+        if (value.EndsWith('m') || value.EndsWith('M'))
```
halspang, line 164: I see a `m` in things like k8s definitions for CPU limits, but I don't see an `M`. If we're going by SI notation, wouldn't an `M` denote a `megaCore` instead of a `milliCore`? 

Though, to be fair, I haven't played around a ton with resource limits so if I'm over thinking this feel free to ignore this/leave in the `M`, but I don't think we need it and I'd prefer a stricter set supported as opposed to a more lenient set that makes assumptions on what we were given.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
halspang: Looking at k8s, it supports these values `E, P, T, G, M, K, Ei, Pi, Ti, Gi, Mi, Ki`. We may not need to support them all, but we should be more strict on what we do support here. It does look like the default unit is `B` so this could lead to some accidentally large memory footprint workers.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// registered activities separately when they connect.
+    /// </summary>
+    internal IList<SandboxActivityMetadata.Activity> Activities { get; } = new List<SandboxActivityMetadata.Activity>();
+
+    /// <summary>
+    /// Adds an activity name and version to the on-demand sandbox worker profile workerProfile.
+    /// </summary>
+    /// <param name="activityName">The activity name.</param>
+    /// <param name="version">The activity version, or <see langword="null" /> for unversioned/wildcard activity execution.</param>
+    public void AddActivity(string activityName, string? version)
```
halspang, line 71: Nit: Might be nice to give a range add here.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Reflection;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.AzureManaged.Internal;
+
+namespace Microsoft.DurableTask.Client.AzureManaged;
+
+/// <summary>
+/// Options for declaring on-demand sandbox activities and the worker image DTS should start for them.
+/// </summary>
+public sealed class SandboxWorkerProfileOptions
```
halspang, line 13: I'd prefer if this matched the proto configuration a bit more. Specifically, let's pull out things like the image settings to an object like it is in the proto. The more the domain and wire types match, the easier they are to maintain, at least in my opinion.

src/Grpc/refresh-protos.ps1
```diff
@@ -17,14 +17,23 @@ $commitDetails = Invoke-RestMethod -Uri "https://api.github.com/repos/microsoft/
...
-$protoFileNames = @(
-    "orchestrator_service.proto"
+$protoFiles = @(
+    @{
+        SourcePath = "orchestrator_service.proto"
+        OutputFileName = "orchestrator_service.proto"
+    },
+    @{
+        SourcePath = "durable-task-scheduler/sandbox_service.proto"
+        OutputFileName = "sandbox_service.proto"
```
halspang: Think we should maintain the path here?

src/Shared/AzureManaged.Sandboxes/SandboxActivitiesGrpcTransport.cs
```diff
@@ -0,0 +1,178 @@
...
+        Proto.SandboxActivities.SandboxActivitiesClient client,
+        bool attachTaskHubMetadata = true)
+    {
+        this.client = Check.NotNull(client);
+        this.attachTaskHubMetadata = attachTaskHubMetadata;
+    }
+
+    /// <inheritdoc/>
+    public async Task<Proto.DeclareSandboxWorkerProfileResult> DeclareSandboxWorkerProfileAsync(
+        Proto.SandboxWorkerProfile workerProfile,
```
halspang, line 95: Just checking, but users won't interact with this directly, right? We tend to not have the proto types in the method signatures and convert them here on behalf of the user to keep the protos a little more isolated.

src/Worker/AzureManaged.Sandboxes/DurableTaskSchedulerSandboxWorkerExtensions.cs
```diff
@@ -0,0 +1,219 @@
...
+        {
+            options.TaskHub = taskHubName;
+        }
+    }
+
+    static void ConfigureDurableTaskSchedulerFromEnvironment(IDurableTaskWorkerBuilder builder)
+    {
+        string endpoint = GetRequiredEnvironmentVariable("DTS_ENDPOINT");
+        string taskHub = GetRequiredEnvironmentVariable("DTS_TASK_HUB");
+        string authentication = GetRequiredEnvironmentVariable("DTS_AUTHENTICATION");
```
halspang: These required env var names should be constants. Carry that through for the other env vars in this file.

src/Worker/AzureManaged.Sandboxes/Logs.cs
```diff
@@ -0,0 +1,68 @@
...
+    [LoggerMessage(
+        EventId = 708,
+        Level = LogLevel.Debug,
+        Message = "Ignoring on-demand sandbox heartbeat pump failure after registration session completion.")]
+    public static partial void SandboxHeartbeatPumpFailureIgnored(ILogger logger, Exception exception);
+
+    [LoggerMessage(
+        EventId = 709,
+        Level = LogLevel.Debug,
+        Message = "Ignoring on-demand sandbox worker session completion failure after heartbeat pump failure.")]
```
halspang: `pump`?

src/Worker/AzureManaged.Sandboxes/SandboxActivityTracker.cs
```diff
@@ -0,0 +1,42 @@
...
+/// Tracks activity execution state for an on-demand sandbox worker process.
+/// </summary>
+sealed class SandboxActivityTracker
+{
+    int activeActivityCount;
+
+    /// <summary>
+    /// Gets the number of activities currently in flight on this worker.
+    /// </summary>
+    public int InFlightCount => Volatile.Read(ref this.activeActivityCount);
```
halspang, line 16: If this is used for any kind of decision making, should we perform a locked read on it? If a user starts a new activity right before a previous one finished, would it deny that activity and not come back to it?

samples/on-demand-sandbox/main-app/WorkerProfiles.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        options.ContainerImage = GetRequiredEnvironmentVariable("DTS_SANDBOX_CONTAINER_IMAGE");
+        options.ImagePullManagedIdentityClientId = GetRequiredEnvironmentVariable("DTS_SANDBOX_IMAGE_PULL_UMI_CLIENT_ID");
+        options.SchedulerManagedIdentityClientId = GetRequiredEnvironmentVariable("DTS_SANDBOX_SCHEDULER_UMI_CLIENT_ID");
+        options.Cpu = "1000m";
+        options.Memory = "2048Mi";
+        options.MaxConcurrentActivities = 1;
+        AddEnvironmentVariable(options, "SAMPLE_REMOTE_MARKER");
+        AddEnvironmentVariable(options, "SAMPLE_REMOTE_DELAY_MS");
+        options.AddActivity(SandboxActivities.RemoteHelloName, SandboxActivities.RemoteHelloVersion);
```
halspang: In the proto, I remember there being a `cmd` or `entrypoint`. Do we not need those to start every image?

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
```diff
@@ -0,0 +1,443 @@
...
+    async Task RunRegistrationLoopAsync(int activityCount, CancellationToken cancellationToken)
+    {
+        TimeSpan retryDelay = this.GetInitialRetryDelay();
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            ISandboxActivityWorkerSession? registrationSession = null;
+            try
+            {
+                Proto.SandboxActivityWorkerMessage startMessage = SandboxWorkerMessageBuilder.BuildWorkerStart(this.options, this.registeredActivities);
+                registrationSession = this.transport.OpenSandboxActivityWorkerSession(startMessage.Start.TaskHub, cancellationToken);
```
halspang, line 205: The first message has to be the start message, but this loop retries. So, do we need the start message after each failure? Or, are we supposed to do it once and retry within the standard stream?

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    async Task RunRegistrationSessionAsync(
+        ISandboxActivityWorkerSession registrationSession,
+        CancellationToken cancellationToken)
+    {
+        using CancellationTokenSource heartbeatCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        Task heartbeatTask = this.PumpHeartbeatsAsync(registrationSession, heartbeatCts.Token);
+        Task<Proto.SandboxActivityWorkerSessionResult> completionTask = registrationSession.WaitForCompletionAsync();
+        Task completedTask = await Task.WhenAny(heartbeatTask, completionTask).ConfigureAwait(false);
```
halspang, line 290: Am I reading this correctly that all this loop does is wait for either the heartbeat or registration to complete and then it ends? Wouldn't this trigger us sending a new start message in the upper loop?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
...
+                            await this.OnRunActivityAsync(
+                                workItem.ActivityRequest,
+                                workItem.CompletionToken,
+                                cancellation).ConfigureAwait(false);
+                        }
+                        finally
+                        {
+                            this.NotifyActivity(ActivityNotificationPhase.Completed);
+                        }
+                    },
```
halspang, line 424: Was this implicitly async/awaited in the previous context? I don't see changes to the method but I do see the configure awaiter. Was this just so we could enable the try..finally?

CHANGELOG.md
```diff
@@ -1,466 +1,469 @@
-# Changelog
```
halspang, line 1: Can you fix this? I'm guessing you have a line-ending issue but it makes it look like you rewrote the entire changelog and I don't like it.

halspang (commented):
Mostly a lot of questions, I'm still not exactly sure what we're accomplishing here but I'll continue familiarizing myself with this and give you a chance to answer my first chunk of questions.

samples/on-demand-sandbox/main-app/WorkerProfiles.cs
YunchuWang: yes they are optional settings, we want to leave them out in sample to make it more intuitive for customers

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            return decimal.TryParse(
+                value[..^1],
+                NumberStyles.Number,
+                CultureInfo.InvariantCulture,
+                out decimal milliCpu)
+                ? (long)milliCpu
+                : null;
+        }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4961681. The m-suffix path now uses integer parsing, so values like 500.5m are rejected instead of being truncated. After parsing, the SDK also validates against ADC sandbox CPU tiers: 250m through 16000m in 250m increments. ADC proof: CPU quantity formats include 500m, 2, and 0.5: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/CpuQuantity.cs#L6. ADC parses m as millicores and decimal values as cores: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Helpers/ResourceParser.cs#L35-L45. Sandbox CPU tiers are enforced by SandboxResourceTierRegistry.IsValidCpu: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/SandboxResourceTier.cs#L18-L54.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
YunchuWang: Fixed in 4961681. The SDK now uses the stricter public subset here: lowercase m is accepted for millicores, but uppercase M is rejected. ADC proof: ADC parser currently accepts both m and M for historical permissiveness, but its documented CPU examples are lowercase m plus decimal cores: 500m, 2, 0.5: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/CpuQuantity.cs#L6-L7. The SDK follows those documented examples and avoids treating SI M as millicores.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
YunchuWang: Fixed in 4961681. I kept memory formats aligned with ADC rather than the full Kubernetes quantity grammar: Mi/Gi are accepted case-insensitively, and a bare number is accepted as MiB. I did not add Ki, B, M, G, etc. support because ADC MemoryQuantity does not support those. ADC proof: MemoryQuantity documents 256Mi, 1Gi, and bare 2048: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L6. MemoryQuantity parses Gi/Mi with StringComparison.OrdinalIgnoreCase, then falls back to parsing a bare number as MiB: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L55-L82. The shared ResourceParser has the same documented bare-MiB behavior: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Helpers/ResourceParser.cs#L55-L57. To address accidental large footprint workers, the SDK now enforces the ADC sandbox service rule that memory cannot exceed the tier maximum for the selected CPU.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        return decimal.TryParse(
+            value,
+            NumberStyles.Number,
+            CultureInfo.InvariantCulture,
+            out decimal cores)
+            ? (long)(cores * 1000)
+            : null;
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4961681. CPU parsing now range-checks decimal core values before converting to millicores, and ADC tier validation rejects anything outside 250m..16000m or not divisible by 250m, so extreme CPU inputs flow through the SDK validation error path instead of raw overflow behavior. ADC proof: ADC parses decimal CPU values as cores: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Helpers/ResourceParser.cs#L42-L45. Sandbox CPU values are constrained to 250m..16000m in 250m increments: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/SandboxResourceTier.cs#L18-L54.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,217 @@
...
+        if (value.EndsWith("Gi", StringComparison.OrdinalIgnoreCase))
+        {
+            return decimal.TryParse(
+                value[..^2],
+                NumberStyles.Number,
+                CultureInfo.InvariantCulture,
+                out decimal gib)
+                ? (long)(gib * 1024)
+                : null;
+        }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4961681. The GiB branch now range-checks before multiplying/casting, and the result must still pass the ADC sandbox memory cap for the configured CPU. ADC proof: ADC memory parsing accepts Gi case-insensitively: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L57-L65. ADC sandbox memory is capped at cores * 2Gi: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/SandboxResourceTier.cs#L30-L40. Sandbox creation rejects memory above that tier maximum: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Global.Common/Services/GlobalSandboxService.cs#L6163-L6172.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,217 @@
...
+        if (value.EndsWith("Mi", StringComparison.OrdinalIgnoreCase))
+        {
+            return decimal.TryParse(
+                value[..^2],
+                NumberStyles.Number,
+                CultureInfo.InvariantCulture,
+                out decimal mib)
+                ? (long)mib
+                : null;
+        }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4961681. The MiB branch now range-checks before casting, and over-tier values are rejected by the ADC memory cap validation for the selected CPU. ADC proof: ADC memory parsing accepts Mi case-insensitively: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L68-L76. ADC sandbox memory is capped at cores * 2Gi: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/SandboxResourceTier.cs#L30-L40. Sandbox creation rejects memory above that tier maximum: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Global.Common/Services/GlobalSandboxService.cs#L6163-L6172.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        return decimal.TryParse(
+            value,
+            NumberStyles.Number,
+            CultureInfo.InvariantCulture,
+            out decimal parsed)
+            ? (long)parsed
+            : null;
```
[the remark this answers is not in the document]
YunchuWang: Fixed in 4961681. The bare-number fallback now range-checks before casting. I kept the fallback because ADC explicitly treats bare memory quantities as MiB, then the SDK applies the same ADC sandbox memory cap for the configured CPU. ADC proof: MemoryQuantity documents bare 2048: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L6. MemoryQuantity falls back to parsing a bare number as MiB: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/MemoryQuantity.cs#L79-L82. Sandbox creation rejects memory above the CPU tier maximum: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Global.Common/Services/GlobalSandboxService.cs#L6163-L6172.

samples/on-demand-sandbox/main-app/Orchestrators.cs
YunchuWang: discussed with Tomer/Nick. the main app does not have activity registered (activity is running in remote worker image), so customer can declare with activity name as string list.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
YunchuWang: Correction: updated in 52f90b2 to accept uppercase M as well, so the SDK now mirrors ADC CpuQuantity parsing here instead of using the stricter lowercase-only subset. The parser still requires the millicore part to be an integer and still applies ADC CPU tier validation after parsing. ADC proof: CpuQuantity accepts both m and M: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/TypedString/CpuQuantity.cs#L57-L65. Sandbox CPU tiers are still enforced as 250m..16000m in 250m increments: https://github.com/microsoft/adc/blob/73f0f1044/src/Adc.Common.Core/Model/SandboxResourceTier.cs#L18-L54.

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
halspang: I was not reading this correctly, I see that both the heartbeat and session only complete their tasks on a terminal message.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
YunchuWang: Addressed in b6cfad9. Added a public `SandboxWorkerProfileOptions.Activity` value type and `AddActivities(IList<Activity>)` so callers can add a range of activity declarations without repeating `AddActivity(...)`. The implementation delegates each item through the existing `AddActivity(string, string?)` path, so name/version trimming and validation remain centralized. Added focused coverage in `SandboxActivitiesClientTests` for the new API surface and range behavior.

YunchuWang: updated

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
YunchuWang: agree todo

src/Grpc/refresh-protos.ps1
YunchuWang: agree todo

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
YunchuWang: Implemented in 3367f6b. `SandboxWorkerProfileOptions` now exposes an `Image` object that mirrors the proto image section: `Image.ImageRef`, `Image.ManagedIdentityClientId`, `Image.Entrypoint`, and `Image.Cmd`. The builder now maps that object in `BuildImage(...)`, and the sample/tests were updated to use `options.Image`. Focused `SandboxActivitiesClientTests` passed locally (33/33).

YunchuWang: Follow-up in a1a974d: changed the public XML docs and validation message for `Image.ManagedIdentityClientId` to use provider-neutral wording (`container provider`) instead of naming ADC in the SDK surface.

YunchuWang: Follow-up in c84d8bf: simplified the public wording further to avoid introducing a provider concept. The XML doc now says the identity is `used to pull the on-demand sandbox worker image`, and the validation message matches that phrasing.

src/Worker/AzureManaged.Sandboxes/DurableTaskSchedulerSandboxWorkerExtensions.cs
YunchuWang: agree todo

src/Worker/AzureManaged.Sandboxes/Logs.cs
YunchuWang: agree confusing naming todo

src/Shared/AzureManaged.Sandboxes/SandboxActivitiesGrpcTransport.cs
YunchuWang: Confirmed: users do not interact with this transport directly. Both `ISandboxActivitiesTransport` and `SandboxActivitiesGrpcTransport` are internal, and the public `SandboxActivitiesClient` only exposes `EnableSandboxActivitiesAsync` / `RemoveSandboxWorkerProfileAsync`. The user-facing worker profile options are converted to `Proto.SandboxWorkerProfile` by `SandboxWorkerProfileBuilder` before reaching this internal gRPC transport boundary.

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
YunchuWang: The retry opens a new registration stream each time. The failed session is disposed in the loop's `finally` block, and the next iteration calls `OpenSandboxActivityWorkerSession` again before sending `Start`. The backend treats `Start` as the first message for each stream and rejects only a second `Start` within the same stream, so each reconnect must send a fresh `Start` message. This is also covered by `SandboxActivityWorkerRegistrationHostedService_ReconnectsAfterRetriableRegistrationFailure`, which verifies the failed and recovered sessions each receive exactly one `Start` message.

src/Worker/AzureManaged.Sandboxes/SandboxActivityTracker.cs
YunchuWang: This value is only used for heartbeat reporting (`SandboxActivityWorkerHeartbeat.ActiveActivitiesCount`), not for activity admission or dispatch decisions. Worker-side concurrency is enforced separately by `DurableTaskWorkerOptions.Concurrency.MaximumConcurrentActivityWorkItems`, which is set from `SandboxWorkerRuntimeOptions.MaxConcurrentActivities` after reading `DTS_SANDBOX_MAX_ACTIVITIES`. The tracker read uses `Volatile.Read`, and updates use `Interlocked`/`CompareExchange`, so a heartbeat can at worst report a slightly stale in-flight count for one interval. It will not deny a new activity or prevent the dispatcher from coming back to it.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
YunchuWang: Yes. The previous delegate already returned the `OnRunActivityAsync(...)` task, and `RunBackgroundTask` awaits the `Func<Task>` via `await handler()`. The `async` lambda is only needed so the awaited activity execution can be wrapped in `try/finally`; that guarantees the sandbox activity tracker receives `Completed` after the activity finishes, including failure and cancellation paths. `Started` intentionally stays before `RunBackgroundTask` so the tracker counts the work item as soon as dispatch accepts/schedules it, rather than waiting for the background task body to begin.

src/Grpc/refresh-protos.ps1
YunchuWang: System.Collections.Hashtable.Body

src/Worker/AzureManaged.Sandboxes/DurableTaskSchedulerSandboxWorkerExtensions.cs
YunchuWang: System.Collections.Hashtable.Body

src/Worker/AzureManaged.Sandboxes/Logs.cs
YunchuWang: System.Collections.Hashtable.Body

CHANGELOG.md
YunchuWang: System.Collections.Hashtable.Body

src/Grpc/refresh-protos.ps1
YunchuWang: Addressed in d7877c6. sandbox_service.proto now keeps the upstream durable-task-scheduler/ path under src/Grpc, and refresh-protos.ps1 writes files to their source-relative paths, creating subdirectories as needed. Verified with dotnet build src/Grpc/Grpc.csproj --no-restore.

src/Worker/AzureManaged.Sandboxes/DurableTaskSchedulerSandboxWorkerExtensions.cs
YunchuWang: Addressed in d7877c6. Added SandboxWorkerEnvironmentVariables constants for the DTS-injected sandbox worker environment variables and updated the worker extension/message builder to use them. Verified with focused Worker.AzureManaged.Tests sandbox tests.

src/Worker/AzureManaged.Sandboxes/Logs.cs
YunchuWang: Addressed in d7877c6. Replaced the confusing pump wording with loop in the sandbox worker log messages and internal method/member names, while keeping the same event IDs and behavior. Verified with focused Worker.AzureManaged.Tests sandbox tests.

CHANGELOG.md
YunchuWang: Checked the current PR diff and this no longer appears to be a line-ending rewrite. git diff --stat origin/main...HEAD -- CHANGELOG.md reports only 3 insertions, and git diff --numstat origin/main...HEAD -- CHANGELOG.md reports 3 0 CHANGELOG.md. Resolving this as already fixed.

test/Client/AzureManaged.Tests/SandboxActivitiesClientTests.cs
```diff
@@ -0,0 +1,448 @@
...
+        await client.RemoveSandboxWorkerProfileAsync("profile-a");
+
+        // Assert
+        callInvoker.RemoveRequest.Should().NotBeNull();
+        callInvoker.RemoveRequest!.WorkerProfileId.Should().Be("profile-a");
+        callInvoker.RemoveHeaders.Should().Contain(header => header.Key == "taskhub" && header.Value == "client-test-taskhub");
+    }
+
+    [Fact]
+    public async Task EnableSandboxActivitiesAsync_SendsWorkerProfileWorkerProfiles()
```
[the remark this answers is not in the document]
YunchuWang: Fixed in the latest push by renaming the test to `EnableSandboxActivitiesAsync_SendsWorkerProfiles`.

test/Client/AzureManaged.Tests/SandboxActivitiesClientTests.cs
```diff
@@ -0,0 +1,481 @@
...
```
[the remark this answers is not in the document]
YunchuWang: Fixed in the latest push by renaming the test to `EnableSandboxActivitiesAsync_SendsWorkerProfiles`.

src/Grpc/orchestrator_service.proto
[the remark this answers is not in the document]
YunchuWang: Fixed in `f4ec355`. The proto stays synced as-is, and I added narrow `CS0612` suppressions around the remaining wire-compatibility reads/writes for `IsPartial` and `ChunkIndex` in the worker, in-process test host, and AzureBlobPayloads interceptor. I rebuilt the affected projects and verified the `IsPartial` / `ChunkIndex` obsolete warnings are gone.

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// Determines whether two activities represent the same activity identity.
+    /// </summary>
+    /// <param name="left">The left activity.</param>
+    /// <param name="right">The right activity.</param>
+    /// <returns><see langword="true" /> if the activities are equal; otherwise, <see langword="false" />.</returns>
+    public static bool ActivityEquals(Activity left, Activity right) =>
+        string.Equals(left.Name, right.Name, StringComparison.OrdinalIgnoreCase)
+        && string.Equals(left.Version, right.Version, StringComparison.Ordinal);
+
```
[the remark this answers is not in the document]
YunchuWang: System.Collections.Hashtable.Body

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Determines whether two activity filters can match overlapping work.
+    /// </summary>
+    /// <param name="left">The left activity.</param>
+    /// <param name="right">The right activity.</param>
+    /// <returns><see langword="true" /> if the activities overlap; otherwise, <see langword="false" />.</returns>
+    public static bool ActivitiesOverlap(Activity left, Activity right) =>
+        string.Equals(left.Name, right.Name, StringComparison.OrdinalIgnoreCase)
+        && (left.Version is null
+            || right.Version is null
+            || string.Equals(left.Version, right.Version, StringComparison.Ordinal));
```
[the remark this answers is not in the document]
YunchuWang: System.Collections.Hashtable.Body

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
```diff
@@ -9,6 +10,7 @@
 using Microsoft.DurableTask.Tests.Logging;
 using Microsoft.DurableTask.Worker;
 using Microsoft.DurableTask.Worker.Grpc.Internal;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
```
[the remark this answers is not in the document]
YunchuWang: System.Collections.Hashtable.Body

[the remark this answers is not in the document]
YunchuWang: System.Collections.Hashtable.Body

samples/on-demand-sandbox/README.md
```diff
@@ -0,0 +1,72 @@
...
+dotnet build .\samples\on-demand-sandbox\remote-worker\remote-worker.csproj
+```
+
+## Build the remote worker image
+
+Run from the repository root:
+
+```powershell
+$image = "<acr-name>.azurecr.io/dts-ondemand-sandbox-sample:<tag>"
+docker build -f .\samples\on-demand-sandbox\remote-worker\Containerfile -t $image .
```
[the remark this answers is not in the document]
YunchuWang: System.Collections.Hashtable.Body

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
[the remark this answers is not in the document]
YunchuWang: Fixed in 4f286f4. `SandboxActivityMetadata.ActivityEquals` now compares versions with `StringComparison.OrdinalIgnoreCase`, matching `TaskVersion`/`TaskVersionKey` semantics. Added focused coverage to ensure same-name/same-version-different-casing declarations deduplicate.

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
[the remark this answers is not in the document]
YunchuWang: Fixed in 4f286f4. `SandboxActivityMetadata.ActivitiesOverlap` now treats versions case-insensitively, matching `TaskVersion`/`TaskVersionKey`; added coverage that cross-profile ownership validation catches `v1`/`V1` overlaps.

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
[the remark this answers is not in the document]
YunchuWang: Fixed in 4f286f4 by removing the unused `Microsoft.Extensions.DependencyInjection` import. Verified with focused `Worker.Grpc.Tests` for `GrpcDurableTaskWorkerTests`.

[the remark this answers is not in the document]
YunchuWang: Fixed in 4f286f4 by removing the unused `Microsoft.Extensions.DependencyInjection` import. Verified with focused `Worker.Grpc.Tests` for `GrpcDurableTaskWorkerTests`.

samples/on-demand-sandbox/README.md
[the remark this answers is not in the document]
YunchuWang: Checked this against Docker docs and the current sample layout. This is already valid: Docker supports Dockerfile-specific ignore files named `<Dockerfile>.dockerignore` placed next to the Dockerfile, and this sample has `remote-worker/Containerfile.dockerignore` next to `remote-worker/Containerfile`. The README uses `docker build -f ./samples/on-demand-sandbox/remote-worker/Containerfile ... .`, so Docker/BuildKit can use that Dockerfile-specific ignore file. Resolving without code change.

test/Client/AzureManaged.Tests/SandboxActivitiesClientTests.cs
```diff
@@ -0,0 +1,499 @@
...
+    {
```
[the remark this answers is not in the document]
YunchuWang: Already fixed by bfc1118: the test is now named `EnableSandboxActivitiesAsync_SendsWorkerProfiles`. This thread is marked outdated, so resolving.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,261 @@
...
+
+    static long? TryConvertMemoryToMiB(decimal value, decimal multiplier)
+    {
+        if (value < 0 || value > long.MaxValue / multiplier)
+        {
+            return null;
+        }
+
+        return (long)(value * multiplier);
+    }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in the latest push. `TryConvertMemoryToMiB` now rejects values that do not convert to a whole MiB instead of truncating them, and `SandboxActivitiesClientTests` covers `0.1Gi`, `512.5Mi`, and bare `1024.5` as invalid memory quantities.

src/Worker/AzureManaged.Sandboxes/SandboxActivityWorkerRegistrationHostedService.cs
```diff
@@ -0,0 +1,443 @@
...
+                this.lifetime?.StopApplication();
+                break;
+            }
+            finally
+            {
+                if (registrationSession is not null)
+                {
+                    this.ClearCurrentSession(registrationSession);
+                    await DisposeSessionAsync(registrationSession, this.logger).ConfigureAwait(false);
+                }
```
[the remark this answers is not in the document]
YunchuWang: Fixed in the latest push. `StopAsync` now only cancels and waits for the registration loop, so the registration loop owns graceful session completion and disposal. Both `CompleteAsync` and `DisposeAsync` now run through the same `streamSync` path, which prevents disposal from racing with an in-flight stream write or completion. Added focused coverage that blocks completion and verifies disposal does not happen until completion finishes.

src/Grpc/orchestrator_service.proto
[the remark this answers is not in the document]
YunchuWang: already fixed

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
             }
             else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.ActivityRequest)
             {
+                this.NotifyActivity(ActivityNotificationPhase.Started);
                 this.RunBackgroundTask(
```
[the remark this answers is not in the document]
YunchuWang: Validated locally. `NotifyActivity` is an instance method on `Processor`, not a local function inside `DispatchWorkItem`, so `this.NotifyActivity(...)` compiles. Evidence: `dotnet build src/Worker/Grpc/Worker.Grpc.csproj /property:GenerateFullPaths=true /consoleloggerparameters:NoSummary` succeeds. No code change needed.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
...
+                        {
+                            await this.OnRunActivityAsync(
+                                workItem.ActivityRequest,
+                                workItem.CompletionToken,
+                                cancellation).ConfigureAwait(false);
+                        }
+                        finally
+                        {
+                            this.NotifyActivity(ActivityNotificationPhase.Completed);
+                        }
```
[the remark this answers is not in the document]
YunchuWang: Validated locally. `NotifyActivity` is an instance method on `Processor`, not a local function inside `DispatchWorkItem`, so `this.NotifyActivity(...)` compiles. Evidence: `dotnet build src/Worker/Grpc/Worker.Grpc.csproj /property:GenerateFullPaths=true /consoleloggerparameters:NoSummary` succeeds. No code change needed.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
```diff
@@ -0,0 +1,262 @@
...
+    /// <summary>
+    /// Normalizes a worker profile ID and throws with the supplied message if it is missing.
+    /// </summary>
+    /// <param name="value">The worker profile ID value.</param>
+    /// <param name="errorMessage">The error message to use when the value is missing.</param>
+    /// <returns>The normalized worker profile ID.</returns>
+    internal static string NormalizeWorkerProfileId(string value, string errorMessage)
+    {
+        return SandboxActivityMetadata.NormalizeRequired(value, errorMessage);
+    }
```
halspang: Nit: Why an extra method when it just calls `NormalizeRequired` like the other one? It doesn't seem to do anything different.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// Options for the on-demand sandbox worker image DTS should start.
+    /// </summary>
+    public sealed class ImageOptions
+    {
+        /// <summary>
+        /// Gets or sets the full OCI container image reference for on-demand sandbox workers.
+        /// Examples: <c>myregistry.azurecr.io/workers/hello:1.0</c> or
+        /// <c>myregistry.azurecr.io/workers/hello@sha256:0123456789abcdef...</c>.
+        /// </summary>
+        public string? ImageRef { get; set; }
```
halspang: Nit: It seems like you have some required fields initialize to `string.Empty` and some are nullable. Is there a distinction? If not, I'd prefer us to be consistent. If it's required and we never expect it to be null, I'd prefer that to be reflected in the object.

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
```diff
@@ -0,0 +1,109 @@
...
+    public static string FormatActivity(Activity activity) =>
+        activity.Version is null ? activity.Name : $"{activity.Name}@{activity.Version}";
+
+    /// <summary>
+    /// Normalizes a worker profile ID.
+    /// </summary>
+    /// <param name="value">The worker profile ID.</param>
+    /// <param name="errorMessage">The exception message to use when the value is empty.</param>
+    /// <returns>The normalized worker profile ID.</returns>
+    public static string NormalizeWorkerProfileId(string value, string errorMessage)
```
halspang: Nit: Mentioned this elsewhere but this method seems extra given we just pass through to `NormalizeRequired`.

halspang (dismissed):
Just some nits, otherwise looks good!

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileBuilder.cs
YunchuWang: Fixed in 00d6247. Removed the extra `SandboxWorkerProfileBuilder.NormalizeWorkerProfileId` pass-through and call `NormalizeRequired` directly.

src/Client/AzureManaged.Sandboxes/SandboxWorkerProfileOptions.cs
YunchuWang: Fixed in 00d6247. Made the required string options non-nullable with `string.Empty` defaults to match the rest of the options object, while preserving validation.

src/Shared/AzureManaged.Sandboxes/SandboxActivityMetadata.cs
YunchuWang: Fixed in 00d6247. Removed the redundant `SandboxActivityMetadata.NormalizeWorkerProfileId` pass-through and updated callers to use `NormalizeRequired` directly.

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
             }
             else if (workItem.RequestCase == P.WorkItem.RequestOneofCase.ActivityRequest)
             {
+                this.NotifyActivity(ActivityNotificationPhase.Started);
```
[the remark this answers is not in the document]
YunchuWang: @copilot why do you see this is a local function, it is not. also it compiles, how do you get to this conclusion?

src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -405,12 +405,23 @@ void DispatchWorkItem(P.WorkItem workItem, CancellationToken cancellation)
...
+                        try
+                        {
+                            await this.OnRunActivityAsync(
+                                workItem.ActivityRequest,
+                                workItem.CompletionToken,
+                                cancellation).ConfigureAwait(false);
+                        }
+                        finally
+                        {
+                            this.NotifyActivity(ActivityNotificationPhase.Completed);
```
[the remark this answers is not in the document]
YunchuWang: ditto, not valid

halspang (approved):

PR #737, opened 2026-06-09
Add opt-in unversioned fallback for versioned task dispatch

torosent:
## The scenario

You're running a worker fleet for `OrderWorkflow`. Today you have one unversioned `OrderWorkflow` class registered. You want to roll out a new implementation — register `OrderWorkflow v2` alongside it — without taking the unversioned class out of service. Or you want one unversioned class to absorb any version you don't have a specific handler for yet, so older clients can keep starting work with `Version = ""`.

You configure `VersioningOptions` on the worker, register both classes (one versioned, one not, same name), and start dispatching.

## What's missing

Once `VersioningOptions` is set, the dispatcher only matches factories whose `(name, version)` exactly equals the work item's version. The unversioned `OrderWorkflow` registered for the same name is ignored. Requests with `Version = ""` (or any version you don't have a specific factory for) get rejected with `"No orchestrator task named 'OrderWorkflow' with version '...' was found."` There's no opt-in to say "fall back to the unversioned class."

The flip side is also missing. There's no explicit way to say "this worker ONLY serves exact `(name, version)` matches — reject everything else cleanly so a different pool picks it up." Today the strictness is implicit. And if you wire that intent up with `UseWorkItemFilters()`, the filter advertises every version the worker knows about, not just the ones it can actually serve, so the worker claims work items it then rejects.

## The change

`UnversionedFallbackMode` on `VersioningOptions`, opt-in per-side via `OrchestratorUnversionedFallback` and `ActivityUnversionedFallback`:

- `Implicit` (default) — fall back to the unversioned factory only when no versioned siblings exist for that name. Today's behavior for anything with versioned registrations.
- `CatchAll` — fall back to the unversioned factory whenever it's registered, even alongside versioned siblings.
- `StrictExactOnly` — never fall back. Missing `(name, version)` surfaces as `"No orchestrator task named '{name}' with version '{version}' was found."`

When `StrictExactOnly` is combined with `UseWorkItemFilters()`, the filter narrows per-name to advertise only versions this worker can actually serve. A DI validator throws `OptionsValidationException` at startup if a filter references a name with no registered factory.

Worker logger gets event IDs 606/607 (warning, emitted once when `CatchAll` is enabled) and 608/609 (debug, per dispatch when a fallback fires).

`samples/UnversionedFallbackSample` demos the new fallback end-to-end; `WorkerVersioningSample`, `ActivityVersioningSample`, and `EternalOrchestrationVersionMigrationSample` cover the companion versioning scenarios.


halspang:
I'm worried that we're over-complicating the versioning contract here. I think we can accomplish a lot of what this PR is attempting to accomplish without introducing any new concepts. Instead of the new options, we can leverage multiple workers. For example:

```csharp
// Worker A — drains old unversioned work
services.AddDurableTaskWorker("legacy", b => {
    b.AddTasks(r => r.AddOrchestrator<OrderWorkflow>());   // unversioned
    b.UseGrpc(...);
    // no UseVersioning  → serves unversioned instances
});

// Worker B — serves v2
services.AddDurableTaskWorker("v2", b => {
    b.AddTasks(r => r.AddOrchestrator<OrderWorkflowV2>());
    b.UseGrpc(...);
    b.UseVersioning(new() { Version = "2.0", MatchStrategy = Strict });
});
```
This keeps the versioning contract of a worker will never work on an unintended version much more solid and already integrates with the existing behavior in the code. We support named workers, registries, and options. You can also work this in with WorkItemFilters so the "fallback" worker can utilize either the base queue or still only work on named, unversioned orchestrations. 

What do you think of an approach like this, @torosent? Personally, I'd prefer to keep the versioning contract as simple as possible and this feels like a cleaner approach in that regard. I could also be missing something in the changes that have happened while I'm out though.

PR #751, opened 2026-06-18
Add multi-version support to [DurableTask] annotations

halspang:
# Summary
## What changed?
The Version property now accepts a comma-separated list (e.g. "v1,v2"), registering the type under each declared version and emitting call helpers that take a required 'version' parameter validated against the declared set. Single-version and unversioned behavior is unchanged.

## Why is this change needed?
This allows customers to specify multiple versions of a single orchestration/activity without having to duplicate the code in their own codebase.

## Issues / work items
- Resolves #
- Related #

---

# Project checklist
- [ ] Release notes are not required for the next release
- [ ] Breaking change?
  - [ ] If yes:
    - Impact:
    - Migration guidance:
---

- [ ] No
- [x] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)

If AI was used:
- Tool(s): Copilot
- AI-assisted areas/files: Mostly tests, some of the parsing logic
- What you changed after AI output: Filtering logic

- [ ] I verified referenced APIs/types exist and are correct
- [ ] I reviewed edge cases/failure paths (timeouts, retries, cancellation, exceptions)
- [ ] I reviewed concurrency/async behavior
- [ ] I checked for unintended breaking or behavior changes

---

# Testing
## Automated tests
- Result: Passed

- Environment (OS, .NET version, components):
- Steps + observed results:
  1. App with multiple versions declared
  2. Filters had correct versions
  3. Correct app served correct versions
- Evidence (optional):

---

# Notes for reviewers
- N/A

### DIFF
--- a/src/Abstractions/DurableTaskRegistry.cs
+++ b/src/Abstractions/DurableTaskRegistry.cs
@@ -161,4 +161,20 @@ public sealed partial class DurableTaskRegistry
         this.ActivitiesByVersion.Add(key, factory);
         return this;
     }
+
+    /// <summary>
+    /// Registers an activity factory under every supplied version. This is the shared fan-out used by the
+    /// type- and singleton-based registration overloads so that a class declaring multiple versions via
+    /// <see cref="DurableTaskAttribute.Version"/> is registered once per declared version.
+    /// </summary>
+    DurableTaskRegistry AddActivityAllVersions(
+        TaskName name, IReadOnlyList<TaskVersion> versions, Func<IServiceProvider, ITaskActivity> factory)
+    {
+        foreach (TaskVersion version in versions)
+        {
+            this.AddActivity(name, version, factory);
+        }
+
+        return this;
+    }
 }
--- a/src/Abstractions/TypeExtensions.cs
+++ b/src/Abstractions/TypeExtensions.cs
@@ -29,14 +29,70 @@ static class TypeExtensions
     /// </summary>
     /// <param name="type">The type to get the durable task version for.</param>
     /// <returns>The durable task version.</returns>
+    /// <remarks>
+    /// When the <see cref="DurableTaskAttribute.Version"/> declares multiple comma-separated versions, this
+    /// returns only the first declared version. Prefer <see cref="GetDurableTaskVersions"/> when all declared
+    /// versions are needed (for example, when registering a type under every version it supports).
+    /// </remarks>
     internal static TaskVersion GetDurableTaskVersion(this Type type)
     {
         // IMPORTANT: This logic needs to be kept consistent with the source generator logic.
         Check.NotNull(type);
-        return Attribute.GetCustomAttribute(type, typeof(DurableTaskAttribute)) switch
+        IReadOnlyList<TaskVersion> versions = type.GetDurableTaskVersions();
+        return versions.Count > 0 ? versions[0] : default;
+    }
+
+    /// <summary>
+    /// Gets every durable task version declared for a type via <see cref="DurableTaskAttribute.Version"/>.
+    /// </summary>
+    /// <param name="type">The type to get the durable task versions for.</param>
+    /// <returns>
+    /// The distinct (case-insensitive) set of declared versions in declaration order. An unversioned type
+    /// (no attribute, or an empty/unset <see cref="DurableTaskAttribute.Version"/>) yields a single
+    /// <see cref="TaskVersion.Unversioned"/> entry so callers can always register at least once.
+    /// </returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown when any comma-separated entry is whitespace-only. This mirrors the source generator's
+    /// <c>DURABLE3005</c> diagnostic so the reflection-based registration path fails closed for types whose
+    /// attribute the generator did not see.
+    /// </exception>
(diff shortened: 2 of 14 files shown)


src/Abstractions/TypeExtensions.cs
```diff
@@ -29,14 +29,70 @@ public static TaskName GetTaskName(this Type type)
...
+        List<TaskVersion> versions = new();
+        foreach (string segment in attr.Version!.Split(','))
+        {
+            if (segment.Length == 0)
+            {
+                // Truly-empty entry (e.g. a trailing or doubled comma). Skip silently.
+                continue;
+            }
+
+            string trimmed = segment.Trim();
```
berndverst, line 78: This trims a single version value that was previously preserved verbatim. For example, `[DurableTask(Version = " v1 ")]` used to register and stamp `" v1 "`; after an upgrade this registers as `"v1"`, so existing in-flight instances can no longer resolve to the task. Please preserve the exact value when no comma is present and apply normalization only to the new list syntax. The source-generator parser needs the same compatibility behavior.

src/Abstractions/DurableTaskRegistry.cs
```diff
@@ -161,4 +161,20 @@ DurableTaskRegistry AddActivity(TaskName name, TaskVersion version, Func<IServic
...
+    /// Registers an activity factory under every supplied version. This is the shared fan-out used by the
+    /// type- and singleton-based registration overloads so that a class declaring multiple versions via
+    /// <see cref="DurableTaskAttribute.Version"/> is registered once per declared version.
+    /// </summary>
+    DurableTaskRegistry AddActivityAllVersions(
+        TaskName name, IReadOnlyList<TaskVersion> versions, Func<IServiceProvider, ITaskActivity> factory)
+    {
+        foreach (TaskVersion version in versions)
+        {
+            this.AddActivity(name, version, factory);
```
berndverst, line 175: This fan-out mutates the registry as it goes. If `v2` is already registered and the attribute declares `v1,v2`, `v1` is added before `v2` throws, so a failed `AddActivity` call still changes dispatch state. Please preflight every key before adding any of them; the same atomicity fix is needed in `AddOrchestratorAllVersions`.

src/Generators/DurableTaskSourceGenerator.cs
```diff
@@ -856,8 +873,33 @@ static string GetStandaloneTaskRegistrationKey(string taskName, string taskVersi
...
+
         static string ToCSharpStringLiteral(string value) => SymbolDisplay.FormatLiteral(value, quote: true);
 
+        // Builds a runtime guard, emitted into a multi-version call helper, that rejects a caller-supplied
+        // version not declared on the task. The condition compares against each declared version literal so
+        // the wire only ever carries a version the task advertises.
+        static string BuildDeclaredVersionGuard(DurableTaskTypeInfo task, string callKind)
+        {
+            string condition = string.Join(
+                " || ", task.TaskVersions.Select(v => $"version == {ToCSharpStringLiteral(v)}"));
```
berndverst, line 891: These generated comparisons are case-sensitive, while `TaskVersion.Equals`, `TaskVersionKey`, parser deduplication, and `CompareVersions` all use case-insensitive semantics. The helper therefore rejects `"V1"` even though the runtime accepts it for a declared `"v1"`. Please emit `string.Equals(..., StringComparison.OrdinalIgnoreCase)` checks or canonicalize to the declared value.

src/Generators/DurableTaskSourceGenerator.cs
```diff
@@ -856,8 +873,33 @@ static string GetStandaloneTaskRegistrationKey(string taskName, string taskVersi
...
+            string declaredVersions = string.Join(", ", task.TaskVersions);
```
berndverst, line 892: `declaredVersions` is later inserted raw into the generated exception-message literal. A valid declaration such as `Version = "v1,v\"2"` produces malformed generated C# and fails the consumer build. Please construct and escape the complete message with `ToCSharpStringLiteral` (and keep raw version text out of generated literals).

berndverst (changes requested):
Requesting changes. The current head has six blocking issues:

- The existing unresolved comma-semantics thread is a source and behavioral/replay break: a value such as `"v1,v2"` used to identify one version, but now registers two versions and changes generated helper signatures. This needs explicit breaking-change approval and migration guidance, or an opt-in representation that preserves legacy values.
- Single-version values are now trimmed, changing existing version identity and potentially preventing in-flight instances from resolving after upgrade.
- The existing unresolved `CurrentOrOlder` filter thread is also a material latency/throughput issue: a post-filter empty list is interpreted as a wildcard, so the backend can stream work that the worker will reject.
- Multi-version registration is not atomic when a later version collides.
- Generated version validation is case-sensitive even though all runtime version identity/comparison paths are case-insensitive.
- Unescaped version text can produce invalid generated C#.

Compatibility verdict: no binary, protobuf, or JSON-format break was found, but there are definite source and behavioral/replay breaks in the shipped Abstractions and generated consumer API surfaces.

Performance verdict: parsing and registry fan-out are configuration-time operations and otherwise look reasonable. The wildcard filter behavior remains a real backend over-fetch and worker-rejection risk.

The PR also currently conflicts with `main` in the generator and changelog. When rebasing, preserve the nullable-input generator fix from #763 and revalidate the resulting diff.

PR #758, opened 2026-07-08
Add large-payload blob auto-purge with a persistent orchestration

YunchuWang:
## Summary

Large orchestration payloads are externalized to Azure Blob Storage by the `AzureBlobPayloads` extension, with a token persisted in SQL instead of the payload bytes. When orchestration state is purged, DTS cannot delete the backing blob itself: it has no customer storage credentials. The customer's worker performs that deletion.

This PR implements the worker/SDK side. The backend preserves cleanup work in a durable tombstone ledger; the worker fetches due tombstones, deletes eligible blobs, and reports every row's outcome so the backend can resolve, reschedule, or quarantine it. The backend prevents a tombstone from being fetched while a live payload row still references the same token.

Companion changes: contract in [microsoft/durabletask-protobuf#76](https://github.com/microsoft/durabletask-protobuf/pull/76), backend in AAPT-DTMB PR 16368738.

## gRPC contract and explicit opt-in

The contract is in `src/Grpc/durable-task-scheduler/large_payload_purge.proto`, under the separate `microsoft.durabletask.largepayloads.LargePayloadPurge` service:

| RPC | Caller and purpose |
|---|---|
| `SetLargePayloadAutoPurge` | An explicit Durable Task client call persists the task hub's enabled/disabled choice. |
| `GetLargePayloadTombstones` | The worker fetches a bounded batch of due cleanup records. |
| `ReportLargePayloadPurgeResults` | The worker reports each attempted cleanup outcome. |

These are not methods on `TaskHubSidecarService`. `GetWorkItemsRequest` has no auto-purge opt-in field, and worker connection does not announce or change this setting.

The existing refresh-protos.ps1 script refreshed all three SDK-vendored proto files from upstream main commit `882583c1177706e148c3b146d4b2ba1c1d242e4c`; commit `c2c6f83fe6ae4a38a0fb3c460d47390e3136880d` records the regenerated three-entry `versions.txt`. All three downloaded contract contents matched upstream and the prior SDK Git blobs. No generated proto was manually edited.

[...]

PR #766, opened 2026-07-16
Make externalized large-payload token self-describing (token v2)

YunchuWang:
## Summary

Makes the externalized large-payload token **self-describing** by embedding the storage account in it (token format **v2**). This is "Option A" for the dashboard-payload-access feature.

Today the token is `blob:v1:<container>:<blobName>` and does **not** carry the storage account — the account is implicit (whatever the worker's `PayloadStore` is currently configured with). That breaks two scenarios:

1. The DTS dashboard reads blobs **directly** from Storage using the signed-in user's own AAD token, so it needs the full blob URL — but the v1 token lacks the account.
2. If the customer repoints the store to a different account, old blobs (in the old account) become unresolvable because nothing records where each payload actually lives.

Fix: write the account into each token at upload time.

## Token format

- **v2 (new, written):** `blob:v2:<fullBlobUrl>` — everything after `blob:v2:` is one complete absolute blob URI, e.g.
  - `blob:v2:https://acct.blob.core.windows.net/mycontainer/<guid>` (DNS style)
  - `blob:v2:http://127.0.0.1:10000/devstoreaccount1/mycontainer/<guid>` (Azurite path style)
  - `BlobClient.Uri` carries no SAS/key, so the token is non-secret and safe to persist.
- **v1 (legacy, still read):** `blob:v1:<container>:<blobName>` — account implicit = configured store.

## Changes

- `BlobPayloadStore.EncodeToken(Uri)` emits v2 from the blob's absolute URI (`blob.Uri`) at upload time.
- `DecodeToken` parses both v2 (via `BlobUriBuilder`, handling DNS- and Azurite path-style URLs) and legacy v1.
- `IsKnownPayloadToken` recognizes both `blob:v1:` and `blob:v2:` so v2 tokens rehydrate.
- `DownloadAsync` honors the token's account:
  - **Same account/container** as the configured store → fast path via the existing container client (works with any auth).
  - **Different account** (post-repoint) → if identity auth (`TokenCredential`) is configured, read cross-account via a new `BlobClient` using the same retry options; otherwise throw a clear `PayloadStorageException` **before any network call**.
  - The download + gzip-decompress body is factored into a shared `DownloadFromBlobAsync` helper.
- Sample `LargePayloadConsoleApp` token check updated to also recognize `blob:v2:`.

## Back-compat

Preview product; breaking existing `blob:v1:` payloads would be acceptable, but v1 **read** back-compat is kept anyway (cheap, avoids breaking in-flight orchestrations).

## Caveat (by design)

Worker cross-account read only works with identity/managed-identity auth that has RBAC on both accounts. With connection-string/account-key auth the key is account-specific, so cross-account is impossible — the clear throw is the correct behavior. The dashboard (out of scope, different repo) is unaffected: it uses the user's own AAD token.

## Tests

Added `test/Grpc.IntegrationTests/BlobPayloadStoreTests.cs` (no live Azure required):
- EncodeToken → v2 token contains container + blob; round-trips through DecodeToken.
- DecodeToken parses v2 DNS-style **and** Azurite path-style URLs.
- DecodeToken still parses legacy v1.
- IsKnownPayloadToken true for v1 + v2, false for a plain string.
- Cross-account guard: v2 token for a different account with no identity credential throws `PayloadStorageException` before any network call.

Full solution builds with 0 errors and no new analyzer warnings; the 6 new tests pass.

## Coordination note — draft PR #758 (auto-purge)

> Read-only inspection; **not** modified here.

PR #758 (`yunchuwang-wangbill-blob-payload-autopurge-sdk`) adds `PayloadStore.DeleteAsync` + `BlobPayloadStore.DeleteAsync`, which still use the **v1-only** `DecodeToken`/`TokenPrefix` and a v1-only `IsKnownPayloadToken`. `DeleteExternalBlobActivity` classifies the `ArgumentException` that the v1 decoder throws on an unrecognized prefix as a **poison token → `Discarded`** (acked so the backend clears the tombstone row).

If #758 merges on top of this change without reconciliation, every **v2** token handed to `DeleteAsync` would fail the `blob:v1:` prefix check, throw `ArgumentException`, and be silently **discarded without deleting the blob** (payload leak) — and v2 tokens wouldn't be recognized as known. Whoever merges second should port #758's `DeleteAsync`/`DecodeToken`/`IsKnownPayloadToken` to the v2-aware versions (account-aware blob resolution, same fast-path/cross-account logic as `DownloadAsync`). There's also a mechanical merge conflict in `BlobPayloadStore.cs` since both branches edit the same token region.

---
Superseded/related (closed separately, **not touched here**): protobuf #77, this repo's #765, and a backend ADO PR.

### DIFF
--- a/src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
+++ b/src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
@@ -13,11 +13,14 @@ namespace Microsoft.DurableTask;
 
 /// <summary>
 /// Azure Blob Storage implementation of <see cref="PayloadStore"/>.
-/// Stores payloads as blobs and returns opaque tokens in the form "blob:v1:&lt;container&gt;:&lt;blobName&gt;".
+/// Stores payloads as blobs and returns self-describing opaque tokens in the form
+/// "blob:v2:&lt;fullBlobUrl&gt;", where the URL is the blob's absolute URI including the storage account.
+/// Legacy "blob:v1:&lt;container&gt;:&lt;blobName&gt;" tokens are still recognized for read back-compatibility.
 /// </summary>
 public sealed class BlobPayloadStore : PayloadStore
 {
-    const string TokenPrefix = "blob:v1:";
+    const string TokenPrefixV1 = "blob:v1:";
+    const string TokenPrefixV2 = "blob:v2:";
     const string ContentEncodingGzip = "gzip";
     const int MaxRetryAttempts = 8;
     const int BaseDelayMs = 250;
@@ -25,6 +28,7 @@ public sealed class BlobPayloadStore : PayloadStore
     const int NetworkTimeoutMinutes = 2;
     readonly BlobContainerClient containerClient;
     readonly LargePayloadStorageOptions options;
+    readonly BlobClientOptions clientOptions;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="BlobPayloadStore"/> class.
@@ -48,7 +52,7 @@ public sealed class BlobPayloadStore : PayloadStore
                 nameof(options));
         }
 
-        BlobClientOptions clientOptions = new()
+        this.clientOptions = new BlobClientOptions
         {
             Retry =
             {
@@ -61,8 +65,8 @@ public sealed class BlobPayloadStore : PayloadStore
         };
 
         BlobServiceClient serviceClient = hasIdentityAuth
-            ? new BlobServiceClient(options.AccountUri, options.Credential, clientOptions)
-            : new BlobServiceClient(options.ConnectionString, clientOptions);
+            ? new BlobServiceClient(options.AccountUri, options.Credential, this.clientOptions)
+            : new BlobServiceClient(options.ConnectionString, this.clientOptions);
 
         this.containerClient = serviceClient.GetBlobContainerClient(options.ContainerName);
     }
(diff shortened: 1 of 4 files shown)


src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -13,18 +13,22 @@ namespace Microsoft.DurableTask;
 
 /// <summary>
 /// Azure Blob Storage implementation of <see cref="PayloadStore"/>.
-/// Stores payloads as blobs and returns opaque tokens in the form "blob:v1:&lt;container&gt;:&lt;blobName&gt;".
+/// Stores payloads as blobs and returns self-describing opaque tokens in the form
+/// "blob:v2:&lt;fullBlobUrl&gt;", where the URL is the blob's absolute URI including the storage account.
+/// Legacy "blob:v1:&lt;container&gt;:&lt;blobName&gt;" tokens are still recognized for read back-compatibility.
```
halspang: Nit: Can you fix the encoding on these strings, comments should be human readable.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -153,7 +161,60 @@ public override bool IsKnownPayloadToken(string value)
...
+            || value.StartsWith(TokenPrefixV2, StringComparison.Ordinal);
+    }
+
+    /// <summary>
+    /// Encodes a self-describing v2 payload token from the blob's absolute URI. The token carries the full blob
+    /// URL (including the storage account) so readers can locate the payload without relying on the currently
+    /// configured store. <c>BlobClient.Uri</c> contains no SAS or account key and is safe to persist.
+    /// </summary>
+    /// <param name="blobUri">The absolute URI of the blob holding the payload.</param>
+    /// <returns>An opaque payload token in the form "blob:v2:&lt;fullBlobUrl&gt;".</returns>
```
halspang: Nit: Human readable.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -153,7 +161,60 @@ public override bool IsKnownPayloadToken(string value)
...
+    internal static string EncodeToken(Uri blobUri) => $"{TokenPrefixV2}{blobUri}";
+
+    /// <summary>
+    /// Decodes a payload token. Supports self-describing v2 tokens ("blob:v2:&lt;fullBlobUrl&gt;") and legacy v1
+    /// tokens ("blob:v1:&lt;container&gt;:&lt;blobName&gt;"), the latter for read back-compatibility.
```
halspang: More encoded strings. I'll stop commenting on them but please fix them everywhere.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -153,7 +161,60 @@ public override bool IsKnownPayloadToken(string value)
...
+    /// </summary>
+    /// <param name="token">The payload token to decode.</param>
+    /// <returns>
+    /// A tuple describing the token: whether it is v2, the container and blob names, and (for v2 only) the
+    /// absolute blob URI and its container-level URI.
+    /// </returns>
+    internal static (bool IsV2, string Container, string Name, Uri? BlobUri, Uri? ContainerUri) DecodeToken(
```
halspang: Nit: With 4 items in the return, I think we've gotten to the point of having a `DecodeTokenResult` or something.

halspang (commented):
Nothing major, mostly styling nits.

halspang (approved):

PR #785, opened 2026-07-24
Cache Azure Blob container initialization to avoid per-upload CreateIfNotExistsAsync

### LINKED ISSUE #771: Performance: cache Azure Blob payload container initialization
## 1. What is the issue?

Every externalized payload upload calls `CreateIfNotExistsAsync` before opening the blob for writing, even after the container has already been created successfully.

## 2. Likely priority

**High** for applications that use Azure Blob payload externalization frequently, especially fan-out workloads with many payload fields above the externalization threshold.

## 3. Impact on performance

Each upload incurs an additional Azure Storage request and billable transaction before the actual blob write. This increases end-to-end latency and can roughly double storage request volume for the upload path.

Payload upload implementation:

https://github.com/microsoft/durabletask-dotnet/blob/883211a3a7ab0b55bb90b9c9e498148e4d65ead9/src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs#L71-L109

`UploadAsync` calls `CreateIfNotExistsAsync` for every upload at line 80.

Payload store registration:

https://github.com/microsoft/durabletask-dotnet/blob/883211a3a7ab0b55bb90b9c9e498148e4d65ead9/src/Extensions/AzureBlobPayloads/DependencyInjection/ServiceCollectionExtensions.AzureBlobPayloads.cs#L33-L41

The payload store is registered once as a singleton, so successful initialization can be shared by all uploads from that service provider.

## 5. Guidance for how to fix

Use a single-flight cached initialization task or an interlocked initialization state so concurrent first uploads issue at most one create/check request. Keep failure retryable, and consider resetting the initialized state after a container-not-found write failure so deliberate container deletion remains recoverable. Preserve cancellation and storage error propagation; this runs in transport infrastructure and does not affect orchestrator replay determinism.

berndverst:
## Summary

`BlobPayloadStore.UploadAsync` previously called `CreateIfNotExistsAsync` before every upload, adding an unnecessary storage transaction after the payload container was already known to exist.

This change caches a successful container-initialization generation for each `BlobPayloadStore` instance. Existing DI registrations create the payload store as a singleton, so the cached state is reused across uploads.

## Implementation

- **Steady-state fast path**: a volatile read returns the cached generation and skips `CreateIfNotExistsAsync`.
- **Concurrency-safe first use**: when no generation is cached, callers enter a `SemaphoreSlim` asynchronously, double-check the cache, and only the lock holder calls `CreateIfNotExistsAsync`.
- **Caller-owned cancellation**: the caller token is used both while waiting for the semaphore and by the Azure Storage create request. Cancelling one caller does not cancel another; a waiting caller can retry after a cancelled initializer releases the gate.
- **Retriable initialization failures**: failed or cancelled initialization publishes no generation, so a later caller performs a fresh attempt without explicit fault eviction or a shared background task.
- **Generation-safe deletion recovery**: each successful initialization publishes a unique marker. If a write returns `ContainerNotFound`, the upload clears the cache only when it still contains the marker that upload used, then retries once. A stale failure cannot invalidate a newer initialization.

If Azure is still processing a container deletion, the retrying create can return `409 ContainerBeingDeleted`; that error propagates until deletion settles, matching the behavior before initialization was cached.

No public API changes are introduced. Main's self-describing v2 payload tokens and cross-account read behavior remain unchanged.

## Testing

- Focused unit coverage for concurrent first use, initialization failure retry, lock-holder cancellation, waiting-caller cancellation, settled deletion recovery, unrelated storage errors, and stale-generation invalidation.
- `AzureBlobPayloads.Tests`: **7/7 passing**, including with `DOTNET_PROCESSOR_COUNT=2`.
- Existing v2 token coverage in `Grpc.IntegrationTests`: **6/6 passing**.
- `AzureBlobPayloads.csproj` builds for `netstandard2.0`, `net6.0`, `net8.0`, and `net10.0`.
- Direct Azurite smoke verification: two uploads and downloads round-tripped successfully with **exactly one** container-create request in the request log.
- The full `LargePayloadConsoleApp` DTS scenario was not run because `DURABLE_TASK_SCHEDULER_CONNECTION_STRING` was not available in this environment.

This change remains independent of other performance work.

Fixes #771


src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -195,4 +253,56 @@ static async Task<string> ReadToEndAsync(StreamReader reader, CancellationToken
...
+    /// Starts container initialization if it hasn't already started, in a single-flight manner:
+    /// only the first caller's task is stored and returned to all callers, including ones racing
+    /// concurrently on other threads.
+    /// </summary>
+    Task BeginContainerInitialization()
+    {
+        Task newInitializationTask = this.CreateContainerIfNotExistsAsync();
+        return Interlocked.CompareExchange(ref this.containerInitializationTask, newInitializationTask, null)
+            ?? newInitializationTask;
+    }
```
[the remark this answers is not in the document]
berndverst: Fixed in 46d028a. Initialization is now a Lazy<Task> published with Interlocked.CompareExchange before Value starts work, so only the winning gate invokes CreateIfNotExistsAsync.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
```diff
@@ -0,0 +1,228 @@
...
+                return (Response<BlobContainerInfo>)null!;
+            });
+
+        BlobPayloadStore store = new(new LargePayloadStorageOptions(), containerClientMock.Object);
+
+        // Act: fire many concurrent uploads against a fresh (uninitialized) store.
+        Task<string>[] uploads = Enumerable.Range(0, 10)
+            .Select(_ => store.UploadAsync("payload", CancellationToken.None))
+            .ToArray();
+        await Task.WhenAll(uploads);
```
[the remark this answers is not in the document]
berndverst: Fixed in 46d028a. The test now starts 16 Task.Run workers behind a Barrier, forcing concurrent entry before asserting a single initialization call.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
```diff
@@ -0,0 +1,511 @@
...
+using Microsoft.DurableTask;
+
+namespace Microsoft.DurableTask.Extensions.AzureBlobPayloads.Tests;
+
+/// <summary>
+/// Unit tests for <see cref="BlobPayloadStore"/>'s container-initialization caching behavior
+/// (see https://github.com/microsoft/durabletask-dotnet/issues/771).
+/// </summary>
+public class BlobPayloadStoreTests
+{
```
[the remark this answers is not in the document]
berndverst: Updated the PR description. It no longer claims five tests; the Testing section now enumerates the focused coverage without a stale count.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -76,33 +149,50 @@ public override async Task<string> UploadAsync(string payLoad, CancellationToken
...
-            await WritePayloadAsync(payloadBuffer, blobStream, cancellationToken);
-            await blobStream.FlushAsync(cancellationToken);
+            // The container existed when we last verified/created it but has since been deleted
+            // (e.g. by an operator). Clear the cached initializer so the next upload attempts to
+            // recreate the container, keeping deliberate deletion recoverable. CompareExchange
+            // against the specific initializer this upload used ensures a stale failure can
+            // never clobber a newer initializer already published by another, faster-recovering
+            // concurrent upload that detected and recovered from the same deletion.
+            _ = Interlocked.CompareExchange(ref this.containerInitializer, null, containerInitializer);
+            throw;
```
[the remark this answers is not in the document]
berndverst: Fixed in b4f8306. On ContainerNotFound, the upload atomically invalidates its initializer generation, recreates the container, and retries that same write once.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -156,6 +194,35 @@ public override bool IsKnownPayloadToken(string value)
...
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+        }
+
+        await initializationTask.ConfigureAwait(false);
+#else
+        await initializationTask.WaitAsync(cancellationToken).ConfigureAwait(false);
+#endif
+    }
```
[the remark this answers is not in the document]
berndverst: Fixed in 2be3fb1. The winning initializer attaches one fault-only observer when published, so a later fault is observed even if every waiter cancels.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -195,4 +278,99 @@ static async Task<string> ReadToEndAsync(StreamReader reader, CancellationToken
...
+    /// garbage-collected. Also invokes <see cref="OnInitializationFaultObserved"/> (a no-op in
+    /// production) so tests can deterministically confirm this continuation ran.
+    /// </summary>
+    void ObserveFaultWithoutAwaiting(Task task)
+    {
+        _ = task.ContinueWith(
+            t => this.OnInitializationFaultObserved?.Invoke(t.Exception!),
+            CancellationToken.None,
+            TaskContinuationOptions.OnlyOnFaulted | TaskContinuationOptions.ExecuteSynchronously,
+            TaskScheduler.Default);
```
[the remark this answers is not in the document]
berndverst: Fixed in 3513558. The observer hook is guaranteed non-null and invoked directly, so evaluating t.Exception is unconditional in production as well as tests.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -195,4 +324,102 @@ static async Task<string> ReadToEndAsync(StreamReader reader, CancellationToken
...
+    /// stale error. The CompareExchange only clears the cache if it still points at this same
+    /// gate (<paramref name="self"/>), so this can never erase a newer initializer already
+    /// published by a racing caller (e.g. one that recovered from a concurrently-deleted
+    /// container) after this attempt started.
+    /// </summary>
+    async Task CreateContainerIfNotExistsAsync(Lazy<Task> self)
+    {
+        try
+        {
+            await this.containerClient.CreateIfNotExistsAsync(PublicAccessType.None, cancellationToken: CancellationToken.None)
```
YunchuWang: **[behavior regression] The shared initializer is now uncancellable by anything.**

Before this PR, `CreateIfNotExistsAsync` received the caller's token. It now receives `CancellationToken.None`, with no other bound.

Combined with the retry policy configured in the primary constructor (`MaxRetries = 8`, `MaxDelay = 10s`, `NetworkTimeout = 2min` — lines 22-25), a create against a throttled or unreachable account can stay in flight for **minutes** and will not observe host shutdown. Every upload arriving in that window attaches to it via `EnsureContainerExistsAsync` rather than failing fast.

I understand *why* it is `None` — the task is shared, so no single caller's token may cancel it for everyone. But "no caller's token" should not have to mean "no bound at all":

```csharp
using CancellationTokenSource cts = new(TimeSpan.FromSeconds(30));
await this.containerClient
    .CreateIfNotExistsAsync(PublicAccessType.None, cancellationToken: cts.Token)
    .ConfigureAwait(false);
```

This is the one item here that is independent of the design discussion below — worth fixing either way.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -195,4 +324,102 @@ static async Task<string> ReadToEndAsync(StreamReader reader, CancellationToken
...
+    /// which starts the real container-creation call - is invoked exactly once even when
+    /// multiple callers race to access <see cref="Lazy{T}.Value"/> concurrently. The winning
+    /// publisher also attaches a single fault-observer (see
+    /// <see cref="ObserveFaultWithoutAwaiting(Task)"/>) to the shared task, once, up front -
+    /// rather than leaving each individual caller responsible for observing failures on its own
+    /// cancellation path - so a shared initialization failure is always observed regardless of
+    /// how many (if any) callers are still waiting on it when it faults, on every target
+    /// framework this library supports.
+    /// </summary>
+    Lazy<Task> PublishNewInitializer()
```
YunchuWang: **[design] Almost all of the complexity here is downstream of one choice: sharing a `Task` between callers.**

To be clear up front: the current code is correct. I traced the racing-publisher, stale-invalidation and fault-self-heal interleavings and could not break it. This is about long-term maintenance cost, not correctness.

Sharing the initialization `Task` is what creates *all* of the following: `WaitForInitializationAsync` and its hand-rolled `#if NETSTANDARD2_0` fork, the unobserved-fault hazard and `ObserveFaultWithoutAwaiting`, the `internal` test hook, `NoOpFaultObserver`, and the "self-heal from the task's own completion path" requirement.

By my count, **seven of the eleven commits on this branch exist only because the task is shared** (`46d028a`, `79b1d06`, `2be3fb1`, `4102547`, `4529c66`, `3513558`, `a7a6d37`). Only `b4f8306` (container-deletion recovery) is inherent to caching at all.

The root cause is not "shared state" — it is that **the shared thing has no owner**. An in-flight async operation belonging to nobody must have its ownership simulated: it has no token (hence `None`), nobody will observe its exception (hence the observer hook), and it caches a *result*, so a failed result must be explicitly evicted (hence the CAS self-heal).

Sharing an *access right* instead of a *result* makes all of that not arise:

```csharp
readonly SemaphoreSlim initLock = new(1, 1);

// 0 = container not known to exist. A positive value identifies the create that verified
// it, so a stale write failure can only invalidate the generation it actually used.
int containerEpoch;
int lastEpoch;   // only mutated while holding initLock

async Task<int> EnsureContainerExistsAsync(CancellationToken ct)
{
    int epoch = Volatile.Read(ref this.containerEpoch);
    if (epoch != 0)
    {
        return epoch;
    }

    await this.initLock.WaitAsync(ct).ConfigureAwait(false);
    try
    {
        epoch = Volatile.Read(ref this.containerEpoch);
        if (epoch != 0)
        {
            return epoch;
        }

        await this.containerClient
            .CreateIfNotExistsAsync(PublicAccessType.None, cancellationToken: ct)
            .ConfigureAwait(false);

        epoch = ++this.lastEpoch;   // monotonic, never reused, never 0
        Volatile.Write(ref this.containerEpoch, epoch);
        return epoch;
    }
    finally
    {
        this.initLock.Release();
    }
}
```

and in `UploadAsync`'s `ContainerNotFound` handler, structurally identical to what is there today:

```csharp
Interlocked.CompareExchange(ref this.containerEpoch, 0, epoch);
```

(The monotonic `lastEpoch` matters — if epoch numbers were reused, a stale CAS could clobber a newer generation.)

Per scenario:

| Scenario | Current | Alternative |
|---|---|---|
| Cold-start burst, N callers | exactly 1 create | exactly 1 create (double-checked) |
| Steady state | volatile read + `Lazy.Value` + `WaitAsync` fast path | volatile read + int compare |
| Init fails | needs the catch + CAS-with-self to evict the cached failure | epoch was never written — **self-heals with zero code** |
| Caller cancellation | needs `WaitForInitializationAsync` + polyfill; SDK forced to `None` | `WaitAsync(ct)`; **token reaches the SDK** |
| Container deleted | CAS on the `Lazy` instance | CAS on the epoch — identical |
| Unobserved fault | needs observer + hook + no-op delegate + 2 tests | **impossible by construction** |
| netstandard2.0 | hand-rolled polyfill, currently untested | `SemaphoreSlim.WaitAsync` is native on all TFMs |
| Cold-start thread use | `Lazy.Value` (ExecutionAndPublication) blocks contending threads on a monitor | fully async, no thread blocking |

Roughly 35 lines of mechanism instead of ~100, and ~8 lines of comment instead of ~150. It also resolves the cancellation issue above for free.

**Where the current design is genuinely better** — I do not want to present this as one-sided:

1. **Failure coalescing.** Under a sustained outage the shared task fails every waiter at once from a single attempt. The semaphore version serializes one attempt per caller, and the SDK's own retry policy compounds that. Partly mitigated by each caller's token actually working in the alternative, and bounded by the observation that if create is failing then `OpenWriteAsync` would be failing too — so the worst case degrades to *exactly the pre-PR behavior*, which is what ships today. But it is a real difference.
2. **A cancelling caller does not abort the in-flight create.** In the alternative, if the lock holder's token fires mid-create the create really is cancelled and the next caller redoes it — `UploadAsync_CancelledCaller_...` would become `Times.Exactly(2)`. That is the unavoidable flip side of honoring the caller's token; you cannot have both.
3. Minor: `SemaphoreSlim` is `IDisposable` on a type that is not, which may trip CA1001 (in practice it needs no disposal unless `AvailableWaitHandle` is touched).

So the question I would put to you: is the thing being protected worth the machinery protecting it? `CreateIfNotExists` is idempotent, cheap, and happens once per process; if a cancelling caller aborts it, recovery costs one more equally cheap idempotent call.

Entirely your call — this is already validated, and rewriting means paying the verification cost again. But if this code is going to be maintained for a while, the delta seems worth one deliberate decision rather than inheriting by default.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -67,6 +91,55 @@ public BlobPayloadStore(LargePayloadStorageOptions options)
...
+    /// set a non-null hook first and so could never observe the regression. With this hook
+    /// guaranteed non-null, that class of bug cannot recur regardless of how the invocation is
+    /// written at the call site, and a test that overrides this hook is exercising the exact same
+    /// code as the unmodified production default - just substituting a different delegate
+    /// instance for <see cref="NoOpFaultObserver"/>.
+    /// </para>
+    /// Each test uses its own <see cref="BlobPayloadStore"/> instance, so no reset between tests
+    /// is needed. Callers of this hook must not assume any particular thread.
+    /// </summary>
+    internal Action<Exception> OnInitializationFaultObserved
```
YunchuWang: **[test seam] This hook exists purely to make an assertion possible.**

It brings a settable `internal` property, a `static readonly NoOpFaultObserver`, a null-normalizing setter, ~30 lines of XML doc explaining why it cannot be null, and `OnInitializationFaultObserved_DefaultsToNonNullNoOpAndRejectsNull` — a test that asserts the *hook's* null-handling rather than any product behavior.

The hazard it guards against only exists because the initialization task is shared (see the comment on `PublishNewInitializer`); removing the sharing removes the hook, the delegate, and both associated tests.

If the current design stays, the hook is defensible — but I would still drop that one test. The behavior that actually matters is already covered end-to-end by `UploadAsync_AllWaitersCancelThenInitializerFaults_ExceptionIsObserved`.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -67,6 +91,55 @@ public BlobPayloadStore(LargePayloadStorageOptions options)
...
+    /// was <see langword="null"/> by default), even though every test that exercised it happened to
+    /// set a non-null hook first and so could never observe the regression. With this hook
+    /// guaranteed non-null, that class of bug cannot recur regardless of how the invocation is
+    /// written at the call site, and a test that overrides this hook is exercising the exact same
+    /// code as the unmodified production default - just substituting a different delegate
+    /// instance for <see cref="NoOpFaultObserver"/>.
+    /// </para>
+    /// Each test uses its own <see cref="BlobPayloadStore"/> instance, so no reset between tests
+    /// is needed. Callers of this hook must not assume any particular thread.
+    /// </summary>
```
YunchuWang: **[docs] These comments document the PR's own review history rather than the code.**

Samples from what ships here and elsewhere in the file: *"the historical bug where..."*, *"which is exactly how the historical `this.OnInitializationFaultObserved?.Invoke(t.Exception!)` regression passed every existing test"*, *"eliminates — by construction, not merely by test coverage —"*. The tests have the same pattern, explaining why an earlier version of the *test* was wrong.

Roughly 150 of the 251 added production lines are comments. That narrative is genuinely valuable — in the review thread and the commit messages, where it already lives. In source it rots on the first refactor and buries the few lines a reader actually needs.

Suggest trimming each block to the invariant that holds today. For example, for `PublishNewInitializer`:

> Publishes a not-yet-started initialization gate. The `Lazy<Task>` is published via `CompareExchange` **before** anything accesses `.Value`, so racing callers cannot each start their own `CreateIfNotExistsAsync`. The winning publisher attaches the fault observer exactly once.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -156,6 +256,35 @@ public override bool IsKnownPayloadToken(string value)
         return value.StartsWith(TokenPrefix, StringComparison.Ordinal);
     }
 
+    /// <summary>
+    /// Awaits a shared initialization task while still honoring the caller's own cancellation
+    /// token, without cancelling the shared task itself.
+    /// </summary>
+    static async Task WaitForInitializationAsync(Task initializationTask, CancellationToken cancellationToken)
```
YunchuWang: **[coverage] The netstandard2.0 branch ships untested.**

The test project targets `net10.0` only, so every test exercises the `Task.WaitAsync` path. The hand-rolled `WhenAny` + `CancellationTokenRegistration` polyfill — the subtler of the two implementations, and the one with the manual fault-observation requirement — has zero coverage but is what ships to netstandard2.0 consumers.

Options, cheapest first: extract the polyfill into an internal helper testable on any TFM; or multi-target the test project (note `net8.0` will not help, since the branch is gated on netstandard2.0 specifically); or remove the fork entirely per the `PublishNewInitializer` comment — `SemaphoreSlim.WaitAsync(ct)` exists on netstandard2.0.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
```diff
@@ -76,36 +149,63 @@ public override async Task<string> UploadAsync(string payLoad, CancellationToken
...
+        // Retry one write after an out-of-band container deletion so the cached path preserves
+        // the recovery behavior that an unconditional CreateIfNotExistsAsync provided before
+        // initialization was cached.
+        bool retryAfterContainerNotFound = true;
+        while (true)
         {
-            BlobOpenWriteOptions writeOptions = new()
+            // Keep the specific initializer instance this upload used so ContainerNotFound
+            // recovery below can invalidate it precisely (see the catch block).
+            Lazy<Task> containerInitializer = await this.EnsureContainerExistsAsync(cancellationToken).ConfigureAwait(false);
```
YunchuWang: **[nit]** This local shadows the field of the same name, and both appear together on line 202:

```csharp
_ = Interlocked.CompareExchange(ref this.containerInitializer, null, containerInitializer);
```

Renaming the local to `initializer` makes that line self-explanatory about which is the target and which is the comparand.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
```diff
@@ -0,0 +1,510 @@
...
+
+        // Act: force genuinely concurrent entry into the single-flight logic. Each worker runs
+        // on its own thread-pool thread (via Task.Run) and blocks at a Barrier until every
+        // worker has arrived, so they all call UploadAsync at (as close to) the same instant as
+        // possible. A simple sequential LINQ + Task.WhenAll loop would not exercise this: the
+        // synchronous prefix of each call (up to the first real await) would run one after
+        // another on the calling thread, letting the first caller publish the cached
+        // initializer before any other caller's code even starts, which trivially "passes" even
+        // a buggy, non-single-flight implementation.
+        using Barrier barrier = new(WorkerCount);
```
YunchuWang: **[test cost] `Barrier` blocks thread-pool threads — measured 11s on a small runner.**

I ran this with `DOTNET_PROCESSOR_COUNT` pinned:

| Cores | Duration |
|---|---|
| 32 | ~0.1s |
| 2 | **11s** |

`SignalAndWait()` blocks its thread, so the pool must hill-climb ~12 extra threads at roughly one per second before the barrier releases. CI is `windows-latest` (4 vCPU), so this single test will dominate the project's suite. It is not flaky — 5/5 full-suite runs passed at 2 cores — just expensive.

An async gate gets the same concurrent entry without blocking:

```csharp
TaskCompletionSource gate = new(TaskCreationOptions.RunContinuationsAsynchronously);
int ready = 0;
// in each worker:
if (Interlocked.Increment(ref ready) == WorkerCount) gate.SetResult();
await gate.Task;
return await store.UploadAsync("payload", CancellationToken.None);
```

(The `await Task.Delay(50)` inside the create mock on line 33 is doing real work — it holds initialization in flight so the other workers pile up. That one should stay.)

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
```diff
@@ -0,0 +1,510 @@
...
+        BlobPayloadStore store = new(new LargePayloadStorageOptions(), containerClientMock.Object);
+
+        using CancellationTokenSource cts = new();
+        Task<string> cancelledUpload = store.UploadAsync("payload", cts.Token);
+        Task<string> otherUpload = store.UploadAsync("payload", CancellationToken.None);
+
+        // Act: cancel the first caller while initialization is still in flight, then let
+        // initialization complete successfully for everyone else.
+        cts.Cancel();
+        await Task.Delay(50);
```
YunchuWang: **[nit]** This delay, and the `await Task.Delay(20)` on line 469, look like dead weight.

`UploadAsync` runs synchronously through to the pending `OpenWriteAsync` (the create mock completes synchronously and `WaitAsync(default)` takes the completed-task fast path), and `cts.Cancel()` runs its registrations synchronously before returning — so both orderings are already deterministic without the sleep.

If either one *is* load-bearing, that points at a race the test is not actually controlling, which is worth making explicit rather than sleeping on.

YunchuWang (commented):
Verified locally on this branch: builds clean, **9/9 tests pass**, no new analyzer warnings on `BlobPayloadStore.cs`, and the modified `.sln` restores cleanly (project GUID, all six configuration mappings, and the nested-project entries are all correct).

The optimization is right and the win is real. On a worker running the default `MaximumConcurrentActivityWorkItems = 100 * ProcessorCount`, this roughly halves requests on the upload path — and more importantly, a fan-out orchestrator reply that externalizes 500 action inputs goes from 500 redundant container calls to zero, since `ExternalizeRequestPayloadsAsync` walks actions sequentially with no `Task.WhenAll`. That is the #771 case and it needs no concurrency at all to hurt.

Things I specifically checked and found correct:

- **Publish-before-start** — `PublishNewInitializer` completes the `CompareExchange` before anything touches `.Value`, with `LazyThreadSafetyMode.ExecutionAndPublication` as a backstop. This properly fixes the original "an async method starts on call, not on await" flaw.
- **Generation-precise invalidation** — using the initializer instance as the CAS comparand in both the `ContainerNotFound` handler and the failure self-heal. I traced racing publishers, stale invalidation, and the fault self-heal path and could not construct an interleaving that produces a redundant create or clobbers a newer initializer.
- **Retry scope** — the `try` covers `OpenWriteAsync`, the writes, `FlushAsync` *and* the `using` disposal, so `ContainerNotFound` is caught wherever the SDK surfaces it, and `retryAfterContainerNotFound` correctly bounds it to one attempt.
- **Singleton registration** — `AddSingleton<PayloadStore>` in both DI extensions, so the cache is actually effective. Worth noting this is a new, unstated coupling: `BlobPayloadStore` is now stateful and only pays off as a singleton.
- **`DownloadAsync` / `IsKnownPayloadToken`** — untouched, correctly so.

One item below (the uncancellable initializer) I would like addressed before merge. The design comment is a recommendation rather than a blocker — I have included the places where the current approach is genuinely better than the alternative I sketch.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
```diff
@@ -0,0 +1,510 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Azure;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Microsoft.DurableTask;
```
YunchuWang: **[nit]** Redundant — the enclosing namespace already resolves `BlobPayloadStore`.

test/Extensions/AzureBlobPayloads.Tests/AzureBlobPayloads.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <AssemblyName>Microsoft.DurableTask.Extensions.AzureBlobPayloads.Tests</AssemblyName>
```
YunchuWang, line 6: **[maintainability] This line is load-bearing but does not look it.**

`src/Directory.Build.targets` grants `InternalsVisibleTo Include="$(AssemblyName).Tests"`, which resolves to `Microsoft.DurableTask.Extensions.AzureBlobPayloads.Tests` only because of this override. Without it, the root `Directory.Build.props` would compute `Microsoft.DurableTask.AzureBlobPayloads.Tests` (no `.Extensions`), the IVT would not match, and the project would fail to compile against the `internal` constructor. Worth a one-line comment so a future cleanup does not delete it.

Two related nits in this file: `<PackageReference Include="Azure.Storage.Blobs" />` on line 9 is redundant (it flows through the `ProjectReference`), and `RootNamespace` still resolves to `Microsoft.DurableTask.AzureBlobPayloads.Tests`, which does not match the namespace declared in `BlobPayloadStoreTests.cs` — harmless but inconsistent.

YunchuWang:
### Follow-up: end-to-end verification against real storage

Separate from the inline comments — this is about test *kind* rather than test *content*, so it did not fit on a line.

As far as I can tell from the PR description and discussion, this has not been exercised against real (or emulated) storage. The Testing section lists focused unit coverage, and the strongest runtime claim is that the project *builds* across all target frameworks. CI (`validate-build.yml`) runs `dotnet test` on `windows-latest` with no Azurite service and no storage configuration, so there is no integration coverage there either. Apologies if I have missed a manual run that happened outside the PR — if so, please ignore this.

Worth noting this is also the first test project this extension has ever had, which is a genuine improvement. The observation below is only that all eight tests use `Mock<BlobContainerClient>` and assert call counts on it — which validates the coordination logic well (I traced it and could not break it), but by construction cannot validate anything where the mock's behavior diverges from real Azure.

#### One concrete instance of that divergence

`UploadAsync_ContainerDeletedAfterInitialization_ResetsCacheAndRecreatesContainer` passes because the mocked re-create returns success immediately. Real Azure does not behave that way.

From the [Delete Container REST docs](https://learn.microsoft.com/rest/api/storageservices/delete-container):

> When a container is deleted, a container with the same name can't be created for at least 30 seconds... While the container is being deleted, attempts to create a container of the same name fail with status code 409 (Conflict).

And `BlobContainerClient.CreateIfNotExistsInternal` in `Azure.Storage.Blobs` 12.27.0 (the version pinned in `Directory.Packages.props`) swallows only one error code:

```csharp
catch (RequestFailedException storageRequestFailedException)
when (storageRequestFailedException.ErrorCode == BlobErrorCode.ContainerAlreadyExists)
```

`ContainerBeingDeleted` is not swallowed, and 409 is not in Azure.Core's retriable set (408/429/5xx). So against real storage:

```
container deleted by operator
  -> OpenWriteAsync            -> 404 ContainerNotFound
  -> cache invalidated, retryAfterContainerNotFound = false, continue
  -> CreateIfNotExistsAsync    -> 409 ContainerBeingDeleted     <-- mock returns success here
  -> not swallowed by the SDK, not retried by the pipeline
  -> throws, and the one-shot retry budget is already spent -> upload fails
```

To be fair about what this is and is not:

- **Not a regression.** Before this PR every upload called `CreateIfNotExists`, so it would hit the same 409. Behavior really is preserved, as the description says.
- **It does work** once the deletion has fully settled (roughly 30s+), which is the path the unit test models.
- The gap is that the commit is titled "Retry upload after container deletion" and the description says it "recreates the container, and retries that same upload once" — and in the timing most likely to occur in practice (operator deletes the container, traffic arrives immediately after) it will not. A mock cannot distinguish those two worlds.

I am not suggesting this needs to be *fixed* here — handling `ContainerBeingDeleted` properly means waiting out a 30s window, which is arguably out of scope. Mostly I would not want the test name and commit title to imply a guarantee that does not hold, so a comment noting the limitation would be enough.

#### Two smaller gaps

- The central premise — container already exists, so skipping `CreateIfNotExists` still leaves `OpenWriteAsync` working — is the entire point of the change and is only ever exercised against a mock.
- The request-count reduction that motivates the PR has not been measured; it currently exists as an argument rather than an observation.

#### Why this is a cheap ask

`samples/LargePayloadConsoleApp` already exists and is already in the solution, and is more or less exactly this harness — its README describes externalizing large payloads to blob storage over Durable Task Scheduler with no sidecar, defaulting to Azurite. It sets `ThresholdBytes = 1024` so everything externalizes, and covers large orchestration input, activity I/O, sub-orchestrations, external events, custom status, entity input/output/state, oversized rejection, and `ThreeLargeActivities` — three parallel 13MB activities, which is the fan-out shape this PR is optimizing.

`dotnet run` against Azurite plus a DTS connection string is maybe ten minutes. Given this extension ships against real DTS and real storage, and this PR changes the write path for every externalized payload, that seems proportionate — and it would let you attach a measured before/after transaction count, which would make the case for the change stronger than any of the unit tests can.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: Addressed in `3f56caa`. I replaced the ownerless shared task with a `SemaphoreSlim`-protected generation. The caller performing initialization now passes its token directly to `CreateIfNotExistsAsync`, while waiting callers cancel their own `WaitAsync`. Separate tests cover cancellation of the initializer and of a waiter; no arbitrary timeout was added to production behavior.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: I adopted this ownership model in `3f56caa`, using a unique object token rather than a numeric epoch so generations cannot wrap or be reused. The steady-state path is a volatile read; the cold path double-checks under `SemaphoreSlim`; stale deletion failures CAS against the exact token they used. This removed the shared task, failure eviction, fault observer, test hook, and target-framework cancellation fork.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: Removed. The observer hook, no-op delegate, null-normalizing property, and both hook-specific tests disappeared with the shared-task design.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: Trimmed as part of the redesign. The remaining comments describe only current invariants: the cached generation, one retry after `ContainerNotFound`, and the settled-deletion limitation.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: The framework-specific wait implementation is gone. Initialization now uses `SemaphoreSlim.WaitAsync(cancellationToken)`, available on every shipped TFM, and the library builds successfully for `netstandard2.0`, `net6.0`, `net8.0`, and `net10.0`.

src/Extensions/AzureBlobPayloads/PayloadStore/BlobPayloadStore.cs
berndverst: Resolved by the redesign. The upload retains a local named `generation`, while the target field remains `this.containerGeneration`, making the generation-aware CAS explicit.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
berndverst: Replaced `Barrier` with a `TaskCompletionSource` async start gate. The delayed create remains to hold initialization in flight. The complete seven-test suite now reports about 0.44s test duration with `DOTNET_PROCESSOR_COUNT=2`.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
berndverst: Removed both ordering sleeps. Cancellation is coordinated by the cancellation tokens themselves, and the stale-generation test relies on the already-pending `OpenWriteAsync` task. The only delays left in these tests are the intentional 50ms in-flight create and 5s diagnostic timeout guards.

test/Extensions/AzureBlobPayloads.Tests/BlobPayloadStoreTests.cs
berndverst: Removed the redundant namespace import.

test/Extensions/AzureBlobPayloads.Tests/AzureBlobPayloads.Tests.csproj
berndverst: Addressed all three points: added a comment explaining the `AssemblyName`/`InternalsVisibleTo` dependency, set `RootNamespace` to `$(AssemblyName)`, and removed the redundant `Azure.Storage.Blobs` package reference.

berndverst:
Thanks for calling this out. I made the deletion limitation explicit in both the focused test and PR description: the recovery test models re-creation after deletion has settled, while an immediate real-Azure retry can surface `409 ContainerBeingDeleted` and still propagates exactly as it did before caching. I also renamed the test accordingly.

I could not run `LargePayloadConsoleApp` end to end because this environment has no `DURABLE_TASK_SCHEDULER_CONNECTION_STRING`. I did run the final merged `BlobPayloadStore` directly against Azurite: two uploads and two downloads round-tripped successfully, and the Azurite request log recorded exactly one `PUT ...?restype=container` request. That verifies both the steady-state skip against an emulator and the request-count reduction on the changed storage path. The PR testing section now records both that result and the DTS prerequisite limitation.

YunchuWang (approved):

PR #788, opened 2026-07-24
Perf: avoid repeated orchestration-history scans for tracing

### LINKED ISSUE #769: Performance: avoid repeated orchestration-history scans for tracing
## 1. What is the issue?

The worker scans the full orchestration history to locate scheduling events for every relevant new completion event, even when no OpenTelemetry listener is active.

## 2. Likely priority

**High** for long-running orchestrations with large histories and fan-in batches.

## 3. Impact on performance

For a work item with `N` new completion events and `M` past events, tracing work can become O(N x M). The scans and LINQ iterator allocations occur before `ActivitySource.StartActivity` determines that tracing is disabled, so the cost is paid in the normal no-listener case.

Relevant worker path:

https://github.com/microsoft/durabletask-dotnet/blob/883211a3a7ab0b55bb90b9c9e498148e4d65ead9/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs#L589-L689

`OnRunOrchestratorAsync` scans `PastEvents` from local lookup functions invoked inside the `NewEvents` loop. It also scans `NewEvents.Concat(PastEvents)` to find `ExecutionStarted` before starting the tracing activity.

Relevant tracing helper:

https://github.com/microsoft/durabletask-dotnet/blob/883211a3a7ab0b55bb90b9c9e498148e4d65ead9/src/Shared/Grpc/Tracing/TraceHelper.cs#L67-L114

The tracing helper only discovers whether an activity can be created after these lookups.

## 5. Guidance for how to fix

Expose a cheap tracing-listener check from `TraceHelper` and skip all trace-event lookup work when no listener is registered. When tracing is enabled, build indexes once per work item for task-scheduled and sub-orchestration-created events, preserving the current first/last lookup semantics, rather than rescanning history for each new event. Add coverage for duplicate IDs, ordering semantics, and no-listener behavior.

berndverst:
## Summary

`OnRunOrchestratorAsync` in `src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs` scanned orchestration history to correlate tracing events unconditionally, and rescanned the full past-events list for every qualifying new event to find the originating `TaskScheduled`/`SubOrchestrationInstanceCreated` event — O(new events × past events), even when no tracing listener was registered.

This is a precise, internal-only performance fix:

- Added `TraceHelper.HasListeners`, backed by `ActivitySource.HasListeners()`, and gated tracing-correlation work behind it so history is not scanned when no listener is registered.
- Replaced the LINQ `Concat` scan for `ExecutionStarted` with a single-pass local function.
- Added `TraceHistoryEventLookup`, which pre-registers only correlation IDs referenced by the current work item's new events, then resolves both scheduling-event types in one lazy history scan. Repeated lookups remain O(1), total correlation work is O(past events + new events), and temporary index storage is O(unique correlation IDs in the current work item), not O(all scheduling events in history).
- Enforced the history invariant that each looked-up scheduling event type has a one-to-one event-ID mapping. Duplicate `TaskScheduled` or `SubOrchestrationInstanceCreated` IDs now throw a diagnostic `InvalidOperationException` instead of silently selecting an ambiguous event. With tracing enabled, the invalid work item is abandoned through the existing error path; valid histories are unaffected.

No public API or replay-output changes. For valid orchestration histories, trace output and orchestration behavior are unchanged.

## Tests

- `TraceHistoryEventLookupTests`: duplicate-ID invariant failures for both correlation types, no-match behavior, event-type filtering, empty input, completed and failed correlation events, lazy construction, one-pass mixed indexing, requested-ID-only storage, and per-ID/type duplicate isolation.
- `GrpcDurableTaskWorkerTests`:
  - Proves the no-listener path invokes neither lookup by completing invalid duplicate-ID history that would throw if indexed.
  - Proves the listener-enabled path abandons duplicate-ID history and never invokes the completion RPC.
- `TracingIntegrationTests.HistoryEventLookupCorrelatesDistinctScheduledOperations`: runs through the real in-process sidecar with two distinctly named activities and a sub-orchestration, then verifies exact client-span correlation for both indexes.
- Full `Worker.Grpc.Tests`: 165/165 passed.
- Full `Grpc.IntegrationTests`: 169/169 passed.
- `Worker.Grpc` builds across `netstandard2.0`, `net6.0`, `net8.0`, and `net10.0` with no errors.

Fixes #769


src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs
```diff
@@ -586,107 +586,110 @@ async Task OnRunOrchestratorAsync(
...
-                                    GetTaskScheduledEvent(newEvent.TaskFailed.TaskScheduledId);
+            static P.ExecutionStartedEvent? FindExecutionStartedEvent(P.OrchestratorRequest request)
+            {
+                foreach (P.HistoryEvent newEvent in request.NewEvents)
+                {
+                    if (newEvent.EventTypeCase == P.HistoryEvent.EventTypeOneofCase.ExecutionStarted)
+                    {
+                        return newEvent.ExecutionStarted;
+                    }
+                }
```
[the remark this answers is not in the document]
berndverst: Keeping the direct loops here. This runs for each traced orchestration work item, and `Where().Select().FirstOrDefault()` would add LINQ iterator allocations beyond the `RepeatedField<T>` enumerator in this performance-sensitive path. The loops preserve NewEvents-first precedence and short-circuit on the first match.

berndverst:
## Quantitative performance impact

I compared live head `d223fc2` with its merge base in a focused Release/.NET 10 harness on Windows x64 (AMD EPYC 7763). It used the actual `TraceHistoryEventLookup`/`TraceHelper.HasListeners` code and the exact old lookup algorithm. Activity creation, the rest of worker execution, and gRPC were excluded, so these are correlation-stage measurements rather than fleet-wide orchestration averages.

Let `H` be past-event count and `N_t`/`N_s` qualifying task/sub-orchestration completion counts.

| Path | Old past-event inspections | New work |
|---|---:|---:|
| No tracing listener | execution-start search + up to `N_t*H + sum(sub match positions)` | one `HasListeners()` check; **zero history enumeration** |
| Listener, task completions | `N_t*H` | one `H`-event task index + `N_t` dictionary lookups |
| Listener, sub-orchestration completions | `sum(sub match positions)` | one `H`-event sub index + `N_s` dictionary lookups |

### Representative, best, and adverse cases

“Representative” below means a selected ordinary shape, not production telemetry.

| Case | Listener / shape | Correlation latency, old -> new | Allocation/work item | Interpretation |
|---|---|---:|---:|---|
| **Average-style representative** | Off; `H=100`, 5 task completions | 4.40 us -> ~6 ns | 1,056 -> 0 B | The history work is eliminated; the new value is near harness overhead, so the ratio is not meaningful |
| Listener-on ordinary task case | `H=100`, 5 task completions | 4.38 -> 2.00 us (**-54.3%**) | 1,056 -> 4,704 B (**+3,648 B**) | Faster CPU path, higher temporary allocation |
| Listener-on ordinary mixed case | `H=100`, 3 task + 3 early sub matches | 2.55 -> 3.43 us (**~+34.5%**) | 1,024 -> 4,280 B (**+3,256 B**) | Small/early-match histories can regress |
| **Best measured large replay** | Off; `H=10,000`, 1,000 task completions | 79.8 ms -> near harness overhead | 160,256 -> 0 B | Removes the entire quadratic correlation scan |
| Listener-on large fan-out | `H=10,000`, 1,000 task completions | 80.46 ms -> 333.5 us (**-99.585%**, ~241x) | 160,256 -> 451,592 B (**+291,336 B**) | Large CPU/latency gain traded for a temporary index |
| Listener-on mixed fan-out | `H=10,000`, 500 task + 500 sub completions | 36.98 ms -> 490.7 us (**-98.673%**, ~75x) | 128,256 -> 431,438 B (**+303,182 B**) | Same tradeoff with two lazy indexes |
| **Constructed adverse case** | Listener on; one sub match at position 2 of `H=10,000` | 166 ns -> 95.7 us (**~576x as long**) | 352 -> 296 B | Old `FirstOrDefault` stops at event 2; new code builds the full lazy index |

The no-listener ~6 ns figures are close to benchmark-harness cost and should be read as “the scan was removed,” not as a credible millions-of-operations throughput claim.

### Memory and GC

The lazy indexes reference existing events; they do not copy protobuf messages. On this x64 runtime, estimated incremental dictionary storage while live is approximately **28-56 B per unique indexed ID**, including capacity slack/array headers but excluding the existing event objects.

- 50 task entries: about **2.7 KiB retained**, 4.7 KiB cumulatively allocated
- 5,000 task entries: about **236 KiB retained**, 452 KiB allocated
- Two 3,333-entry indexes: about **227 KiB retained**, 431 KiB allocated

Thus:

- **No listener:** both scan CPU and correlation allocations fall to zero.
- **Listener + repeated lookups:** CPU falls from `O(N*H)` to `O(H+N)`, but temporary memory becomes `O(unique history IDs)`.
- **Listener + one early sub lookup:** CPU can regress from `O(1)` expected early exit to `O(H)` index construction.
- A listener that samples every activity out still makes `HasListeners()` true and pays indexing costs.

### Whole-worker throughput interpretation

Correlation-stage speedup does not translate directly to orchestration throughput. Using Amdahl's law:

- If the old correlation work is 10% of turn CPU and the new path removes it, modeled turn throughput improves by about **11.1%**.
- If a large fan-out replay spends 50% of turn CPU in old correlation scans and that work becomes negligible, the modeled ceiling is about **2x** turn throughput.

Those are sensitivity examples. Actual gains require telemetry for listener prevalence and the distributions of `H`, qualifying new events, event mix, and match positions.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,92 @@
...
+using P = Microsoft.DurableTask.Protobuf;
+
+namespace Microsoft.DurableTask.Tracing;
+
+/// <summary>
+/// Provides indexed lookups of past history events by event ID, used to correlate new completion/failure
+/// events (e.g. "TaskCompleted") back to the history event that scheduled them (e.g. "TaskScheduled").
+/// </summary>
+/// <remarks>
+/// The indexes are built lazily, at most once per instance, and cached for the lifetime of the instance. This
```
sophiatev: Can we modify this comment? It's misleading because the index is not cached for the lifetime of the instance, but rather just for the lifetime of the workitem for the instance. The instance could have multiple work items, in which case we recreate this map, right?

So maybe just changing to "cached for the lifetime of the instance work item" would be more accurate

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,92 @@
...
+        this.pastEvents = pastEvents;
+    }
+
+    /// <summary>
+    /// Gets the "TaskScheduled" history event with the given event ID, if any.
+    /// </summary>
+    /// <param name="eventId">The event ID to look up.</param>
+    /// <returns>The matching event, or <see langword="null"/> if none is found.</returns>
+    /// <remarks>
+    /// If more than one "TaskScheduled" event shares the given event ID, the last one encountered (in history
```
sophiatev: We should actually throw an exception in this case, this situation should be impossible. It should be 1:1

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,92 @@
...
+            : null;
+    }
+
+    /// <summary>
+    /// Gets the "SubOrchestrationInstanceCreated" history event with the given event ID, if any.
+    /// </summary>
+    /// <param name="eventId">The event ID to look up.</param>
+    /// <returns>The matching event, or <see langword="null"/> if none is found.</returns>
+    /// <remarks>
+    /// If more than one "SubOrchestrationInstanceCreated" event shares the given event ID, the first one
```
sophiatev: Same here

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
```diff
@@ -364,6 +366,154 @@ public async Task DispatchWorkItem_ActivityRequest_NotificationFailure_Completes
...
+        // Act
+        InvokeDispatchWorkItem(processor, orchestratorWorkItem, CancellationToken.None);
+        await completed.Task.WaitAsync(TimeSpan.FromSeconds(5));
+
+        // Assert: work item processing still completes normally with no listener registered.
+        clientMock.VerifyAll();
+    }
+
+    [Fact]
+    public async Task DispatchWorkItem_OrchestratorRequest_WithActivityListener_UsesFirstAndLastWinsSemantics()
```
sophiatev: As per the comment in `TraceHistoryEventLookup`, we actually want to throw an exceptoin in this case, so let's update the test appropriately

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
```diff
@@ -364,6 +366,154 @@ public async Task DispatchWorkItem_ActivityRequest_NotificationFailure_Completes
         logs.Should().Contain(log => log.Message.Contains("Activity notification callback failed for phase 'Completed'"));
     }
 
+    // The following two tests both touch the process-wide "Microsoft.DurableTask" ActivitySource used by
+    // TraceHelper, so they are kept in this class (whose test methods xunit runs sequentially by default) to
+    // avoid flaky interference between them.
+    [Fact]
+    public async Task DispatchWorkItem_OrchestratorRequest_NoActivityListeners_SkipsTracingWorkAndCompletes()
```
sophiatev: Can we confirm that the `TraceHistoryEventLookup` methods are not invoked in this case?

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
sophiatev, line 1: Can we add any integration tests for this?

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
```diff
@@ -0,0 +1,134 @@
...
+
+using Microsoft.DurableTask.Tracing;
+using P = Microsoft.DurableTask.Protobuf;
+
+namespace Microsoft.DurableTask.Worker.Grpc.Tests;
+
+public class TraceHistoryEventLookupTests
+{
+    [Fact]
+    public void GetTaskScheduledEvent_DuplicateEventIds_ReturnsLastMatch()
```
sophiatev: Same comment about new semantics

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        // Act
+        P.HistoryEvent? result = lookup.GetTaskScheduledEvent(1);
+
+        // Assert
+        result.Should().NotBeNull();
+        result!.TaskScheduled.Name.Should().Be("SecondScheduled");
+    }
+
+    [Fact]
+    public void GetSubOrchestrationInstanceCreatedEvent_DuplicateEventIds_ReturnsFirstMatch()
```
sophiatev: And here

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
berndverst: Updated in e736e30. The remarks now say the indexes are cached for the lifetime of the orchestrator work item (and rebuilt for a later work item), rather than for the lifetime of the orchestration instance.

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
berndverst: Added in e736e30. `HistoryEventLookupCorrelatesDistinctScheduledOperations` runs through the real in-process sidecar with two distinctly named activities and a sub-orchestration, then verifies the exact client spans produced by both indexed correlation paths.

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
berndverst: Updated in e736e30. With a tracing listener registered, duplicate correlation IDs now cause the processor to abandon the work item; the test also verifies that the completion RPC is never invoked.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
berndverst: Updated in e736e30 as well: duplicate `SubOrchestrationInstanceCreated` IDs now throw the same clear invariant-violation exception.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
berndverst: Agreed. e736e30 now treats duplicate `TaskScheduled` IDs as an invariant violation and throws an `InvalidOperationException` with the event type and duplicate ID instead of selecting a match.

test/Worker/Grpc.Tests/GrpcDurableTaskWorkerTests.cs
berndverst: Confirmed and made explicit in e736e30. The no-listener test uses duplicate IDs that would throw if either lookup method built its index; successful work-item completion therefore proves the tracing fast path invokes neither lookup.

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
berndverst: Updated in e736e30. The focused `TaskScheduled` duplicate-ID test now requires `InvalidOperationException` and checks that the diagnostic identifies the event type and ID.

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
berndverst: Updated in e736e30. The focused sub-orchestration duplicate-ID test now requires the same fail-fast behavior and diagnostic detail.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,93 @@
...
+            catch (ArgumentException exception)
+            {
+                throw new InvalidOperationException(
+                    $"Past orchestration history contains multiple '{eventType}' events with event ID "
+                    + $"'{historyEvent.EventId}'.",
+                    exception);
+            }
+        }
+
+        return index;
```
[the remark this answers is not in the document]
berndverst: This is intentional follow-up behavior from Sophia's review: she clarified in both lookup threads that duplicate IDs are impossible under the history invariant and must throw rather than retain first/last-wins behavior. The prior PR description was stale, so I have updated it to document the one-to-one invariant, fail-fast behavior for invalid history, and unchanged behavior for valid histories. The focused, processor, and integration tests now match that reviewed contract.

berndverst:
## Updated quantitative impact after demand-shaped indexing

This supersedes the listener-on index measurements in my earlier comment. Head `3830a05` now pre-registers only correlation IDs referenced by the current work item's `NewEvents`, scans `PastEvents` once for both scheduling-event types, and retains only those requested IDs.

I measured the old repeated-scan algorithm, the prior full-history index, and the current demand-shaped index in a focused Release/.NET 10 harness on Windows x64 (AMD EPYC 7763), using real protobuf `HistoryEvent` objects. These numbers isolate correlation work; they are not whole-worker throughput measurements.

| Scenario | Shape | Old repeated scan | Prior full index | Current demand-shaped index | Current vs old |
|---|---|---:|---:|---:|---:|
| **Average-style task case** | `H=100`, 50 scheduled tasks, 5 completions | 3.160 us / 832 B | 1.377 us / 4,736 B | **0.793 us / 616 B** | **-74.9% latency, -26.0% allocation, ~4.0x correlation throughput** |
| Average-style mixed case | `H=100`, 33 task + 33 sub events, 3 lookups each | 2.516 us / 1,024 B | 1.845 us / 4,344 B | **0.888 us / 648 B** | **-64.7% latency, -36.7% allocation, ~2.8x throughput** |
| **Best listener-on fan-in** | `H=10,000`, 5,000 scheduled tasks, 1,000 completions | 54.33 ms / 160,034 B | 147.1 us / 451,641 B | **80.1 us / 102,368 B** | **-99.853% latency, -36.0% allocation, ~678x throughput** |
| Large mixed fan-in | `H=10,000`, 3,333 task + 3,333 sub events, 500 lookups each | 35.57 ms / 160,066 B | 210.4 us / 431,611 B | **91.1 us / 96,408 B** | **-99.744% latency, -39.8% allocation, ~390x throughput** |
| **Constructed worst case** | listener on; one sub match at position 2 of `H=10,000` | 89.8 ns / 184 B | 35.67 us / 320 B | **39.00 us / 352 B** | **~434x latency, +168 B** |

Compared with the prior full index, the current large task case reduces index latency another **45.5%** and allocation **77.3%**; the large mixed case reduces latency **56.7%** and allocation **77.7%**. Storage is now `O(unique correlation IDs in NewEvents)` rather than `O(all scheduling IDs in PastEvents)`.

The no-listener path is unchanged from the earlier measurement: it performs one `HasListeners()` check and **zero tracing-history enumeration or correlation allocation**. The earlier representative no-listener case (`H=100`, 5 task completions) fell from 4.40 us and 1,056 B to approximately harness overhead and 0 B; the large no-listener replay (`H=10,000`, 1,000 completions) removes roughly 79.8 ms of correlation work.

The constructed single-early-sub case remains the adverse boundary. The old `FirstOrDefault` could return at position 2, while the reviewer-requested one-to-one invariant requires scanning the applicable history to detect a later duplicate and fail rather than silently selecting one. That `O(H)` validation cost cannot retain the old early-exit complexity while also guaranteeing duplicate detection. The current design removes the avoidable all-history storage and keeps that cost confined to listener-enabled tracing.

Reciprocal latency is used only as a correlation-stage throughput proxy. Whole-worker gains depend on the fraction of turn CPU previously spent in tracing correlation, listener prevalence, history sizes, and completion-batch distributions.

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
```diff
@@ -0,0 +1,317 @@
...
+                        SubOrchestrationInstanceFailed =
+                            new P.SubOrchestrationInstanceFailedEvent { TaskScheduledId = eventId },
+                    }
+                    : new P.HistoryEvent
+                    {
+                        SubOrchestrationInstanceCompleted =
+                            new P.SubOrchestrationInstanceCompletedEvent { TaskScheduledId = eventId },
+                    };
+                newEvents.Add(newEvent);
+            }
```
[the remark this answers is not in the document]
berndverst: Applied in `0792af0`: the sub-orchestration correlation-event projection now uses `Select` with `AddRange`.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,177 @@
...
+/// The indexes contain only scheduling event IDs referenced by the current work item's new completion/failure
+/// events. They are built together in one lazy scan and cached for the lifetime of the orchestrator work item.
+/// This avoids both re-scanning the full set of past events for every new event and retaining unrelated history
+/// events, bounding lookup storage by the number of correlation IDs in the current work item.
+/// </remarks>
+sealed class TraceHistoryEventLookup
+{
+    readonly IEnumerable<P.HistoryEvent> pastEvents;
+
+    Dictionary<int, P.HistoryEvent?>? taskScheduledEventsByEventId;
```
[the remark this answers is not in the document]
berndverst: Applied in `0792af0`: this dictionary reference is now `readonly`; constructor initialization and later element updates remain unchanged.

src/Shared/Grpc/Tracing/TraceHistoryEventLookup.cs
```diff
@@ -0,0 +1,177 @@
...
+    Dictionary<int, P.HistoryEvent?>? subOrchestrationInstanceCreatedEventsByEventId;
```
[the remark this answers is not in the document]
berndverst: Applied in `0792af0`: this dictionary reference is now `readonly`; its contents remain mutable during lazy index population.

test/Worker/Grpc.Tests/TraceHistoryEventLookupTests.cs
```diff
@@ -0,0 +1,317 @@
...
+                    ? new P.HistoryEvent
+                    {
+                        TaskFailed = new P.TaskFailedEvent { TaskScheduledId = eventId },
+                    }
+                    : new P.HistoryEvent
+                    {
+                        TaskCompleted = new P.TaskCompletedEvent { TaskScheduledId = eventId },
+                    };
+                newEvents.Add(newEvent);
+            }
```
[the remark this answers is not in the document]
berndverst: Applied in `0792af0`: the task correlation-event projection now uses `Select` with `AddRange`.

